PR #851, opened 2014-12-19
Add concretize_service_request function

### LINKED ISSUE #896: ServiceRequest and its performer don't allow for using different regions
Right now (actually, not yet in master, but in the concretize-service-request branch) when you create a dispatcher that can support `ServiceRequest`, you specify the region up front. But we have cases where requests to certain services (metrics) need to go to a specific region, which is not the same as the deployment region.

I think there are a couple of options here:

- we could just add a region parameter to ServiceRequest, and have code explicitly provide it when the default is not desired
- we could allow special-casing those specific services (metrics), potentially in a configuration-driven way, e.g. the ServiceRequest performer could look for a region in config['services'][service_name]['region_override'] or something.

radix:
This branch introduces a new intent, `TenantScope`, which allows specifying a tenant for any effect which might perform `ServiceRequest`s, and a performer for `TenantScope` which itself provides a performer for `ServiceRequest`. This is a pretty clever (not to be mistaken for hacky, IMO ;) use of the new Effect performer/dispatcher mechanics introduces in Effect 0.1a12. Please feel free to grill me for a better understanding of how it works, if necessary.

Some points for reviewers:

- `concretize_service_request` and `perform_tenant_scope` should bear the brunt of your reviewing efforts.
- The tests for `concretize_service_request` are hack-'n'-pasted from the get_request_func tests. Once we get rid of get_request_func, we can delete the old tests.
- This branch also changes ServiceRequest from taking `success_codes` to the newer `success_pred` system.
- This change draws our attention to #896. This isn't exactly a new problem, as request_func had the same issue. I've added a comment to the place in metrics.py where we're creating the metrics-specific request_func and dispatchers. So: things should work with this branch, but we need to resolve #896 before I'm really happy with ServiceRequest.


manishtomar:
@radix You might wanna forward-merge this.

radix:
@manishtomar - thanks! done.

otter/effect_dispatcher.py
```diff
@@ -28,3 +32,15 @@ def get_dispatcher(reactor):
         }),
         make_twisted_dispatcher(reactor),
     ])
+
+
+def get_full_dispatcher(reactor, authenticator, log, service_mapping, region):
```
manishtomar: How is it different than having one dispatcher? `TenantScope` is not going to be used if it's effect is not going to be performed.

radix: I'm... not sure what you mean. Are you asking why there's both a get_simple_dispatcher and a get_full_dispatcher? get_simple_dispatcher is pretty much only used for tests, but also temporarily in places like _rcv3.py. It should either be deleted or be relegated to purely a testing utility soon.

manishtomar: Ok cool. That's what I wanted to know.

radix: yay :)

otter/test/test_http.py
```diff
@@ -128,8 +146,138 @@ def test_defaults(self):
...
+        """
+        input_json = {"a": 1}
+        output_json = {"b": 2}
+        svcreq = service_request(ServiceType.CLOUD_SERVERS, "GET", "servers",
+                                 data=input_json).intent
+        eff = self._concrete(svcreq)
+        next_eff = resolve_authenticate(eff)
+        result = resolve_effect(next_eff,
+                                stub_pure_response(json.dumps(output_json)))
+        self.assertEqual(next_eff.intent.data, json.dumps(input_json))
```
manishtomar: Minor nitpick: It might be easier for reader to move this at line 214 after getting `next_effect`. And a space after the check would be nice. Something like:
```python
# input json is as expected
next_eff = resolve_authenticate(eff)
self.assertEqual(next_eff.intent.data, json.dumps(input_json))

# output json is as expected
result = resolve_effect(next_eff, stub_pure_response(json.dumps(output_json)))
self.assertEqual(result, output_json)
```

manishtomar:
LGTM modulo nitpick above :)

@lvh might want to take a look though.

Also, would like to run RCv3 tests. I'll do so.

otter/http.py
```diff
@@ -113,7 +122,7 @@ def service_request(
         service_type, method, url, headers=None, data=None,
         log=None,
         reauth_codes=(401, 403),
-        success_codes=(200,),
+        success_pred=has_code(200),
```
lvh: Huh. Why is this showing up in this diff? master already has this change.

radix: master doesn't have this change for service_request.

lvh: Ah, right; my bad. I saw that signature and forgot that `ServiceRequest` was already there, next to the old request_fn stuff :-)

otter/http.py
```diff
@@ -145,11 +155,105 @@ def service_request(
...
+class TenantScope(object):
+    """
+    An intent that specifies a tenant for any effect which might make
+    :obj:`ServiceRequest`s.
+
+    In other words, something like this::
+
+        perform(dispatcher, TenantScope(Effect(ServiceRequest(...))))
+
+    will make the ServiceRequest bound to the specified tenant.
```
lvh: Nitpick, non-blocking: I think it might be useful to document that this class doesn't really introduce behavior, but only describes that intent, and that you still need to have some extra stuff somewhere else (i.e. a dispatcher that knows about TenantScopes) for this to work.

radix: I'm not totally sure what you mean, but I can extend the docs here in a few reasonable ways. (no intents introduce behavior; they all require performers in dispatchers. in this case, though, I agree it's a good idea to explicitly call out the performer since you'd only ever want to use one performer)

lvh: Right, but you might reasonably expect a delay to work on any performer, but this requires an extra-special one.

otter/http.py
```diff
@@ -145,11 +155,114 @@ def service_request(
...
+
+    Use a partially-applied :func:`perform_tenant_scope` as the performer
+    for this intent in your dispatcher.
```
lvh: Beautiful!

lvh:
LGTM! I'm really happy with how this scoping effect turned out. Might be useful for other stuff in the future. Not merging yet pending RCv3 test concerns, build, but please consider this a +1 from me and merge whenever you'd like :-)

manishtomar:
RCv3 tests pass :) There is problem with the tests. This is good to go.

PR #852, opened 2014-12-22
Update otter to work with the latest changes to the Effect dispatcher API

radix:
This PR is GOOD TO GO! (I think)

I worked on improving the Effect API over the weekend, and this PR updates Otter to work with that changed API.

The main changes in Effect:

- The `perform_effect` method on intents is now deprecated
- The `perform` function now requires you to pass in a dispatcher that supports all the intents you'll use in your application
- `TypeDispatcher` does type(intent)-based dispatching to a performer function
- `ComposedDispatcher` composes multiple dispatchers together
- resolve_stubs also requires a dispatcher, just like perform.


otter/test/utils.py
```diff
@@ -599,5 +601,19 @@ def resolve_retry_stubs(eff):
     separately to determine that the policy is as expected.
     """
     assert type(eff.intent) is Retry
-    is_error, intermediate_result = guard(resolve_stubs, eff.intent.effect)
+    try:
```
radix: `guard` became a private function implementation detail of effect. I knew it was a bit dodgy to use it here when I first added this code...

lvh:
Looks like this needs to be forward-merged.

lvh:
Huh, I'm confused by the number ordering. #853 should be done first, and then #853 merged into this one, and then merged into master?

radix:
@lvh - no, this isn't a PR-to-PR merge request, it's just that I merged in the branch for #853 into this one to reduce the number of eventual conflicts. My hope was that #853 would be much quicker to get merged since it's such a simple change.

otter/test/test_auth.py
```diff
@@ -743,8 +746,8 @@ def test_authenticate(self):
         mock_auth = iMock(IAuthenticator)
         mock_auth.authenticate_tenant.return_value = succeed(result)
         log = object()
-        intent = Authenticate(mock_auth, 'tenant_id1', log)
-        self.assertEqual(self.successResultOf(intent.perform_effect(None)), result)
+        eff = Effect(Authenticate(mock_auth, 'tenant_id1', log))
+        self.assertEqual(sync_perform(get_dispatcher(None), eff), result)
```
manishtomar: Non-blocking FYI: Its a little strange to use `sync_perform` when `Authenticate` is a `deferred_performer`.

otter/test/test_retry.py
```diff
@@ -482,6 +482,12 @@ def get_exc_info():
         return sys.exc_info()
 
 
+def _perform_func(eff):
```
manishtomar: Nitpick: `_peform_func_intent`?

otter/util/retry.py
```diff
@@ -340,12 +340,14 @@ class Retry(object):
...
     """
-    def perform_effect(self, dispatcher):
-        """
-        Invoke :func:`effect.retry.retry` with the effect and the
-        should_retry function.
-        """
-        return effect_retry(self.effect, self.should_retry)
+
+
+def perform_retry(dispatcher, intent):
```
manishtomar: `dispatcher` is not used since this is not really _performing_ the effect. Maybe change the function to `invoke_effect_retry`?

otter/test/util/test_pure_http.py
```diff
@@ -19,15 +19,16 @@
     add_content_only, add_error_handling, add_json_response,
     add_json_request_data)
 from otter.util.http import APIError
-from otter.test.utils import stub_pure_response, StubResponse, StubTreq
+from otter.test.utils import stub_pure_response, StubResponse, StubTreq, resolve_stubs
+from otter.effect_dispatcher import get_dispatcher
 
 
-Constant = lambda x: StubIntent(ConstantIntent(x))
+ESConstant = lambda x: Effect(Stub(Constant(x)))
```
manishtomar: This is pretty useful for other tests too. How about putting it in `test.utils.py`?

otter/util/pure_http.py
```diff
@@ -27,18 +28,20 @@ class Request(object):
...
-        returnValue((response, content))
+@deferred_performer
+@inlineCallbacks
+def perform_request(dispatcher, intent):
+    """
+    Perform the request with treq.
+
+    :return: A two-tuple of (HTTP Response, content as bytes)
+    """
+    response = yield intent.treq.request(intent.method.upper(), intent.url,
```
manishtomar: Now that there is clear decoupling of intent and performer, should implementation detail like `treq` be there in `intent`?

otter/test/test_auth.py
radix: Think of it like using successResultOf for a Deferred

otter/util/retry.py
radix: In Effect's parlance, this function is a "Performer", and the convention is to name performers as `perform_<intent_name>`, or sometimes `perform_<intent_name>_with_<strategy>`. Performers can be implemented by returning Effects instead of actually doing stuff imperatively.

The `dispatcher` argument being unused is the common case - it's only necessary if your performer needs to make a *recursive* call to perform. Most performers, even ones that don't return Effects, don't use it. For example, the performer for our pure_http.Request doesn't. But stuff like ParallelEffects does need it, since it needs to perform its children effects.

otter/test/util/test_pure_http.py
radix: I've actually added them to effect.testing, but not in this release. I'll make that change to otter in another PR.

otter/util/pure_http.py
radix: You're right, it definitely shouldn't, and it was a known compromise from the very beginning. But I think that should be resolved in another PR, or in fxhttp, when we (tomprince and I and whoever else wants to help) implement that. I don't mind this code-smell for a little bit longer :)

otter/util/retry.py
manishtomar: > Performers can be implemented by returning Effects instead of actually doing stuff imperatively.

Hmm. I am not sure I completely agree with this or understand this. In any case, this doesn't block the PR :) We can continue this conversation someplace else.

manishtomar:
LGTM before the lints :P

manishtomar:
Also would like to run RC3 changes before merging.

manishtomar:
RCv3 test pass. Good to merge :)

Issue #817, opened 2014-12-11
refactor execute_convergence to do less stuff


radix:
execute_convergence should look a lot more like the "convergeration" function as outlined in #801. It's okay if it's not _exactly_ the same, but we should definitely reduce its surface area.

- the code that composes together gathering should be moved to otter.convergence.gathering
- the code that generates steps and then optimizes them should be moved to otter.planning
- the core steps -> effect code should be moved to otter.convergence.effecting

here's the goal, to reiterate from #801 :

```python
def convergeration(tenant):
   return (gathering.gather(tenant)
           .on(success=planning.plan)
           .on(success=effecting.reify))
```

radix:
This was fixed by #853.

PR #853, opened 2014-12-22
Factor out more code from execute_convergence

radix:
`execute_convergence` still has a bit too much going on, so I've factored out some stuff to other `otter.convergence` modules.

- `plan` was added to `otter.convergence.planning`, which creates an optimized plan
- `steps_to_effects` was added to `otter.convergence.effecting`, so `execute_convergence` doesn't need to care about the step -> request step.
- `get_all_convergence_data` was added to `otter.convergence.gathering`, which was the biggest win for the size of `execute_convergence`.

At the same time, `json_to_LBConfigs` and `get_desired_group_state` were moved from `gathering` to `composition` because I realized they're not about *gathering* data -- they're about converting pre-existing Otter data structures (LB configs and launch configs) to convergence data structures. `composition` isn't really the perfect place for them, but I think it's better than `gathering`.


otter/test/convergence/test_composition.py
```diff
@@ -25,25 +70,33 @@ def setUp(self):
...
         self.servers = [
-            {'id': 'a', 'state': 'ACTIVE', 'created': now(),
-             'addresses': {'private': [{'addr': 'ip1', 'version': 4}]}},
-            {'id': 'b', 'state': 'ACTIVE', 'created': now(),
-             'addresses': {'private': [{'addr': 'ip2', 'version': 4}]}}
+            NovaServer(id='a', state=ServerState.ACTIVE, created=0, servicenet_address='ip1'),
+            NovaServer(id='b', state=ServerState.ACTIVE, created=0, servicenet_address='ip2'),
         ]
 
+    def _get_get_all_convergence_data(self, servers, group_id, reqfunc):
```
radix: Well, here are the options I know of, and which are really a generalization of the whole problem of testing, dependency injection, and abstraction.

- don't parameterize it, and just patch/mock it in the tests.
- parameterize it, which is really only marginally more reliable than patch/mocking...
- don't parameterize it at all, all the way down to get_scaling_group_servers et al, and have the tests check all the Effects of HTTPRequests that they'll get back. this is pretty undesirable since it means effectively duplicating a lot of the tests for those lower level functions.
- abstract the whole thing with a new Intent that describes fetching scaling group servers, etc. This is a big step that's kind of tedious and one I'm wary to do too much, as it leads to a massive proliferation of new abstractions. Maybe I'm too wary, and it's a great idea, but I'm actually worried that I've been seeing too many nails with this hammer lately (ServiceRequest, Authenticate, etc).

I am quite open to the idea that I've been getting stuck in a particular pattern of thinking, so if you have any recommendations, I'm happy to hear them :)

radix:
I've addressed all review comments except for lvh's "iffy api design" one. I'll wait a bit before marking as pending merge.

otter/convergence/composition.py
```diff
@@ -66,3 +46,36 @@ def tenant_is_enabled(tenant_id, get_config_value):
     """
     enabled_tenant_ids = get_config_value("convergence-tenants")
     return (tenant_id in enabled_tenant_ids)
+
+
+def json_to_LBConfigs(lbs_json):
```
manishtomar: Wouldn't this function make more sense in `gathering.py`?

radix: I moved it out of gathering.py because it's not related to fetching data -- it's only to convert otter's own format for LB configs to the LBConfig object (actually, maybe it would make more sense in model.py?)

otter/test/convergence/test_composition.py
manishtomar: No specific opinion about choosing the methodology above :) but `_get_gacd_func` name might be more clear?

otter/convergence/composition.py
manishtomar: I don't have strong opinion but I would argue that it is more related to gathering as this function is based on raw JSON knowledge of gathered response. It's also ok to put it in model. I am actually fine either way. Probably just bike-shedding. You can ignore it.

radix: My point is it that it's not about a "gathered response", as you said. it's from our own launch configuration format. I just want to make sure that the function's purpose is clear :)  It's not about gathering nor about processing gathered data, but just about modeling some of our own launch configuration format.

otter/test/convergence/test_effecting.py
```diff
@@ -112,3 +114,36 @@ def test_multiple_requests_of_different_type(self):
...
+class Steppy(object):
+    """A dummy step."""
+    def as_request(self):
+        """Return a simple GET .../whatever on CLOUD_LOAD_BALANCERS"""
+        return Request(service=ServiceType.CLOUD_LOAD_BALANCERS,
+                       method="GET",
+                       path="whatever")
+
+
+class StepsToEffectTests(SynchronousTestCase):
```
manishtomar: This seems to be testing `_reqs_to_effect` indirectly. If so, then are `RequestToEffectTests` required?

radix: I think it's reasonable. It's only testing one case, just to make sure that it's likely to be composed correctly.

manishtomar:
LGTM modulo `_get_gacd_func` comment above :) (which is non-blocking) 

Also, would be nice to run rc3 tests against this as it touches that code.

radix:
Oh, to make the link explicit, this PR is to resolve issue #817

lvh:
Looks like this, too, needs to be forward-merged again.

otter/test/convergence/test_effecting.py
lvh: I agree; `_reqs_to_effect` is an implementation detail, and one that we're hoping to eliminate. However, this test should keep working.

lvh:
I forward-merged this, so I think this is ready for review again (at least depending on what Jenkins feels).

I think it generally makes sense to merge nontrivial PRs oldest first, so, let's try to get this one in before it has to be merged forward again.

lvh:
That said, I think @radix should give this a once-over, since that's really the only way to verify that this PR is *semantically* correct.

radix:
So, things got a little wonky on this branch, and it has some other branches merged into it that don't technically need to be there: #876. let's just hope those get merged soon and then this branch should be good to go :)

manishtomar:
Isn't #876 wants this to be merged first?

lvh:
So, the dependent branch that needs to be merged first is #870, then #876, or alternatively just merge this after #870. I think the three-PR route is cleaner, because it means you don't pollute a PR with changes it accidentally incorporated.

manishtomar:
I am going to run RCv3 tests on this.

manishtomar:
The tests are fine. Good to merge :)

radix:
hmm... you added the pending merge label and now I'm not sure if I should click the green button yet... :)

Issue #855, opened 2014-12-23
Poll until resource deletion is confirmed


sam-falvo:
As one test finishes, another starts up, even though the resources created in one test aren't fully disposed of yet.  This affects, in particular, the `test_create_scaling_group_with_pool_counts` test.

First and foremost, I need to check for server deletion, I *think*.  But, other kinds of resources may be required.  Further testing is required.

cyli:
@sam-falvo Is this issue complete?

PR #856, opened 2014-12-23
RCv3 positive integration tests

### LINKED ISSUE #840: (EST 5) Clean up test logic
- [x] Extraneous print statements
- [x] time.sleep() replaced with something more reasonable.
- [x] Doc-strings
- [x] Interior comments
- [x] Double-check logic matches the descriptions
- [x] Other PEP8 issues
- ...others as appropriate.  Add them here.

sam-falvo:
Closes #840.


autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
```diff
@@ -0,0 +1,757 @@
...
+    def test_create_scaling_group_with_pool_counts(self):
+        """
+        Test that it is possible to create a scaling group with min_servers servers
+        connected to an RCv3 LB pool.
+
+        This is a simple smoke test that only checks the node counts.
+        """
+        # This sleep is necessary because other tests have just completed, but
+        # their resources haven't yet been completely freed up.  TODO(sfalvo):
+        # See Github issue #855.
```
lekhajee: Is that cause delete scaling group is taking long or failing in unknown ways. 

I am not sure I understand how polling on the delete scaling group will help us fix the issue (#855) of disposing all the resources from the previous test, cause receiving a 204 on a delete group does not mean that the resources in it have been deleted successfully. I still think performing the empty scaling group followed by polling to see that the actually group empties in the tearDown would help.

sam-falvo: I'm not sure how I can communicate this more clearly.

What I'm saying we need is to add polling to delete_X, for whatever X we care to implement.  X doesn't have to be scaling group, and in all honesty, it's rarely the cause of our intermittent failures.  However, to use it as an example, if delete_scaling_group can be altered so that it lists its resources prior to actually issuing the delete, then we have plenty of information on-hand to poll to make sure that all its attached resources are properly disposed of.  OR, if we detect an error in the process, we can fail an assert.

It's a lot better than just waiting and hoping for the best.

lekhajee: ouch!

sam-falvo: I know, right?!  :(  A lot more coding effort than I'd like to put in, but it's what's gotta be done.  :(  I have a separate issue open for it right now, and was hoping to work on that once this one is reviewed.

autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
```diff
@@ -0,0 +1,757 @@
...
+
+        cls.cloud_servers_on_node = []
+        cls.private_network = {'uuid': '11111111-1111-1111-1111-111111111111'}
+        cls.public_network = {'uuid': '00000000-0000-0000-0000-000000000000'}
+        cls.rackconnect_network = {'uuid': cls.rcv3_cloud_network}
+        cls.min_servers = 1
+
+        # Since RCV3 pools must be created by human intervention,
+        # verify that at least one exists before proceeding with tests
+        tenant_pools = cls.rcv3_client.list_pools().entity.pools
```
sam-falvo: Good catch.  I removed `_get_available_pools`.  It wouldn't work in this context anyway because it's a class method instead of an object method.

autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
```diff
@@ -0,0 +1,757 @@
...
+        self.assertEquals(self.create_resp.status_code, 201,
+                          msg='The create failed with {0} for group '
+                          '{1}'.format(pool_group_resp.status_code, pool_group.id))
+        self.assertEquals(pool_group.launchConfiguration.loadBalancers[0].loadBalancerId,
+                          self.pool_1.id,
+                          msg='The launchConfig for group {0} did not contain the load balancer'
+                          .format(pool_group.id))
+        self.assertEquals(pool_group.launchConfiguration.loadBalancers[0].type, "RackConnectV3",
+                          msg='Load balancer type {0} is not correct for RackConnect pools'
+                          .format(pool_group.launchConfiguration.loadBalancers[0].type))
```
lvh: Looks like it is; it's just cutting off before the "common scaling group assertions".

autoscale_cloudcafe/autoscale/config.py
```diff
@@ -369,3 +369,34 @@ def autoscale_nno_lno_aa(self):
...
+        Get the catalog name of the RCV3 endpoint
+        """
+        return self.get('rcv3_endpoint_name')
+
+    @property
+    def rc_load_balancer_pool(self):
+        """
+        Name and id of the shared RCV3 LB
+        """
+        print "Get rc_load_balancer_pool ..... "
```
lvh: Is this print statement still supposed to be here?

autoscale_cloudcafe/autoscale/config.py
```diff
@@ -369,3 +369,34 @@ def autoscale_nno_lno_aa(self):
...
+
+    @property
+    def rcv3_endpoint_name(self):
+        """
+        Get the catalog name of the RCV3 endpoint
+        """
+        return self.get('rcv3_endpoint_name')
+
+    @property
+    def rc_load_balancer_pool(self):
```
lvh: Is the naming (rc vs rcv3) here intentional?

lvh:
FYI, the linter in #877 is materially better. I don't know when what will get merged, but you may want to merge forward regardless.

autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
sam-falvo: I guess what I meant was, why isn't GitHub clearing this from the list of comments (and showing it as "commented on an outdated diff").

manishtomar:
@sam-falvo Could you please forward merge master on this branch? We are using this branch to test any RCv3 changes in otter.

lvh:
(That will also get you a kind-of better linter. You should still run `make lint` manually; until #912 is resolved.)

autoscale_cloudcafe/autoscale/behaviors.py
```diff
@@ -538,3 +578,90 @@ def get_webhooks_properties(self, webhook_list):
...
+            asserter = DefaultAsserter()
+
+        # If we're checking up on Autoscale, we need to make sure that our
+        # group is even attempting to build the right number of servers.  If
+        # we're not, there's no point in continuing; we'll never attain the
+        # desired capacity anyway.
+        #
+        # Unfortunately, there's no way to check this on RackConnect's API,
+        # since it has no mechanism to report back its current "plans."
+        if api == 'Autoscale':
```
manishtomar: This `if` block broke the tests. It is only checking if desired matches and returns instead of ensuring active list is reached. Also, I don't understand its purpose. Is it meant to be an optimization by checking desired first? If so, that is already done line 641 below.

kivattik:
In theory, this is ready for final review except for compliance with the new PEP8 line limit standard.

cyli:
Unfortunately, I haven't figured out a way to run just the `system.group` tests, or just the `system.policy` tests.

Looking at the opencafe code, it looks like the package is only the the first folder, and the module regex only matches by the `test_*.py` name.  

So I think the only way to run all tests except for the non-passing mimic RCv3 tests is to enumerate every single python module in `autoscale_cloudroast/test_repo/system`  :|  

I have some ideas for automatically skipping RCv3 tests if RCv3 is not available, or skipping or running certain tests based on the presence of mimic, but we can do that in a separate PR.

As a temporary measure, to not break our existing testing jobs/make the deploy pipeline somewhat easer, can we move the rcv3 tests into their own package temporarily (e.g. out of `system/integration` and into `rcv3` or something).

I volunteer to address the auto-skipping and re-org in a later PR.

cyli:
retest this please

lvh:
@kivattik @sam-falvo There's also @manishtomar's comment here: https://github.com/rackerlabs/otter/pull/856#discussion_r23180433

manishtomar:
@cyli They've fixed the code so that this doesn't happen but I am still failing to understand the necessity of the if statement. I'll look at it again.

autoscale_cloudcafe/autoscale/behaviors.py
sam-falvo: We added a comment explaining why the duplication exists.  https://github.com/rackerlabs/otter/pull/856/files#diff-ba323cc90dc07574ea6a28176ff6a932R661

manishtomar: FWIW, that code is unnecessary since otter currently returns desired to be active + pending (which I think is a bug). So, you only need to check if pending + active = expected. Sorry about the confusion :/

autoscale_cloudcafe/autoscale/client.py
```diff
@@ -870,3 +902,86 @@ def delete_node(self, load_balancer_id, node_id, requestslib_kwargs=None):
...
+                            response_entity_type=RackConnectNetworkInfo)
+
+    def get_pool_info(self, pool_id):
+        """
+        :summary: Get information on a given load_balancer_pool
+        :return: Response Object
+        :rtype: Response Object
+        """
+        url = self.url + '/load_balancer_pools/{0}'.format(pool_id)
+        print("... RCV3 request ... ", url)
```
cyli: Would it make sense to remove this print statement and the three below it?  These seem more like debug print statements than actual useful information in the log outputs (partially because they don't convey a lot of information other than the URLs, and because these calls seem to be made over and over and over in a loop, and seeing that it's polling RCv3 and alone is not necessarily helpful in reading test output).

cyli:

cyli:

cyli:

cyli:

autoscale_cloudcafe/autoscale/client.py
cyli: Possibly also some of the other prints?

cyli: Just to be clear, this is non blocking.  If all the tests pass this PR is getting merged.  Just wondering if this could be a future change.

cyli:
Ok, ran the full suite of integration tests.  The only failure was a [spurious schedule policy test failure](https://github.com/rackerlabs/otter/issues/958).

Lint now passes, so I think this is probably good to go and any further changes can happen in later PRs.

PR #862, opened 2014-12-24
Grab the correct ServiceNet IP

### LINKED ISSUE #740: Make sure we add the right IP address to the load balancer
Discovered https://github.com/rackerlabs/otter/pull/737#discussion-diff-19557780 that there can be multiple IPv4 addresses under 'private in server details.

Since we want to add only the ServiceNet address to the load balancer, we should make sure we have correctly identified the ServiceNet address.

Repasting relevant part of the discussion here:

There is only 1 IPv4 address for service net, which is what we add to the load balancers - that's a 10.x.x.x address.  If the user wanted other private networks (badly named), it's named the same thing as what the user specified.  So if the user added the server also to a private network named MySuperAwesomeNetwork, the addresses would look like:

    "private": [{"addr": "10.0.0.1", "version": 4}],
    "MySuperAwesomeNetwork": [{"addr": "192.168.4.3", "version": 4}, {"addr": "::::", "version 6"}],
    "public": [{"addr": "50.50.50.50", "version": 4}, {"addr": "::::", "version 6"}]

However, if the user chooses to name their private network 'private' (which is allowed), then what shows up is indeed what you have here, with the 2 IPv4 addresses listed under 'private', and the servicenet address (which we want to add to the load balancer) is not necessarily the first one.  For instance:

    "private": [{"addr": "192.168.3.1"}, {"addr": "10.0.0.1", "version": 4}],
    "public": [{"addr": "50.50.50.50", "version": 4}, {"addr": "::::", "version 6"}]

So if we want to add only the servicenet address, we might want to check that the address is of the form 10.x.x.x.  :|

lvh:
This closes #740.


cyli:
I think this may need to have master merged in and conflicts resolved, but after that please merge away!

Issue #821, opened 2014-12-11
RackConnectV3 convergence


manishtomar:
Handle RackConnectV3 load balancer pools in convergence. This can be split up in following tasks:
- [x] Define model to represent RC3 LB pool and its nodes
- [ ] Gather RCv3 load balancer pools and nodes in above model 
- [x] Take it as input to `converge` and return steps needed to add it to the pool
- [x] [Update _lb_configs_from_metadata](https://github.com/rackerlabs/otter/pull/1102#discussion_r24850105)

lvh:
We could really use these model types for building integration tests.

lvh:
I added the [Update _lb_configs_from_metadata](https://github.com/rackerlabs/otter/pull/1102#discussion_r24850105) requirement to the description.

lvh:
On second thought, this ticket probably shouldn't exist. There's a milestone for that. I'll rip the requirements apart and turn them into separate tickets.

PR #872, opened 2014-12-29
Generic LB model (interfaces) for converger (mergable without queuing).

cyli:
Some generic LB model interfaces (and specific CLB interface providers) so that the converge logic can operate on general LBs (part of fixing #821).

Here is [a sketch of what planning will look like (not full changes, no tests)](https://github.com/rackerlabs/otter/blob/a70fc77f226905422eaac91d0c8412394a18273f/otter/convergence/planning.py) - mainly, look at [_remove_from_lb_with_draining](https://github.com/rackerlabs/otter/blob/a70fc77f226905422eaac91d0c8412394a18273f/otter/convergence/planning.py#L20) and [_converge_lb_state](https://github.com/rackerlabs/otter/blob/a70fc77f226905422eaac91d0c8412394a18273f/otter/convergence/planning.py#L68).  (Will be separate PRs)

There will also be specific RCv3 implementations of ILBDescription and ILBNode (the RCv3 version may not be drainable). (Again, future PRs.)

Also looking to implement `AddToLoadBalancing` etc. as a function which checks what the types are and produces the right specific steps (in future PR).  Something like:

    def AddServerToLoadBalancing(server, description):
        if isinstance(description, CLBDescription):
            return AddNodesToLoadBalancer(description.lb_id, (server.servicenet_address, description))

Naming corrections/feedback about above welcome and requested :)

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -102,7 +102,7 @@ class NovaServer(object):
 
 
 @attributes(['launch_config', 'desired',
-             Attribute('desired_lbs', default_value=(), instance_of=tuple),
+             Attribute('desired_lbs', default_factory=dict, instance_of=dict),
              Attribute('draining_timeout', default_value=0.0, instance_of=float)])
 class DesiredGroupState(object):
     """
@@ -110,7 +110,8 @@ class DesiredGroupState(object):
 
     :ivar dict launch_config: nova launch config.
     :ivar int desired: the number of desired servers within the group.
-    :ivar tuple desired_lbs: A tuple of :class:`ILBDescription` providers.
+    :ivar dict desired_lbs: A mapping of load balancer IDs to lists of
+        :class:`LBConfig` instances.
     :ivar float draining_timeout: If greater than zero, when the server is
         scaled down it will be put into draining condition.  It will remain
         in draining condition for a maximum of ``draining_timeout`` seconds
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -9,6 +9,9 @@ from pyrsistent import freeze
 
 from twisted.python.constants import Names, NamedConstant
 
+from zope.interface import implementer, Interface
+from zope.interface import Attribute as IAttribute
+
 
 class CLBNodeCondition(Names):
     """Constants representing the condition a load balancer node can be in"""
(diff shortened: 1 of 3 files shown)


manishtomar:
> Another review question - luckily CLBDescription and LBConfig are pretty similar - wondering if we should decouple the object from the AddNodeToCLB step though, since who knows if it may need change again in the future.

I thought `LBConfig` will be replaced by `CLBDescription`. If so, then `AddnodeToCLB` should take `CLBDescription` in `address_config`.

cyli:
@manishtomar Yes, it will (see https://github.com/rackerlabs/otter/compare/generic_lb_models...use_generic_lb_models), it does take CLBDescription.  I'm just thinking if it this changes in the future.  I guess there's no hurry to make that change.

cyli:
To clarify the above discussion with @manishtomar, basically I was suggesting making the step take a `dict` instead of an object.  But super unimportant, was just an idle thought.

Here is a sketch of what the RCv3 models will look like, and some of the CLB planning steps:

https://gist.github.com/cyli/0303219c930ff86cbd74

manishtomar:
Thanks for the gist @cyli . This LGTM :)

manishtomar:
`AddNodeToLoadBalancing` should be renamed to `AddNodeToLB` ?

cyli:
@manishtomar Do you mean [`AddNodesToCLB` (already in master, it's the renamed CLB step)](https://github.com/rackerlabs/otter/blob/master/otter/convergence/steps.py#L80), or do you mean to just shorten the name to `AddNodeToLB`?  Mostly I left it long to distinguish it from CLB, but happy to change also.

manishtomar:
> do you mean to just shorten the name to AddNodeToLB?

Yes. Also I wasn't clear the reason for `LoadBalacing` instead of `LoadBalancer`

cyli:
@manishtomar Because for RCv3 they're technically load balancing pools, and not load balancers, so I wanted to denote some kind of load balancing entity.

I would like to make sure there's enough of a visual difference between `AddNodeToLB` and `AddNodesToCLB` though - how do you feel about changing `AddNodesToCLB` to `AddAddressesToCLB`?

manishtomar:
> @manishtomar Because for RCv3 they're technically load balancing pools, and not load balancers, so I wanted to denote some kind of load balancing entity.

In that case, I would suggest `LoadBalancingEntity` since `LoadBalacing` seems more like a verb than a noun. 

> how do you feel about changing AddNodesToCLB to AddAddressesToCLB?

Sounds good.

PR #874, opened 2014-12-30
Use generic lb models from

cyli:
This is based off of #872 - this actually removes the current `LBNode` and `LBConfig` and uses `CLBNode` and `CLBDescription` in the current code.  This is part of starting on issue https://github.com/rackerlabs/otter/issues/821 (RCv3 in convergence).

Future work (other PRs) left to do after this:  changing the converge logic to operate on generic `ILBNode`s and `ILBDescription`s.

Review question: wondering if we should change AddNodeToCLB to just take dicts of useful information, so if we change the underlying model again we don't have to change it to take that model instead (luckily in this case `LBConfig` and `CLBDescription` are almost the same).

These are the actual changes between #872 and this:  https://github.com/rackerlabs/otter/compare/generic_lb_models...use_generic_lb_models


manishtomar:
`DesiredGroupSate.desired_lbs` should probably be renamed to `desired_clbs` or even better `clb_descriptions` as the prefix `desired` is redundant.

cyli:
@manishtomar WRT to naming `desired_lbs` -> `desired_clbs`, am hoping the list will have both CLBs and RCv3 load balancers, so the converge logic will work on both.

Cool with naming them `lb_descriptions` though.

manishtomar:
If you want to have both LB configs, then `DesiredGroupState.lb_descriptions` should be solely based on `ILBDescription` and `ILBNode` which means it should just be list of `ILBDescriptions`

cyli:
@manishtomar Totally agree - I was going to do that in a different PR (when we implement the generic logic) - this was just a straight up regex replace.  :)

manishtomar:
LGTM :)

Issue #871, opened 2014-12-28
Improve automated code quality tools


lvh:
Our automated code quality checking is less than perfect:
- The lint standards are underspecified, so it's not clear what "lint" is.
- We're not using the best tooling available. 
- The tooling we use is outdated.

Furthermore:
- We don't check code coverage as part of our CI process.
- We don't check docstrings for quality at all.

Fixing this requires several steps:
- [x] Make a CONTRIBUTING.md that's unambiguous about which standards we follow: #914
- [x] Use flake8 + flake8-import-order: #877 + #906
- [ ] Use pep8-naming: #907
- [x] Check if packages are outdated as part of the linting process (shouldn't fail the build) (#877)
- [x] Fail on reduced test coverage. The easiest way to do this actually probably involves waiting until we run the tests on Travis with Mimic, and then using coveralls or something.
- [ ] Use doc8 + pep257 + sphinx-build + sphinxcontrib-spelling + pyechant (this means Sphinx docs with all the API docs in it). Alternatively: pylint; although apparently that's not smart enough to know about docstring standards.

PR #877, opened 2014-12-30
Linter improvements

lvh:
This resolves parts of #871. This also removes some pinned deps of deps, because terrarium is gone.

This should be merged after #894.

This is not really doing everything:

- Docstrings aren't being tested basically because pep257 is not yet in a place were we can configure it, but that's going to be resolved Soon(TM; at least whenever they cut a release. This is #897.
- ~~iso8601 can't be upgraded until #891 which may or may not happen in this PR~~ iso8601 got upgraded in this PR
- ~~klein can't be upgraded until #889~~ klein did a new release but we're still using some implementation details, see #889


lvh:
This appears to be failing because Jenkins doesn't have `libffi` headers.

radix:
this is failing on some linter stuff, is it expected? If this PR is intended to be in-progress could you label it as such?

lvh:
I was expecting it to pass, and thought I did label it as such. Sorry.

lvh:
OK, so:

- Some pep257 checks are bogus and no longer actually in pep257, specifically `D203: Expected 1 blank line *before* class docstring, found 0`
- pep257 isn't picking up my config files, and I can't figure out why. none of the names it claims to support actually work.
- there's a verbose flag which causes it to log stuff like which config files it's looking at and why, but it's only in 0.4.0
- 0.4.0 hasn't been released
- The commit that tags 0.4.0 is broken.

As a result, I am going to get rid of flake8-docstrings for now, run pep257 manually (command line tricks do work :-)), and create a ticket to be revisited at some point.

lvh:
FYI, the review tag wasn't a waffle-induced mistake, this is now actually ready for review.

cyli:
Apologies, I think I have forgotten the outcome of this discussion.  Are we actually going to pin dependencies of dependencies anywhere for deploys?  I vaguely remember that we thought it'd be a good thing, just maybe not in the requirements file?

lvh:
I reverted the "Remove unitgration" stuff.

lvh:
@cyli I thought the only reason we transiently pinned dependencies right now was terrarium, which is gone anyway? What's the problem we're trying to solve by doing that?

lvh:
(If it's "I want some guarantees that the same software runs on prod as in staging", that's totally valid, but I'd suggest Docker is the real way you fix that one)

cyli:
@lvh:  Pretty much what you said - I would like to make sure that we test the exact versions that get shipped to prod.  I totally agree it's better to do that in our deploy, and we should work toward that, but we're not quite there yet.  :)  

I'm fine with "temporarily put pinned dependencies in the deploy repo instead of here" too.  

It might also be nice to have a record of "what pinned dependencies have we actually tested with and guarantee works" in this repo - I don't feel particularly strongly either way about the interim solution.

lvh:
@cyli Yep. Docker is going to fix this one properly; I think it's fine not to transitively pin until then, and the benefits outweigh the costs.

cyli:
@lvh Fair enough

Makefile
```diff
@@ -40,8 +39,16 @@ run:
...
 
-lint:
-	${PYTHONLINT} ${PYDIRS}
+lint: listoutdated flake8diff
+
+listoutdated:
+	pip list --outdated --allow-external=cafe,cloudcafe
+
+flake8diff:
+	git diff --patch --no-prefix origin/master | flake8 --diff
```
cyli: 1.  Is there a config file or something that specifies the maximum line length?  `pep8` is 80, and we were using 105 before.

2.  Doesn't `git diff --patch --no-prefix origin/master` mean that if your branch is behind master, it'd also lint all the files that are in master but not in your branch yet?  This shouldn't matter as we move forward and start using this lint style, but for instance if you run `make lint` now (there are a bunch of errors from changes that have been merged to master but not this branch).

lvh: 1. Nope. That could be added, though. (I think not having it at 105 is an improvement.)
2. Yep. That's intentional. The full lint on master is just too big to fix in one PR. (run `make flake8full` to get an idea)

cyli:
> Yep. That's intentional. The full lint on master is just too big to fix in one PR. (run make flake8full to get an idea)

Right, I understand that we don't want to change all of master at once.  Just wondering if we want `make lint` to run only on the changes in the current branch, not on both the changes in the current branch + the changes in github master that are not currently in this branch.  So only lint the changes since where the branch diverged, not `origin/master`.

Basically instead of `git diff --patch --no-prefix origin/master`, which currently also includes integration tests changes that are on origin/master but not this branch, we can use something like `git diff --patch --no-prefix $(git merge-base master $(git rev-parse --abbrev-ref HEAD))`, or `git diff --patch --no-prefix master...$(git rev-parse --abbrev-ref HEAD)`.

I think even `git diff --patch --no-prefix origin/master...$(git rev-parse --abbrev-ref HEAD)` works.

That only lints the changes that appear in this PR.

Makefile
cyli: > Yep. That's intentional. The full lint on master is just too big to fix in one PR. (run make flake8full to get an idea)

Right, I understand that we don't want to change all of master at once.  Just wondering if we want `make lint` to run only on the changes in the current branch, not on both the changes in the current branch + the changes in github master that are not currently in this branch.  So only lint the changes since where the branch diverged, not `origin/master`.

Basically instead of `git diff --patch --no-prefix origin/master`, which currently also includes integration tests changes that are on origin/master but not this branch, we can use something like `git diff --patch --no-prefix $(git merge-base master $(git rev-parse --abbrev-ref HEAD))`, or `git diff --patch --no-prefix master...$(git rev-parse --abbrev-ref HEAD)`.

I think even `git diff --patch --no-prefix origin/master...$(git rev-parse --abbrev-ref HEAD)` works.

That only lints the changes that appear in this PR.

UPDATE: note I put this in the wrong place before, deleted the other comment and moved it here instead.

cyli:
Would it make sense to add the new make targets to the README?

Makefile
lvh: That's definitely an improvement. I had something similar at first, but then realized that I wanted the current staging area linted as well. I suppose it's better to just do it in two steps instead :)

lvh:
@cyli Thanks for your comments! I've documented the targets, and added both your suggestion and something that specifically diffs between HEAD and tree.

cyli:
This LGTM!  I'd just check with @radix, @manishtomar about the 80-line thing.  :)

radix:
105 4 lyfe

manishtomar:
I am fine with 80-char line :) We can definitely start with it and worst case revert if its sounds too difficult.

lvh:
Awesome!

PR #859, opened 2014-12-31
Accept string CLB ID in loadBalancerId in launch config

cyli:
Launch config's `loadBalancerId` has been taking int before RCv3 since it only took ID of CLB which is an integer. Now it is also taking RCv3 LB pool ID in the same field which is a UUID. This change of types broke Reach. So, otter should accept CLB ID in `loadBalancerId` as string. But also allow int to keep it backwards compatible.


manishtomar:
LGTM again :)

Issue #868, opened 2014-12-26
Blow up somehow if there is no ServiceNet IP and a CLB is configured 


cyli:
The bug discovered https://github.com/rackerlabs/otter/pull/862 seems to indicate that if the server was created without a ServiceNet IP, (or without a valid one) then no failure occurs - it just doesn't get added to any CLBs.

Possibly now, a blank IP will be attempted to be added to a CLB.  Possibly a failure should occur instead - at the very least, an error log message.

Based on discussions, this should have several tasks attached:
- [x] Validate that a user cannot submit a launch config that has CLBs without ServiceNet
- [ ] Log in case we encounter a server that should be on a CLB and no ServiceNet
  - Now that we are not moving servers to new load balancers, we shouldn't have an issue with old servers and new load balancers, but the user can still remove servicenet from their servers at any point
- [ ] Possibly tag the server in metadata?

manishtomar:
Can this be closed now that #879 is merged?

cyli:
Probably the companion piece to #879 is to log in convergence if we encounter an old server without service net and that we are just going to give up on it.

I think we can probably close once that is in, also?

lvh:
We probably *also* want to log that in the server's metadata somehow (can be generic if needed) so that when the fine day comes that we do automated rolling upgrades, we can hit the bad servers first. (Arguably, we could just kill the server anyway, since it's not working...)

cyli:
Updated the description since, now that we are doing per-server load balancer configs and not just moving every server to new load balancers, we won't encounter the old server -> load balancer issue.

But it is still possible for the user to break things by manually removing servicenet, so we should probably still log.

PR #879, opened 2014-12-31
Validate launch config to require servicenet with CLBs

cyli:
This is part of addressing #868, to prevent users from putting in bad data in the first place.

I am not sure it's worth splitting the json schema into different versions (CLB and no CLB), since our schema is getting pretty complicated already, so I made this a function.  Also, this makes for a nicer user error message.

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/functional/launch_config/test_launch_config_negative.py
+++ b/autoscale_cloudroast/test_repo/autoscale/functional/launch_config/test_launch_config_negative.py
@@ -141,6 +141,53 @@ class LaunchConfigNegtaiveTest(AutoscaleFixture):
                               msg='Create group with invalid flavor id was successsful with'
                               ' response {0}'.format(create_group_response.status_code))
 
+    def test_create_scaling_group_with_CLBs_but_no_ServiceNet(self):
+        """
+        Scaling group creation fails with a 400 when a launch config has one or
+        more CLBs configured, but no ServiceNet configured.
+        """
+        create_group_response = self.autoscale_client.create_scaling_group(
+            gc_name='test_no_servicenet',
+            gc_cooldown=self.gc_cooldown,
+            gc_min_entities=self.gc_min_entities,
+            lc_name=self.lc_name,
+            lc_image_ref=self.lc_image_ref,
+            lc_flavor_ref=self.lc_flavor_ref,
+            lc_networks=[
+                {"uuid": "4ebd35cf-bfe7-4d93-b0d8-eb468ce2245a"},
+                {"uuid": "00000000-0000-0000-0000-000000000000"}
+            ],
+            lc_load_balancers=[{'loadBalancerId': '1234', 'port': 80,
+                                'type': 'CloudLoadBalancer'}])
+        self.assertEquals(
+            create_group_response.status_code, 400,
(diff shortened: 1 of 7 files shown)


cyli:
Going to add some functional tests...

autoscale_cloudroast/test_repo/autoscale/functional/launch_config/test_launch_config_negative.py
```diff
@@ -141,6 +141,53 @@ def test_create_scaling_group_invalid_flavorid(self):
...
+            lc_flavor_ref=self.lc_flavor_ref,
+            lc_networks=[
+                {"uuid": "4ebd35cf-bfe7-4d93-b0d8-eb468ce2245a"},
+                {"uuid": "00000000-0000-0000-0000-000000000000"}
+            ],
+            lc_load_balancers=[{'loadBalancerId': '1234', 'port': 80,
+                                'type': 'CloudLoadBalancer'}])
+        self.assertEquals(
+            create_group_response.status_code, 400,
+            msg=('Create group with no CLBs but no ServiceNet was successsful '
```
lvh: I think the message here has a typo: group with *no* CLBs?

autoscale_cloudroast/test_repo/autoscale/functional/launch_config/test_launch_config_negative.py
```diff
@@ -141,6 +141,53 @@ def test_create_scaling_group_invalid_flavorid(self):
...
+                {"uuid": "4ebd35cf-bfe7-4d93-b0d8-eb468ce2245a"},
+                {"uuid": "00000000-0000-0000-0000-000000000000"}
+            ],
+            load_balancers=[{'loadBalancerId': '1234', 'port': 80,
+                             'type': 'CloudLoadBalancer'}])
+
+        status = update_launch_config_response.status_code
+        self.assertEquals(
+            status, 400,
+            msg=('Create group with no CLBs but no ServiceNet was successsful '
```
lvh: Same here

otter/json_schema/group_schemas.py
```diff
@@ -321,6 +322,21 @@
 }
 
 
+def validate_launch_config_servicenet(lc):
+    """
+    Validate that if CLBs are provided, ServiceNet is also provided.
+    """
+    clb = any([lb.get('type', 'CloudLoadBalancer') == 'CloudLoadBalancer'
+               for lb in get_in(('args', 'loadBalancers'), lc, default=())])
+    servicenet = get_in(('args', 'server', 'networks'), lc, default=None)
```
lvh: I'm not sure I understand the naming here. Isn't this all networks, and then you're checking if it's actually ServiceNet?

lvh: (Because if this is None, ServiceNet will definitely be available (by default), otherwise, ServiceNet has to be in this list, right?)

lvh:
LGTM! Thanks for working on this; I think this is a wonderful safety feature :-) There's some comments inline, but they're just naming nitpicks.

otter/json_schema/group_schemas.py
cyli: Yes, sorry, the naming should be `networks` - the next line lets it go if it's None.

PR #880, opened 2015-01-02
Force group delete fails sometimes

cyli:
Test `test_system_force_delete_group_with_minentities_over_zero` in https://as-jenkins.rax.io/job/otter-cloudcafe-prod-dfw-slow-system-tests/507/ failed because `DELETE ...groups/<groupId>/?force=TRUE` returned 403 instead of 204. This should never happen. It should always succeed. The most likely reason is that `update_config` to set min/max to 0 is done with consistency ONE and `obey_config_change` after that didn't notice config change group deletion. I think the right solution would be to change default consistency to QUORUM. There is already facility to do that.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -217,7 +217,8 @@ _consistency_levels = {'event': {'fetch': ConsistencyLevel.QUORUM,
                                  'insert': ConsistencyLevel.ONE,
                                  'delete': ConsistencyLevel.QUORUM},
                        'group': {'create': ConsistencyLevel.QUORUM},
-                       'state': {'update': ConsistencyLevel.QUORUM}}
+                       'state': {'update': ConsistencyLevel.QUORUM},
+                       'partial': {'update': ConsistencyLevel.QUORUM}}
 
 
 def get_consistency_level(operation_name, resource_name,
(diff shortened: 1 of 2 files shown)


lvh:
LGTM! Also the idea of gradually moving things to quorum sounds good, but maybe there should be two metrics in place first:

1. Whatever it is we're trying to fix by having it not be QUORUM
2. How much QUORUM is costing us

manishtomar:
> Whatever it is we're trying to fix by having it not be QUORUM

These kind of silent bugs that fail on rare occasions. Basically, cases where immediate read after write fails. This could be a transaction that has write->read in the process or it could be a client (like test) that say creates a policy and gets it immediately. We haven't seen this as the time it takes for any create response to reach client the created resource is replicated on cass nodes. However, the transaction situation does happen from time to time. We may want to identify all these cases before blindly changing to QUORUM.

> How much QUORUM is costing us

We can easily measure it https://dev-otterlogs.fedsso.rackspace.com/#/dashboard/elasticsearch/Cass%20query%20times.

cyli:
@manishtomar @lvh Just to be clear, should I mark this as pending-merge?

manishtomar:
From my side: yes. Pending merge.

lvh:
Yeah, this looks fine to me. I'll change the tags.

cyli:
Closing this in favor of #905.  If that gets reverted, this should probably be re-opened.

PR #886, opened 2015-01-06
Config safety checks!

cyli:
Need to ensure config.json is never checked in with creds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,14 @@ CONTROL_KEYSPACE ?= OTTER
 REPLICATION_FACTOR ?= 3
 CLOUDCAFE ?= $(shell which cafe-runner)
 
+mkfile_dir := $(shell dirname "$(MAKEFILE_LIST)")
+
 .PHONY: targets env-precheck
 
 hooks:
-	echo "#!/bin/bash" > .git/hooks/pre-commit
-	echo "python `pwd`/scripts/config_check.py" >> .git/hooks/pre-commit
-	echo "`pwd`/scripts/python-lint.py --git-precommit" >> .git/hooks/pre-commit
+	cp ${mkfile_dir}/scripts/config_check.py ${mkfile_dir}/.git/hooks
+	echo "#/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
+	echo "python .git/hooks/config_check.py" >> ${mkfile_dir}/.git/hooks/pre-commit
 
 targets:
 	@cat README.md
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,11 @@ CLOUDCAFE ?= $(shell which cafe-runner)
 
 .PHONY: targets env-precheck
 
+hooks:
+	echo "#!/bin/bash" > .git/hooks/pre-commit
+	echo "python `pwd`/scripts/config_check.py" >> .git/hooks/pre-commit
+	echo "`pwd`/scripts/python-lint.py --git-precommit" >> .git/hooks/pre-commit
+
 targets:
 	@cat README.md
 
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ mkfile_dir := $(shell dirname "$(MAKEFILE_LIST)")
 
 hooks:
 	cp ${mkfile_dir}/scripts/config_check.py ${mkfile_dir}/.git/hooks
-	echo "#/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
+	echo "#!/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
 	echo "python .git/hooks/config_check.py" >> ${mkfile_dir}/.git/hooks/pre-commit
 	chmod a+x ${mkfile_dir}/.git/hooks/pre-commit
 
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ hooks:
 	cp ${mkfile_dir}/scripts/config_check.py ${mkfile_dir}/.git/hooks
 	echo "#/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
 	echo "python .git/hooks/config_check.py" >> ${mkfile_dir}/.git/hooks/pre-commit
+	chmod a+x ${mkfile_dir}/.git/hooks/pre-commit
 
 targets:
 	@cat README.md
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -18,6 +18,11 @@ CLOUDCAFE ?= $(shell which cafe-runner)
 
 .PHONY: targets env-precheck
 
+hooks:
+	echo "#!/bin/bash" > .git/hooks/pre-commit
+	echo "python `pwd`/scripts/config_check.py" >> .git/hooks/pre-commit
+	echo "`pwd`/scripts/python-lint.py --git-precommit" >> .git/hooks/pre-commit
```
lvh: 1. Anyone who can change config_check.py now can screw with everyone's git hooks. Have you considered copying the script?
2. `python-lint.py` is pretty slow, but I guess it's probably OK.
3. `--git-precommit` doesn't work anymore with the outstanding PR to update the linters.

Makefile
```diff
@@ -18,6 +18,11 @@ CLOUDCAFE ?= $(shell which cafe-runner)
 
 .PHONY: targets env-precheck
 
+hooks:
+	echo "#!/bin/bash" > .git/hooks/pre-commit
+	echo "python `pwd`/scripts/config_check.py" >> .git/hooks/pre-commit
```
glyph: You don't need to interpolate ``` `pwd` ``` here; git hooks are run from the root of the working tree.  (I can't find an official doc reference for this but experimentally the versions of git that I tried on ubuntu OS X, etc, did so.)

glyph: consider

```
omgwtf := $(shell dirname "$(MAKEFILE_LIST)")
```

Makefile
cyli: Removed `python-lint.py` and now `config_check.py` is copied to the git hooks directory and run from there.

Makefile
```diff
@@ -16,8 +16,15 @@ CONTROL_KEYSPACE ?= OTTER
...
 CLOUDCAFE ?= $(shell which cafe-runner)
 
+mkfile_dir := $(shell dirname "$(MAKEFILE_LIST)")
+
 .PHONY: targets env-precheck
 
+hooks:
+	cp ${mkfile_dir}/scripts/config_check.py ${mkfile_dir}/.git/hooks
+	echo "#/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
+	echo "python .git/hooks/config_check.py" >> ${mkfile_dir}/.git/hooks/pre-commit
```
glyph: I might just replace this with a `ln -s` and put a `#!/usr/bin/env python` shebang and executable flag on config_check.py itself; fewer files to keep updated.

cyli: @lvh specifically advised against linking because he didn't want someone to just change `config_check.py` to disable stuff.  Granted, modifying `config_check.py` and then running make hooks is the same, but I guess more steps to break things.

glyph:
Hmm. I don't think this works as desired:

glyph:
```
$ add -f config.json 
-bash: add: command not found
$ git add -f config.json 
$ make hooks
cp ./scripts/config_check.py ./.git/hooks
echo "#/bin/bash" > ./.git/hooks/pre-commit
echo "python .git/hooks/config_check.py" >> ./.git/hooks/pre-commit
$ git ci -m yay
[config_safety 27a9f0e] yay
 1 file changed, 63 insertions(+)
 create mode 100644 config.json
$ 
```

Makefile
cyli: Also `--no-verify` would also disable the check, but I think this was a slight precaution against accidental breakage.

glyph:
Aah. And here's the fix:

```
$ chmod a+x .git/hooks/pre-commit
$ git uncommit
$ git add -f config.json 
$ git ci -m yay
  File ".git/hooks/config_check.py", line 38, in <module>
    raise Exception("DO NOT commit 'config.json'")
Exception: DO NOT commit 'config.json'
```

Makefile
glyph: @lvh - by "to disable stuff" you mean to disable the hook locally?

cyli:
@glyph why does this work?

```
$ sed -ie "s|REPLACE_WITH_REAL_USERNAME|username|g" config.example.json 
$ git add config.example.json
$ git ci -m yay
  File ".git/hooks/config_check.py", line 42, in <module>
    look_for_password(json.load(fp))
  File ".git/hooks/config_check.py", line 33, in look_for_password
    look_for_password(val)
  File ".git/hooks/config_check.py", line 29, in look_for_password
    raise Exception('Password values should always be '
Exception: Password values should always be "REPLACE_WITH_REAL_USERNAME"
```

glyph:
@cyli did you manually `chmod` the hook on your machine?

cyli:
@glyph no

glyph:
@cyli Is it executable?

cyli:
@glyph yes

cyli:
@glyph I guess git maybe changes it?

glyph:
@cyli Perhaps you already had a hook initialized there?

```
$ touch x
$ ls -l x
-rw-r--r--  1 glyph  staff  0 Jan  5 19:10 x
$ chmod a+x x
$ ls -l x
-rwxr-xr-x  1 glyph  staff  0 Jan  5 19:10 x
$ echo new line > x
$ ls -l x
-rwxr-xr-x  1 glyph  staff  9 Jan  5 19:10 x
```

my version of git definitely doesn't seem to change it by default.

cyli:
@glyph I did, but I removed it before running make hooks

glyph:
@cyli possibly your shell's default `umask` is different from mine because of init files, not really sure

glyph:
ok lgtm

Makefile
```diff
@@ -16,8 +16,16 @@ CONTROL_KEYSPACE ?= OTTER
 REPLICATION_FACTOR ?= 3
 CLOUDCAFE ?= $(shell which cafe-runner)
 
+mkfile_dir := $(shell dirname "$(MAKEFILE_LIST)")
+
 .PHONY: targets env-precheck
 
+hooks:
+	cp ${mkfile_dir}/scripts/config_check.py ${mkfile_dir}/.git/hooks
+	echo "#/bin/bash" > ${mkfile_dir}/.git/hooks/pre-commit
```
lvh: This does not look like a shebang?

lvh:
LGTM!

PR #888, opened 2015-01-07
Simplify encoding/decoding

### LINKED ISSUE #827: add_bind_root randomly coerces to bytes with utf-8 and ascii
Figure out which one it has to be and why, or why we even have to do this right now. Maybe a better resolution would be to not stuff unicode in?

Original ask: https://github.com/rackerlabs/otter/pull/822#issuecomment-66709604

Code got copied a bunch of times, no-one remembers why, so the answer is probably "totally arbitrarily".

I think it should probably be ascii, I don't know if everything down the chain knows about punycode, and if it does, it should probably be taking `unicode` not `str`.

lvh:
Allegedly closes #827.

### DIFF
--- a/otter/util/pure_http.py
+++ b/otter/util/pure_http.py
@@ -230,14 +230,9 @@ def add_bind_root(root, request_func):
     The URL given is expected to be quoted if required. This decorator does not
     quote the URL.
     """
-    if isinstance(root, unicode):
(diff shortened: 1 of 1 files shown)


radix:
+1

lvh:
I added an empty commit just to close the issue when this gets merged.

PR #894, opened 2015-01-08
Kill unitgration

lvh:
I don't even know.

This needs to be merged before #877.


cyli:
LGTM

lvh:
@cyli: Are you comfortable with this as-is, or would you like some schema script tests first? I can't remember if you mentioned that after or before your +1 :-)

cyli:
Apologies, the schema scripts tests thing came after the +1.  :(

In principle I'd prefer to have script tests first because it'd be nice to not have a lapse in coverage of at least being able to load up our schema in Cassandra.  And while I think we're not looking to make any changes in the very near future, we are looking changing some CI infrastructure and the scripts test could also serve to highlight misconfigurations there.

You're pretty great about finishing TODO items like that really quickly, though, so I don't feel super strongly and am fine with merging this first if there is a good argument for doing so.

cyli:
Also think I remember @manishtomar saying he wanted to think about this some more, and the schema tests might make sense to do in that interim.

manishtomar:
This should be merged after #901 is implemented.

lvh:
So, this is basically too much pain to do right now if I have to wait for preprod to work, so I'm instead retracting this PR.

PR #895, opened 2015-01-08
pretty health json

manishtomar:
This is useful when manually curling it

### DIFF
--- a/otter/test/rest/test_application.py
+++ b/otter/test/rest/test_application.py
@@ -554,7 +554,7 @@ class HealthCheckTestCase(RestAPITestMixin, TestCase):
         self.root = otter.app.resource()
 
         resp = self.assert_status_code(200)
-        self.assertEqual(resp, json.dumps(jsonobj, indent=4))
+        self.assertEqual(resp, json.dumps(jsonobj, indent=4, sort_keys=True))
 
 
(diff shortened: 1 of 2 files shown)


radix:
looks good :)

otter/test/rest/test_application.py
```diff
@@ -545,14 +545,16 @@ def test_health_check_endpoint_calls_store_health_check(self):
...
         def health_check():
-            return succeed({'blargh': 'boo'})
+            return succeed(jsonobj)
 
         otter = Otter(self.mock_store, 'region', health_check)
         self.root = otter.app.resource()
 
         resp = self.assert_status_code(200)
-        self.assertEqual(resp, json.dumps({'blargh': 'boo'}))
+        self.assertEqual(resp, json.dumps(jsonobj, indent=4))
```
radix: Oh, actually, there may be room for a spurious error in the unit test, since the order of keys in a json object is not sorted by default.

You could either change the test to only have one key in the dictionary, or you could use sort_keys=True to the implementation and unit test. probably some other options too but that's what comes to me now.

manishtomar: Good catch. Thanks!

radix:
+1 LGTM

manishtomar:
I'll test it once on dev vm and merge it.

PR #899, opened 2015-01-10
Taking identity credentials from file

manishtomar:
and deleting after reading from it. One consequence of this change is that `generate_authenticator` has visibly become impure causing unfortunate mocking while testing. I _could_ have returned `Effect` instead but that would've caused lot of changes in upstream functions using it. So, decided against it.


cyli:
Possibly one way to keep it pure is that perhaps when the config is read in the service making file, the service file can then read the password file and then pass that as a dict to the generator?  That way the service file would be doing the reading and deleting, as opposed to the generator.

lvh:
I like @cyli's suggestion, because it concentrates obvious places to grab some stuff from some config files.

Why don't we just delete the entire config? Having two files where some stuff is stored sounds like asking for trouble. Now that we've finally gone through the hassle of making sure `config.json` never accidentally gets committed, it doesn't seem like making sure `config.json` doesn't have any important information in it is the most reasonable next step.

manishtomar:
> I like @cyli's suggestion, because it concentrates obvious places to grab some stuff from some config files.

I like it too but I prefer reading from the file in one place and reading it when creating service will do that in 2 places. But I think it should be doable by having a util function. I'll think about it.

> Now that we've finally gone through the hassle of making sure config.json never accidentally gets committed, it doesn't seem like making sure config.json doesn't have any important information in it is the most reasonable next step.

I agree that we don't need this split after ensuring `config.json` is not uploaded. However, I think the root of the problem was having sensitive information in `config.json` in the first place and its probably bad practice to continue doing that.

lvh:
>> I like @cyli's suggestion, because it concentrates obvious places to grab some stuff from some config files.
> I like it too but I prefer reading from the file in one place and reading it when creating service will do that in 2 places. But I think it should be doable by having a util function. I'll think about it.

Yes.

>> Now that we've finally gone through the hassle of making sure config.json never accidentally gets committed, it doesn't seem like making sure config.json doesn't have any important information in it is the most reasonable next step.
>I agree that we don't need this split after ensuring `config.json` is not uploaded. However, I think the root of the problem was having sensitive information in `config.json` in the first place and its probably bad practice to continue doing that. 

What's the difference between having sensitive information in a file called `config.json` versus having it in a file called `credentials.json`, if you're going to delete them anyway?

manishtomar:
> What's the difference between having sensitive information in a file called config.json versus having it in a file called credentials.json, if you're going to delete them anyway?

The end result is the same during deploy. The distinction is during development. Using `credentials.json` keeps a clear distinction between the purposes of the two files. `config.json` is expected to grow as features are added. I expect multiple versions of it to be floating around in dev's machine and in worst case its ok if it gets committed. Its also difficult to do development if one has to keep recreating the file after each run. However, `credentials.json` will be separate and will encourage not having it in the first place. This is just matter of taste. I don't feel very strongly about it since there are safe-guards against committing it but I do feel it should be there as _practice_. If you don't find it particularly useful I can close this PR and we can have chef changes to delete config.json.

lvh:
Okay, so the problem you're trying to solve is developer happiness. That's a good problem to solve.

You're saying that always regenerating `config.json` is annoying because you have to keep recreating it. However, that wouldn't be fixed by having a second file, right? You'd still have to recreate `credentials.json`, too.

As for how you'd recreate the single `config.json` file that contains everything (including passwords), what's stopping a developer from doing:

```
cp config.whatever.json config.json; actually_run_otter
```

FWIW, there's no concern that any of these files that aren't named exactly `config.json` but which may have passwords in them will get committed, because [there's a pre-commit hook for that](https://github.com/rackerlabs/otter/blob/master/scripts/config_check.py).

It sounds like your argument hinges on developers having credentials versus not having them. I disagree that the file that contains them being called `credentials.json` vs `config.json` actually changes anything; people have them because they need to, and it doesn't really matter where you expect them to put it, right? Conversely, someone who does not need them should just not have them, regardless of what the file is named.

manishtomar:
@lvh OK. I think we can delete `config.json` after loading.

@cyli Do you think its ok to make changes to delete `config.json` after loading in chef?

cyli:
@manishtomar do you mean do the delete from chef?  

Chef can definitely delete the file, but it needs to make sure that it has actually finished loading the file and has started first.  Not quite sure how to do that - @shawnashlee?

cyli:
Oh wait hm... maybe we can put `$SHELL -c "sleep 10; rm /etc/otter.json"` in the the service init script?

lvh:
Recap from an IRL discussion we just had here:

- I'm not sure why you'd want to artificially create a window of time where the credentials are still available. There's one piece of code that clearly, unambiguously knows when it's done caring about that file, and it's in otter; so the delete should live there.
- While deleting makes a lot of sense for prod, it's maybe not the most obvious thing to do for dev, even thought it's fairly easy to work around (see snippet I posted above). The real security win is in rotating credentials anyway, so it doesn't actually *matter* if they get compromised.

So, TL;DR: not sure if should delete, but if we do delete, I'm sure it should be in otter.

shawnashlee:
could probably use a `notifies` in the runit_service block to trigger a bash script removing it

radix:
I haven't read all this but as a general rule I think it's always better to keep secrets as isolated as possible vs the general config file. config files should be safe to store in some VCS somewhere, and secrets should be somewhere else. I realize that when you auto-generate the config file on the host this argument is weakened, but I still think it makes sense -- for example, you may want to allow developers accounts to read and edit a config file but not the secrets file.

manishtomar:
Thanks @radix for expressing my thoughts clearly :). I was thinking of not failing if `credentials.json` is not found or specified. This way the `config.json` used by the developer can work without credentials.

manishtomar:
> I was thinking of not failing if credentials.json is not found or specified. This way the config.json used by the developer can work without credentials.

Actually I take that back since it can cause failure when scaling up/down and the developer would find out after some amount of debugging. 

@lvh Any change of opinion after @radix comment.

lvh:
Not really. Config *is* safe to store; that's why we have `config.example.json`. If otter has a dev-mode with autogenerated credentials, what's stopping it from using config.example.json (or `config.devvm.json`) by default?

In the long run, docker volume disconnection is probably a better way of doing it than unlinking. Even better, but slightly trickier: privilege dropping.

manishtomar:
I am closing this for now since deleting config file from otter alone is not enough. repose also should have this feature.

PR #900, opened 2015-01-12
Increase build timeout from 1 to 2 hours

lvh:
I also refactored the "get defaults from a function by name" behavior, since there are now 2 tests that use it.


manishtomar:
LGTM

PR #904, opened 2015-01-13
Allow tests to pass autoscale client a list of networks on creating a group.

cyli:
Just provide default networks if none are listed - do not overwrite the networks in the cloudcafe autoscale client if networks are provided.


lvh:
LGTM!

PR #905, opened 2015-01-13
Attempt to use quorum everywhere.

cyli:
Change the default consistency to QUORUM.

As per discussions with @lvh, if this is something that we want to make a feature of the system (consistency by default) then maybe it shouldn't be in the config?  Open to discussions.

GET requests being consistent would make CI (or scripts that create and then GET) more reliable.

Can be reverted as a whole if this causes problems.


otter/test/tap/test_api.py
```diff
@@ -399,35 +399,18 @@ def test_cassandra_store(self):
     def test_cassandra_scaling_group_collection_with_default_consistency(self):
         """
         makeService configures a CassScalingGroupCollection with a callable
-        that returns the default consistencies with the default exceptions,
-        if no other values are configured.
+        that returns the default consistencies with the default consistency
+        of one.
```
lvh: That consistency level is now QUORUM.

lvh:
LGTM modulo comment :)

manishtomar:
It makes sense to have always QUORUM for mutable data (i.e. all our configs/state) but I can imagine it be ONE for immutable data like scheduled events.

cyli:
@manishtomar I think the current exceptions for events are:  `QUORUM` for fetching the events, `ONE` for insert, and `QUORUM` for delete?

Does that still make sense, or should I add more?

lvh:
@manishtomar That's reasonable. I started digging, and discovered that `get_oldest_event` uses ` self.get_consistency('check', 'event')`, but `GetConsistencyTests` only makes assertions about `"fetch", "event"` (and asserts that it's quorum).

That looks like it might be a bug:

```
~/P/r/otter: ag \[\"\'\]check\[\"\'\],                                                                                                                                                              12:17:44
otter/models/cass.py
1377:                                    self.get_consistency('check', 'event'))
~/P/r/otter: ag \[\"\'\]fetch\[\"\'\],                                                                                                                                                              12:17:47
otter/models/cass.py
1354:                                    self.get_consistency('fetch', 'event'))

otter/test/models/test_cass_models.py
143:        level = get_consistency_level('fetch', 'event', ConsistencyLevel.ONE,
159:        level = get_consistency_level('fetch', 'event')
```

lvh:
Looks like both of those are used for getting events, so... resolution:

1. Change otter/models/cass.py L1377 to say `fetch`.
2. To actually address @manishtomar's comment, `test_cass_models.GetConsistencyTests.test_event_fetch` should use `ONE`, not `QUORUM`.

manishtomar:
@lvh `get_oldest_event` is expected to be ONE. It is only used during health check. Not when scheduling events. I'll look at the tests to ensure it asserts that.

cyli:
@manishtomar would it make sense then to add `get` of events as a special exception of ONE then?

cyli:
Also:  https://dev-otterlogs.fedsso.rackspace.com/#/dashboard/elasticsearch/Cass%20times%20by%20query  :|  No 99th percentile or trends, but we can probably check how things were this week compared to last week, etc.

lvh:
@manishtomar I'm not sure I understand. Are you saying the `fetch` vs `check` thing isn't a bug? Or that that test shouldn't check for consistency level `ONE`? Or neither?

cyli:
@lvh:  I think he means it isn't a bug because  'check' is only used for a health check, so that's fine as consistency one.  The tests do assert that check get_oldest_event uses consistency one.

cyli:
At least, that is the understanding currently reflected in the branch :D

lvh:
I thought he was saying that fetching events should be consistency `ONE` (that's the immutable, either extant or non-extant, thing, right?). Deleting should still be `QUROUM`, but it looks like [both are `DEFAULT_CONSISTENCY`](https://github.com/rackerlabs/otter/pull/905/files#diff-1bed9efd41b5b5053bae065c1dda62d5R1291), which I thought this PR changed to `QUORUM`.

manishtomar:
@lvh There are two functions to get events:
1.  `fetch_and_delete` which internally fetches and deletes as QUORUM. This could be argued to be ONE since its immutable. @cyli and I wanted to think a little more before making it ONE.
2. `get_oldest_event` which only _gets_ the oldest event for health check. This is ONE since it is not really concerned with consistency
Finally inserting is ONE. AFAIK, the tests for above all assert the expected consistencies.

cyli:
@lvh @manishtomar:  I'm a little concerned about fetch being ONE mainly because if DELETE is quorum and FETCH is one, depending on how long between fetch and deletes, it might be possible to fetch something that has been deleted, and thus execute it twice.  

Fetch grabs everything before the current time.  Insert can be one because at some point, SOMEONE will grab it, even if it's not immediately.

lvh:
@cyli Okay, that's fair; let's keep them both `QUORUM` then.

cyli:
@lvh: to clarify, by "them both" you mean fetch and delete?

lvh:
Yes, [this one and the one above it](https://github.com/rackerlabs/otter/pull/905/files#diff-1bed9efd41b5b5053bae065c1dda62d5R1291).

cyli:
@lvh cool, thanks.    

All: Up for review again then.  :)

lvh:
LGTM!

otter/models/cass.py
```diff
@@ -1435,7 +1372,7 @@ def get_counts(self, log, tenant_id):
         fields = ['scaling_group', 'scaling_policies', 'policy_webhooks']
         deferred = [self.connection.execute(_cql_count_for_tenant.format(cf=field),
                                             {'tenantId': tenant_id},
-                                            self.get_consistency('count', 'group'))
+                                            DEFAULT_CONSISTENCY)
```
manishtomar: This can be ONE since this operation can be expensive and we don't need to be accurate. However this function is not used anywhere.

manishtomar:
LGTM modulo that one comment :)

otter/models/cass.py
cyli: Good catch, thanks!  I'll just change it to ONE for now and add an issue to revisit this functionality.

lvh:
retest this please

cyli:
Have created #908 to address the issue with `get_counts`.  Also, what should I do about all the lint?

cyli:
retest this please

PR #906, opened 2015-01-14
Improve flake8-based linting

lvh:
- Previously didn't work on Jenkins, because it pulls the weird Github-created PR ref. The new code diffs against `HEAD^1` on Jenkins (which is the diff produced by the PR), and against the merge-base of the current branch vs master.
- flake8 configuration moved to setup.cfg.
- application-import-names added so that otter imports go separately from others.
- removed pep8 naming because it doesn't support the few camelCaseMethods we have (notably setUp, a bunch of Twisted methods, and probably some logger stuff in the tests).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,16 +44,19 @@ lint: listoutdated flake8diff
 listoutdated:
 	pip list --outdated --allow-external=cafe,cloudcafe
 
-
-# If building a PR on Jenkins, ghprbTargetBranch is the target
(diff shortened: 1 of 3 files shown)


lvh:
We could make this a bit better by making `master` configurable, but I don't think anyone regularly works against branches other than `master`?

radix:
LGTM. I've been testing this config while cleaning up a big PR.

PR #910, opened 2015-01-15
Clean up test_scheduler

lvh:
I tried cleaning up some of the test comments. It's not always just style; they don't always accurately communicate intent. We can't really just drop everything and fix every single test right now, so perhaps this should just be a collective effort for PRs in the future.

```
~/P/r/otter: make flake8full | wc -l                                                                                                                                                                18:34:50
make: *** [flake8full] Error 1
    3881
~/P/r/otter: git checkout master                                                                                                                                                                    18:34:54
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
~/P/r/otter: make flake8full | wc -l                                                                                                                                                                18:35:00
make: *** [flake8full] Error 1
    3963
```

### DIFF
--- a/otter/test/test_scheduler.py
+++ b/otter/test/test_scheduler.py
@@ -35,10 +35,11 @@ class SchedulerServiceTests(SchedulerTests, DeferredFunctionMixin):
     """Tests for `SchedulerService`"""
 
     def setUp(self):
-        """
-        mock all the dependencies of SchedulingService that includes logging,
-        store's fetch_and_delete, TxKazooClient stuff, TimerService, check_events_in_bucket
-        and twisted.internet.task.Clock is used to simulate time
+        """Mock all the dependencies of SchedulingService.
+
+        This includes logging, store's fetch_and_delete, TxKazooClient
+        stuff, TimerService, check_events_in_bucket and
+        twisted.internet.task.Clock is used to simulate time.
         """
         super(SchedulerServiceTests, self).setUp()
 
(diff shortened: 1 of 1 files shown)


otter/test/test_scheduler.py
```diff
@@ -79,31 +94,32 @@ def _start_service(self):
         self.scheduler_service.running = True
 
     def test_start_service(self):
-        """
-        startService() calls super's startService() and creates SetPartitioner object
-        """
```
manishtomar: I like this style of docstring more than the new one :(

radix: uh, yeah, what the heck is that new format from? it's certainly not pep257.

radix: sorry for that short comment, to elaborate:

> Unless the entire docstring fits on a line, place the closing quotes on a line by themselves. This way, Emacs' fill-paragraph command can be used on it. 

Additionally, having the starting line of a docstring come on the next line after the opening quotes _is_ conformant: 

>  The summary line may be on the same line as the opening quotes or on the next line. 

So I don't understand why we need to change all these docstrings.

lvh: We don't; this is pyformat's default mode of operation.

radix: hmm. is it possible to tweak it to try to reduce the churn? ideally it'd be nice if it kept whatever format was there if it was already PEP257-compliant, which `"""Single-line-docstrings"""`, as well as the

>"""
>balanced
>"""

format. The biggest problem, though, is that this format isn't actually conformant to PEP257, since it puts the closing lines on the same line as the content.

lvh: Yeah, I'll just discard the hunks. First, to find an efficient way to revert the hunks in this one...

lvh: Unfortunately, it also does nice things to the docstrings, like caps and punctuation, but I guess we'll just only do that for new code.

manishtomar:
LGTM

PR #911, opened 2015-01-15
added notes about admin user required for SSH keys in launchConfig and mo...

MariaAbrahms:
...re notes about how lunchConfig handles updates


docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -461,12 +458,39 @@ is external so it's handy for Getting Started to share -->
...
-            <xi:include href="chapters/glossary.xml" xpointer="element(Glossary-000)"/>
-            <!-- -->
+        <section xml:id="How_Auto_Scale_Works">
+            <title>How Auto Scale Works</title>
+            <para>Rackspace &PROD; is an API-based tool that automatically scales resources in response to
+                an increase or decrease in overall workload based on user-defined thresholds.</para>
+            <para>&PROD; functions by linking these services:
+                <itemizedlist>
+                    <listitem>
+                        <para>&PROD; API, described in this guide.</para>
```
cyli: Sorry, I don't really know docbook very well.  Is `&PROD;` Autoscale?  If so, shouldn't this list item be Nova, because Autoscale links Nova and load balancers (both cloud load balancers and rackconnect load balancers)?

docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -461,12 +458,39 @@ is external so it's handy for Getting Started to share -->
...
+                <note>
+                    <para>&PROD; does not configure any information within a server. You must configure your
+                        services to function when each server is started. We recommend automating your
+                        servers' startup processes with
+                        <link xlink:href="http://www.opscode.com/chef/">Chef</link> or a similar tool.</para>
+                </note>
+            </para>
+            <para>
+                <note>
+                    <para>&PROD; can utilize all Rackspace Cloud Server API operations. For more details, see the
```
cyli: I might be misreading this, but I'm not sure if "all Rackspace Cloud Server API operations" is accurate, the operative word being "operations".  For instance, we definitely do anything with key creation or deletion.

We would like to do more at some later point, but pretty much all we do right now are create and delete cloud servers, but we do support all the same parameters that are listed here specifically:  http://docs.rackspace.com/servers/api/v2/cs-devguide/content/CreateServers.html.

docbook/src/docbkx/autoscale-gettingstarted.xml
```diff
@@ -399,7 +409,14 @@ xmlns:html="http://www.w3.org/1999/xhtml" xml:id="autoscale-getting-started" xml
                     <para>&PROD; does not configure any information within a server. You must configure your
                         services to function when each server is started. We recommend automating your
                         servers' startup processes with
-                        <link xlink:href="http://www.opscode.com/chef/">Chef</link>or a similar tool.</para>
+                        <link xlink:href="http://www.opscode.com/chef/">Chef</link> or a similar tool.</para>
+                </note>
+            </para>
+            <para>
+                <note>
+                    <para>&PROD; can utilize all Rackspace Cloud Server API operations. For more details, see the
```
cyli: Same comment as above with regards to "all Rackspace Cloud Server API operations".

docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -2201,6 +2229,20 @@ config at https://github.com/rackerlabs/autoscaling-chef/blob/master/site-cookbo
                 server is, and which cloud load balancer or
                 RackConnectV3 load balancer pool the new server
                 connects to.</para>
+            <note>
+                <itemizedlist>
+                    <listitem><para>
+                        The <parameter>launchConfiguration</parameter> uses the <emphasis role="bold">admin</emphasis> user’s SSH key pair name,
```
cyli: Maybe "The launchConfiguration uses the admin user to scale up, usually the first admin user found on the tenant.  Only that particular admin user's SSH key pair names can be used in the launchConfiguration."  ?

The reason being that the key pair name only matters if the user specifies a key name in the launch configuration - if it's unspecified, Cloud Servers does not do anything with keys and hence the key pair names do not matter.

docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -57,9 +57,9 @@
         <releaseinfo>API v1.0</releaseinfo>
         <productname>&PRODNAME;</productname>
         <!-- enter the publication date of this document; in revhistory, explain what changed -->
-        <pubdate>2014-12-04</pubdate>
+        <pubdate>2015-01-14</pubdate>
         <copyright>
-            <year>2014</year>
+            <year>2015</year>
```
lvh: I'm not a lawyer, but if I recall correctly your options in the US are a) don't specify a year (which is fine) or b) specify all the years?

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: yes, &PROD; is "Auto Scale". There was a bit of a chat on slack about this last night and i had the impression that we wanted to emphasize Rax Cloud Serves API and try to leave Nova out of it. If i'm wrong, if it is more truthful/useful; to say Nova, i am happy to make that change.

cyli: Oh apologies, by "Nova" I meant our Rax Cloud Servers.  :)  Sorry, I tend to use them interchangeably.

I just meant should this bullet point be something cloud server related, rather than "Auto Scale".

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: Done

docbook/src/docbkx/autoscale-gettingstarted.xml
MariaAbrahms: Updated

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: Updated

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: Hmmm, Auto Scale's first year was 2014, so I think this is right

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: I changed the wording, but left in "operations". Let me know...

docbook/src/docbkx/autoscale-devguide.xml
cyli: I'm not sure what this looks like when it's built, but it sounds like if `&PROD;` is "Auto Scale", then it'd say something like:

Auto Scale functions by linking these services: 
- Auto Scale API, described in this guide, 
- Load-balanced Cloud Servers.  <link to load balancers>

Should the first bullet point be "Rackspace Cloud Servers" and the second one be "Cloud Load Balancer service", since that's the link?

docbook/src/docbkx/autoscale-devguide.xml
cyli: A previous comment was changed (~line 551), this one says could also use the same change though.  :)

docbook/src/docbkx/autoscale-devguide.xml
lvh: As in, it should say 2014-2015, or as in the commit is correct?

docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -461,12 +458,40 @@ is external so it's handy for Getting Started to share -->
...
+                <note>
+                    <para>&PROD; does not configure any information within a server. You must configure your
+                        services to function when each server is started. We recommend automating your
+                        servers' startup processes with
+                        <link xlink:href="http://www.opscode.com/chef/">Chef</link> or a similar tool.</para>
+                </note>
+            </para>
+            <para>
+                <note>
+                    <para>&PROD; can utilize all create and delete server Rackspace Cloud Server API operations. For more details,
```
cyli: WRT to the discussion https://github.com/rackerlabs/otter/pull/911#discussion_r23041379 about the word "operations":

I am probably just biased because I deal too much with web requests and so an "operation" to me implies GET/PUT/PATCH/POST/DELETE to a particular URL.  So "create server" would be an API operation, and "delete server" would be an api oepration.  But since they've been enumerated already, that seems to make the phrase "all API operations" feel redundant?

We certainly make use of all the possible configuration of parameters in create server, so if you are counting every possible combination of parameters a different create server operation, I guess that makes sense.  It seems like an odd term to me, but I might just have a different context.

There aren't any parameters for delete servers though, so that's just a single operation.

radix: Yeah, I would go with "API parameters" or "configurations" or something, not "operations" for this case.

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: yes, updating

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: got it

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: We had an initiative in DevDoc on this issue last year. The copyright is drawn from a template, so, yes, that's the way they want it. This change I made does not actually print out (I know because I tried to make it 2014-2015 and it was not taken).

docbook/src/docbkx/autoscale-devguide.xml
MariaAbrahms: got it :-)

MariaAbrahms:
I was not able to get the xi:include to work :-( Some major changes in DocBook production are being, or have been, put into place and that might be why. Seeing as how I plan on a re-org of the concepts anyway, and I have to upgrade my production process tomorrow (or not be able to publish), I simply made the GSG and Dev guide Concepts sections identical (except for xml IDs). To assuage my bruised soul, I also reworked some titles to be sentence-case and follow new-ish DevDoc standards. Am re-submitting the PR now and I'll include this note in github as well. The changes may look extensive, but essentially all I did is combine the GSG concepts and the Dev guide concepts into one and duplicate them; when I do the re-org I will make the xi:include work.

lvh:
@MariaAbrahms ping to see how notifications work

PR #914, opened 2015-01-15
Add a CONTRIBUTING.md

radix:
This formalizes our review process and some other bits of process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,6 @@ discussion about a change would be useful.
 
 ## Pull Requests
 
-Heres are the requirements for making a PR:
-
 Code must be unit tested. run "make lint" and "trial otter" before submitting
 your PR.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,3 +44,8 @@ their own changes.
   merge the branch immediately after making the change (and only that change).
 - If the PR has a "LGTM" but it conflicts with master, any party may merge
   it forward, give it a sanity check, and merge it.
+- If conversation on a PR comes up between multiple parties, anyone who wishes
+  to merge the PR must make a *reasonable* attempt at confirming consensus
+  before merging.
+- If any core contributor is outright opposed to a PR, it must not be merged
+  until consensus is reached.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,20 +28,19 @@ their own changes.
 
 ## Review Process
 
-- At least one review is required per Pull Request.
+- At least one review is required per Pull Request, before it can be merged.
 - If a PR does not have the "in progress" label, it is considered up for
   review. If an author realizes a PR needs more work (usually after a reviewer
-  has made a bunch of comments), they may add the "in progress" label.
+  has made some comments), they may add the "in progress" label.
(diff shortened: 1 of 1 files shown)


radix:
@manishtomar: since you haven't yet responded on the mailing list, can I get a specific review from you (as well as anyone else who wants to review this :))

radix:
Input from @sam-falvo, @kivattik, and @MariaAbrahms would also be appreciated, as this document makes (very light) reference to documentation and testing procedures.

lvh:
LGTM! Only thought is re: In Progress label vs just closing the PR if it's not open for review/merging?

manishtomar:
Sorry for not responding earlier. I actually like the review tag. It makes it very clear that a PR is up for review. Its good that the tag is not automatically assigned but will be nice that the author can add it later when they think it is ready for review.

lvh:
I would prefer to not have 1 bit (open/closed) + 1 bit (review tag, no review tag) + 1 bit (in progress tag, no in progress tag) so we can encode 1 bit of information. Otherwise, we have a bunch of inconsistent states that probably mean something very different to a random passer-by than they formally do according to the development standard.

radix:
@manishtomar well, we already use an "in progress" label, so we're not getting any additional information out of the "review" label -- I think it's clear that we only need one or the other. My argument for getting rid of "review" and keeping "in progress" is that "in progress" is the uncommon case, so it's less of a process burden to require people to explicitly add "in progress" when they don't want it to be reviewed yet.

radix:
@lvh I hadn't thought of closing PRs. I wonder how well that works in practice... 

I remember that some of the Twisted developers discussed it when considering how to translate the Twisted review process to GitHub, and the conclusion was that it was a bad idea because it can seem like an aggressively negative thing to a third-party contributor. However, that may not be a concern for us, since third-party contributors to Otter are extremely rare (I can't say "non-existent", thanks to @tomprince :)

lvh:
Yeah. It's not terribly common, even though it's supposedly the canonical Github thing to do. I guess we could leave a friendly message? One of the issues is that people without commit access can not re-open PRs. As a result, I'm also willing to accept a single tag. I agree with @radix reasoning to use in progress; also because when I open a PR to a project on Github, I don't expect to have to do more stuff for that to mean "please look at my PR".

Then, the encoding works like:

- open, not in progress: ready for review
- open, in progress: open, but don't bother looking
- closed, not in progress: retracted/closed as invalid
- closed, in progress: invalid state, probably shouldn't happen

manishtomar:
Sounds good about "in progress". LGTM

radix:
Thanks @manishtomar and @lvh.

Since it's relevant to this very PR, I've added some words about what to do when a conversation or disagreement erupts on a PR. Please check out the most recent change.

lvh:
LGTM!

manishtomar:
LGTM about disagreement.

radix:
I just added a line about reviewers being able to merge a PR. That was one of my main goals with this doc but I forgot to spell it out.

lvh:
Still LGTM :)

sam-falvo:
LGTM as well.  I'd like to steal this for my Kestrel's repo.

radix:
Alright, this has had enough eyeballs for now. If anyone else things something should change, submit a PR :)

PR #917, opened 2015-01-16
Cassandra tests lint fixes

lvh:
It's still complaining about import order, but I can't figure out why. There's a REVIEW comment in there... somewhere.

### DIFF
--- a/otter/test/models/test_cass_models.py
+++ b/otter/test/models/test_cass_models.py
@@ -10,7 +10,7 @@ import itertools
 import mock
 from jsonschema import ValidationError
 from kazoo.protocol.states import KazooState
-from silverberg.client import ConsistencyLevel, CQLClient
+from silverberg.client import CQLClient, ConsistencyLevel
 from testtools.matchers import IsInstance
 from twisted.internet import defer
 from twisted.internet.task import Clock
--- a/otter/test/models/test_cass_models.py
+++ b/otter/test/models/test_cass_models.py
@@ -2,7 +2,6 @@
 Tests for :mod:`otter.models.mock`
 """
 from collections import namedtuple
-from functools import partial
 import json
 import mock
 from datetime import datetime
--- a/otter/test/models/test_cass_models.py
+++ b/otter/test/models/test_cass_models.py
@@ -74,7 +74,8 @@ def _de_identify(json_obj):
 def _cassandrify_data(list_of_dicts):
     """
     To make mocked up test data less verbose, produce what cassandra would
-    return from a list of dictionaries.  So for instance, passing the following:
+    return from a list of dictionaries.  So for instance, passing the
+    following:
 
         [{'policyId': 'group1', 'data': '{}'},
          {'policyId': 'group2', 'data': '{}'}]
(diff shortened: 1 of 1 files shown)


lvh:
Oops, looks like I actually screwed up a test...

lvh:
Okay, should be fixed now.

Remaining lint (which I don't understand...):

```
~/P/r/otter: make flake8diff                                                                                                                                                                        09:09:05
git diff --patch --no-prefix 157afd27badbd44de96bae0453012b881db5eddd | flake8 --diff
./otter/test/models/test_cass_models.py:11:1: I201 Missing newline before sections or imports.
./otter/test/models/test_cass_models.py:12:1: I201 Missing newline before sections or imports.
./otter/test/models/test_cass_models.py:13:1: I201 Missing newline before sections or imports.
./otter/test/models/test_cass_models.py:14:1: I201 Missing newline before sections or imports.
./otter/test/models/test_cass_models.py:15:1: I201 Missing newline before sections or imports.
make: *** [flake8diff] Error 1
```

lvh:
Thanks for the help with the import lint, @radix!

radix:
I'm reviewing this.

radix:
+                                 # REVIEW: is ^ this actually realistic?
+                                 # Shouldn't this be an iso8601-ish timestamp?

I assume the content in there doesn't actually matter for the test - like, active and pending values also don't look like they would for real.  I don't think it's a problem. The test is just proving that it returns the same data it fetches.

You changed a bunch of cases of the integer `6` to ConsistencyLevel.TWO, but ConsistencyLevel.TWO is 7 for me. QUORUM is 6. Though I see that you changed it in both the test data and the assertion, so I guess it doesn't matter.

Why did you change all the GroupState calls to pass arguments by keyword, even though they take them positionally?

This LGTM, but for future lint-fixes I would recommend trying to minimize the changes you make to expedite review :) I won't merge in case you want to clean up some things like the "REVIEW" comment

lvh:
> I assume the content in there doesn't actually matter for the test - like, active and pending values also don't look like they would for real.  I don't think it's a problem. The test is just proving that it returns the same data it fetches.

Possibly. Sometimes it is problematic. Consider what happened with IPs (`"ipv4"` vs an actual IP) and dates (upgrading iso8601). Both of those caused test breakage because a test used some bogus test data.

> You changed a bunch of cases of the integer `6` to ConsistencyLevel.TWO, but ConsistencyLevel.TWO is 7 for me. QUORUM is 6. Though I see that you changed it in both the test data and the assertion, so I guess it doesn't matter.

I didn't check which is which; I just wanted to use the real thing. I'm not supposed to know/care that those things are actually integers...

> Why did you change all the GroupState calls to pass arguments by keyword, even though they take them positionally?

I don't think anyone can reasonably tell the umpteen positional arguments apart when they're a bunch of (usually empty) literals.

> This LGTM, but for future lint-fixes I would recommend trying to minimize the changes you make to expedite review :) I won't merge in case you want to clean up some things like the "REVIEW" comment

Okay.

lvh:
Thanks for reviewing this, @radix!

PR #918, opened 2015-01-16
Metrics lint

manishtomar:
flake8 fixes for `metrics.py` and `test_metrics.py`

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -205,14 +205,14 @@ def add_to_cloud_metrics(request_func, conf, region, total_desired,
     to Cloud metrics
 
     :param dict conf: Metrics configuration, will contain tenant ID of tenant
-                      used to ingest metrics and other conf like ttl
+        used to ingest metrics and other conf like ttl
     :param str region: which region's metric is collected
     :param int total_desired: Total number of servers currently desired
-                              in the region
+        in the region
     :param int total_actual: Total number of servers currently
-                             there in the region
+        there in the region
     :param int total_pending: Total number of servers currently
-                              building in a region
+        building in a region
 
     :return: `Deferred` with None
     """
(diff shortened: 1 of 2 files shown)


otter/metrics.py
```diff
@@ -180,35 +191,40 @@ def get_all_metrics(cass_groups, authenticator, service_mapping, region,
...
-    :param str identity_url: URL of identity API to authenticate users given in `conf`
+    :param dict conf: Metrics configuration, will contain tenant ID of tenant
+                      used to ingest metrics and other conf like ttl
     :param str region: which region's metric is collected
-    :param int total_desired: Total number of servers currently desired in a region
-    :param int total_actual: Total number of servers currently there in a region
-    :param int total_pending: Total number of servers currently building in a region
-    :param _treq: Optional treq implementation for testing
+    :param int total_desired: Total number of servers currently desired
+                              in the region
```
lvh: Sorry, I may have been unclear about this: the indentation is supposed to be 4, not up until the text.

lvh:
LGTM modulo docstring comments. Sorry about that :( Thanks for working on this!

lvh:
As pointed out by @radix, we temporarily became more lax about unused imports, because we went from full to diff. Therefore, you might also want to run pyflakes otter, although I doubt it matters for this PR, since it's just about lint fixes.

otter/metrics.py
manishtomar: Why doesn't flake8 capture this? I ran it on the whole file.

lvh: flake8 doesn't by default have a docstring linter. pep257 takes care of part of it, but not all of it. sphinx has some plugins, but they're not installed yet.

After those few PRs, we're definitely not done with a fully-functional linter.

manishtomar:
FYI: pyflakes is good with these files.

PR #919, opened 2015-01-16
fix the Retry performer to actually conform to the performer interface

radix:
There was a bug with perform_retry -- it didn't take the box argument and deal with it properly. Fix by using @sync_performer.

This wasn't caught in integration tests because none of our integration tests exercise Retries yet.

### DIFF
--- a/otter/test/test_retry.py
+++ b/otter/test/test_retry.py
@@ -3,7 +3,9 @@ Tests for :mod:`otter.utils.retry`
 """
 import sys
 
-from effect import Constant, Delay, Effect, Func
+from effect import (
+    ComposedDispatcher, Constant, Delay, Effect, Func, TypeDispatcher,
+    base_dispatcher, sync_perform)
 from effect.testing import Stub, resolve_effect
 
 import mock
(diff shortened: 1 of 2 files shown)


otter/util/retry.py
```diff
@@ -342,6 +342,7 @@ class Retry(object):
     """
 
 
+@sync_performer
 def perform_retry(dispatcher, intent):
```
manishtomar: Aha. So I did find something fishy with this function in last PR :wink:

otter/test/test_retry.py
```diff
@@ -456,13 +458,19 @@ class EffectfulRetryTests(SynchronousTestCase):
     # an object which calls a function with arguments specified as public
     # attributes, I guess...
 
+    def setUp(self):
+        self.dispatcher = ComposedDispatcher([
```
manishtomar: +1. Yes please make custom dispatchers instead of using global one :) for the tests. Also, I wonder if there _should_ be a global dispatcher. But that is not in the scope of this PR.

otter/test/test_retry.py
```diff
@@ -456,13 +458,19 @@ class EffectfulRetryTests(SynchronousTestCase):
     # an object which calls a function with arguments specified as public
     # attributes, I guess...
 
+    def setUp(self):
```
manishtomar: Docstring?

otter/test/test_retry.py
radix: I'm not sure what you mean about a global dispatcher. are you talking about the functions in otter.effect_dispatcher? Did we have a global dispatcher somewhere?

My intention is for the ideal to be having one main dispatcher in an actual running program, but not for it to be a module-global variable -- more like a local variable in something like main() or a REST handler.

manishtomar:
LGTM :)

otter/util/retry.py
radix: I sadly can't remember it. Was it in the PR by tom?

manishtomar: https://github.com/rackerlabs/otter/pull/852#discussion_r22957988

radix:
Thanks Manish!

PR #920, opened 2015-01-16
Bring back full pyflakes coverage in "make lint"

radix:
The linting changes recently switched everything to diff-based linting, but we were already at 100% pyflakes coverage. This PR brings back full coverage for pyflakes and makes it a mandatory part of "make lint".


lvh:
LGTM!

PR #921, opened 2015-01-16
Use ServiceRequest in otter.convergence.gathering and (some of) metrics

radix:
This changes the get_all_server_details function to return Effects of ServerRequest instead of using a parameterized request_func.

The users of this code are in both the convergence process and in metrics.py. I've updated all callers to use it. metrics.py has the first use of TenantScope to scope the ServiceRequests to a tenant ID.

@manishtomar, it'd be nice if you could either show me how to test the metrics code myself in my test VM or let me know if this branch works with metrics.


radix:
This is ready for review!

otter/http.py
```diff
@@ -257,6 +257,11 @@ def perform_tenant_scope(
     to be partially applied, and the result is a performer that can be put into
     a dispatcher.
     """
+    if _concretize is None:
+        # The reason this isn't set as the default kwarg value above is because
+        # of a mock in test_metrics.py.
```
manishtomar: I am sorry about this :(. This would've cause pain while debugging.

radix: oh no, it wasn't a pain for me -- I added the mock myself. I'm sure it'll be a pain for the next person that tries to change some mock though ;)

lvh: Is there a plan right now to get rid of it, or are we waiting until it becomes a problem?

lvh:
The code LGTM; not merging because I don't know where we are re: testing it :)

otter/http.py
radix: So, yeah, the tests that are doing it are the `GnarlyGetMetricsTests`. They are already almost entirely redundant with the tests for `get_all_metrics_effects`, except that they ensure that the thin layer of non-Effect-based code wrapped around it (in `get_all_metrics`) works. That layer is shrinking with this branch, fortunately, but the main reason that function needs to be non-effectful now is because of the whole _perform_limited thing, to maintain limited parallelism while performing the child effects.

There are two solutions to this, one quick, and one better:

- implement our own LimitedParallel intent, which just does what _perform_limited does now in its performer
- add the parallelism limit feature to Effect's `parallel`.

Once we do one of those things, get_all_metrics will be pure and Effectful, and we can get rid of the Gnarly tests. (it also means we'll be able to start refactoring `collect_metrics` to be Effectful, too).

manishtomar:
Tested it. Code works fine :). However, I am having a look at it merely for my own understanding. Feel free to merge if it is blocking other PRs since it already has a +1.

manishtomar:
`get_clb_contents` still uses `request_func`. Will be nice to change that. Can be done in this or another PR. Non-blocking based on your preference :)

radix:
There's a lot of stuff that uses request_func. This PR is very intentionally just one function :)

otter/http.py
radix: The one redeeming quality of this change is that I got rid of _two_ mocks and replaced them with one (different) one.

lvh:
HURRRGHH MERGE FORWARD VAGARIES

radix:
nooo! haha :) I'll do it.

Issue #883, opened 2015-01-03
converge should prepare launch config


manishtomar:
Each `CreateServer` step returned from `converge` should have launch config that corresponds to  [`otter.worker.launch_server_v1.prepare_launch_config`](https://github.com/rackerlabs/otter/blob/master/otter/worker/launch_server_v1.py#L555)

lvh:
Looks like we can probably just copy that function over to a place in convergence?

manishtomar:
Probably. However to ensure purity of `converge` I was thinking having uuid implementation be part of `DesiredGroupState` which will have real world implementation when running as a whole and dummy implementation during tests. Will think about it some more.

manishtomar:
I am working on this.

radix:
Name generation, which is the hard part of this (since it's side-effectful) has been resolved with #930. The only other important piece of metadata that I'm aware of is rax:auto_scaling_group_id, which we use during gathering to discover the correct servers for a group. I'll do a PR to add that.

Issue #854, opened 2014-12-23
Replace convergence.steps.Request with ServiceRequest


lvh:
Once the `ServiceRequest` intent exists, it *really* should replace `convergence.steps.Request`, because all that's doing is holding the data, and that's the job of `ServiceRequest`.

The good news is that `reqs_to_effect` no longer needs to exist, because it's *already* an effect, so it's really just `parallel([s.as_request() for s in optimized(steps)])` or something. Maybe `as_request` should be renamed.

lvh:
Closed by #936.

PR #922, opened 2015-01-17
Convergence prepares launch config

manishtomar:
Implements #883. `DesiredGroupState` has iterable of launch configs instead of one fixed launch config. This iterable is generated on the fly using `prepare_launch_config`.

### DIFF
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -2,10 +2,10 @@
 Code for composing all of the convergence functionality together.
 """
 
+import itertools
+import time
 from collections import defaultdict
 from functools import partial
-import time
-import itertools
 
 from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -5,11 +5,13 @@ Code for composing all of the convergence functionality together.
 from collections import defaultdict
 from functools import partial
 import time
+import itertools
 
 from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
 from otter.convergence.model import DesiredGroupState, CLBDescription
 from otter.convergence.planning import plan
+from otter.worker.launch_server_v1 import prepare_launch_config
 
 
 def execute_convergence(request_func, group_id, desired_group_state,
@@ -66,17 +68,23 @@ def json_to_LBConfigs(lbs_json):
     return lbd
 
 
-def get_desired_group_state(launch_config, desired):
+def get_desired_group_state(group_id, launch_config, desired,
+                            _prepare_lc=prepare_launch_config):
     """
     Create a :obj:`DesiredGroupState` from a launch config and desired
     number of servers.
 
+    :param str group_id: Scaling group ID
     :param dict launch_config: Group's launch config as per
         :obj:`otter.json_schema.group_schemas.launch_config`
     :param int desired: Group's desired capacity
+    :param callable _prepare_lc: Optional implementation of
+        `prepare_launch_config` useful for testing
     """
     lbs = json_to_LBConfigs(launch_config['args']['loadBalancers'])
+    server_lc = {'server': launch_config['args']['server']}
     desired_state = DesiredGroupState(
-        launch_config={'server': launch_config['args']['server']},
+        launch_configs=(_prepare_lc(group_id, server_lc)
+                        for _ in itertools.count()),
         desired=desired, desired_lbs=lbs)
     return desired_state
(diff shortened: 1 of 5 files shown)


lvh:
I read both tickets, but I don't understand why `DesiredGroupState` now has an iterable of launch configs.

otter/convergence/planning.py
```diff
@@ -190,16 +190,17 @@ def converge(desired_state, servers_with_cheese, load_balancer_contents, now,
...
 
     # create servers
-    create_steps = [create_server] * (desired_state.desired
-                                      - (len(servers_in_active)
-                                         + len(waiting_for_build)))
+    servers_to_create = (
+        desired_state.desired - (len(servers_in_active) +
+                                 len(waiting_for_build)))
+    create_steps = [
+        CreateServer(launch_config=next(desired_state.launch_configs))
```
radix: Using `next` here means that the converge function now has side-effects.

otter/convergence/composition.py
```diff
@@ -66,17 +68,23 @@ def json_to_LBConfigs(lbs_json):
...
     :param int desired: Group's desired capacity
+    :param callable _prepare_lc: Optional implementation of
+        `prepare_launch_config` useful for testing
     """
     lbs = json_to_LBConfigs(launch_config['args']['loadBalancers'])
+    server_lc = {'server': launch_config['args']['server']}
     desired_state = DesiredGroupState(
-        launch_config={'server': launch_config['args']['server']},
+        launch_configs=(_prepare_lc(group_id, server_lc)
+                        for _ in itertools.count()),
```
radix: This is really surprising. Why are we defining launch_configs as an infinite iterable of launch configs? I take it prepare_launch_config has side-effects (some random number stuff?) but this solution strikes me as very unexpected.

There's really no difference between this and just adding a function called "get_launch_config" that returns a new, randomized launch config. Using an iterable just means that code that wants to access this has to use `next` to get the return value, which is a pretty strange interface. I think it's more reasonable to call prepare_launch_config directly from that convergence code than to hide this side-effect in what used to be a simple data structure.

lvh: My thoughts exactly (see PR comment), but let's have the discussion here by the code :)

manishtomar:
I started with having `prepare_launch_config` callable in `DesiredGroupState` but then moved to this for reasons I don't really know.

radix:
If you really want to avoid causing side-effects in `converge` (which I think is an admirable goal), then the proper way to do that would be to encode in as an Effect. Right now we have a very restricted path from convergence to Effects (converge must return `Step`s, which can only be converted to `step.Request`s, with no mechanism for other Effects to be inserted anywhere). We may want to widen that interface so that `CreateServer` can return an *Effect* which 1) generates random data to put into the metadata and 2) creates a ServiceRequest.

radix:
... and to do that, we'd need to expand the IStep interface to have as_effect instead of as_request.

manishtomar:
I didn't close it. Just reopened it.

manishtomar:
If I was doing this in Haskell,  I would have `RandomGen` (a seeded random generator) instance be part of `DesiredGroupState`. This will be passed to `prepare_launch_config` (which can also be part of `DesiredGroupState`) which uses this generator to generate launch config. This ensures purity. 

In our case the lower level function that does random stuff is `generate_server_name` that needs 4-byte random string. How would you feel if `DesiredGroupState` has a `gen_random_string` function that gets passed to `prepare_launch_config` and eventually to `generate_server_name`?

radix:
So, I think we should refactor this a bit, and then it'll be more clear. I don't think `DesiredGroupState` is the right place for this behavior. I think it should be closer to where we're converting an abstract representation of a server we want (`CreateServer`) into the concrete thing. Here's what I suggest:

1. we refactor prepare_launch_config to not have any side effects, and to take the server name as an argument.
2. Steps change to have as_effect instead of as_request.
3. CreateServer's as_effect looks like this:

```python
@do
def as_effect(self):
    name = yield Effect(FuncIntent(generate_random_name))
    lc = yield prepare_launch_config(name, self.foo, self.bar)
    return service_request(...)
```

Does this make sense?

manishtomar:
Sounds good.

radix:
@manishtomar, if you would like to work on that proposed change I'd be happy to help you in any way :)

radix:
#854 is relevant to this

manishtomar:
@radix I am closing this as it is of no use. Would you mind creating tickets for the effect changes? I am guessing: 1) Replace all `as_request` to `as_effect` and change/remove `steps_to_effect`. 2) Implement `CreateServer.as_effect` as discussed above.

radix:
sure thing!

radix:
#929 and #930 filed.

PR #923, opened 2015-01-19
allow per-service region configuration.

radix:
This should resolve issue #896.

This repurposes the `service_mapping` into a more general `service_configs`, where each value is a dict of `name` and `region`, instead of just being the name. This means that each service can be configured with its own region, though in practice it's only the metrics service that uses anything other than the default `config['region']`. The logic about how to calculate the regions based on the config file remains in `get_service_configs` (renamed from `get_service_mapping`).

All in all, a pretty simple change to solve this problem. 

Notes for reviewers: 

- the main change here is in `otter/http.py`'s `concretize_service_request`, which pulls the region out of `service_config[service_type]['region']`
- `metrics.py`, in `collect_metrics`, now creates a more general dispatcher, with no *specific* region being passed to it, just the `service_configs` mapping. (note that we could now refactor the two different creations of dispatchers in metrics.py, but I'll wait until after #921 is done to avoid too many conflicts.)
- there was a lot of renaming of `service_mapping` to `service_configs`, but I didn't do it in code which should be deleted Real Soon Now.

### DIFF
--- a/otter/constants.py
+++ b/otter/constants.py
@@ -11,17 +11,32 @@ class ServiceType(Names):
     CLOUD_METRICS_INGEST = NamedConstant()
 
 
-def get_service_mapping(config):
+def get_service_configs(config):
     """
-    Return a mapping of :class:`ServiceType` members to the actual configured
-    service names to look up in tenant catalogs.
+    Return service configurations for all services based on the config data.
+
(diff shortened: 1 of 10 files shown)


lvh:
<- reviewing this

lvh:
LGTM!

PR #926, opened 2015-01-20
Unify the dispatchers in metrics.py

radix:
There used to be two dispatchers created in metrics.py: one for making requests to cloud servers (and most other services, if necessary), and one for making requests to the metrics service, because the ServiceRequest handler didn't support services in different regions. Now that that's fixed, we can unify this dispatchers.

Along with that, I've done a little bit of tidying up of metrics.py.

- this pretty drastically simplifies `get_all_metrics`, by removing a bunch of the parameters that were only being passed so that it could create a dispatcher.
- un-gnarly-ified `get_all_metrics` tests
- replaced some mocks with some explicit parameterization of functions

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -231,8 +231,7 @@ def connect_cass_servers(reactor, config):
 
 @defer.inlineCallbacks
 def collect_metrics(reactor, config, client=None, authenticator=None,
-                    get_full_dispatcher=get_full_dispatcher,
-                    _print=False):
+                    _print=False, get_full_dispatcher=get_full_dispatcher):
     """
     Start collecting the metrics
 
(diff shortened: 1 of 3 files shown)


otter/metrics.py
```diff
@@ -301,7 +296,7 @@ def collect_metrics(reactor, config, client=None, authenticator=None,
                            reverse=True)
         print('groups sorted as per divergence', *group_metrics, sep='\n')
 
-    # Diconnect only if we created the client
+    # Disconnect only if we created the client
```
manishtomar: +1. Thanks for the fix :)

manishtomar:
LGTM :) Let me test the metrics once.

manishtomar:
@radix I've small modification to make it work with script. Let me know if thats ok with you and we can merge it.

radix:
oh, sorry about that! change is good. I'll merge.

PR #927, opened 2015-01-20
make get_clb_contents use service_request

radix:
Not much to say, this is along the lines of the other request_func -> service_request changes.

### DIFF
--- a/otter/test/convergence/test_gathering.py
+++ b/otter/test/convergence/test_gathering.py
@@ -230,7 +230,7 @@ class GetLBContentsTests(SynchronousTestCase):
             ShouldDelayAndRetry(can_retry=retry_times(5),
                                 next_interval=exponential_backoff_interval(2)))
         req = eff.intent.effect.intent
-        response = self.reqs[req.method, req.url, req.json_response]
+        response = self.reqs[(req.method, req.url, req.json_response)]
         return resolve_effect(eff, response)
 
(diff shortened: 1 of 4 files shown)


otter/test/convergence/test_gathering.py
```diff
@@ -220,28 +220,23 @@ def setUp(self):
...
+        Resolve a :obj:`ServiceRequest` based on ``self.reqs`` and assert
+        that it's wrapped in a Retry with the expected policy.
         """
         self.assertEqual(
             eff.intent.should_retry,
             ShouldDelayAndRetry(can_retry=retry_times(5),
                                 next_interval=exponential_backoff_interval(2)))
-        return resolve_retry_stubs(eff)
+        req = eff.intent.effect.intent
+        response = self.reqs[req.method, req.url, req.json_response]
```
manishtomar: Hmm. Didn't know you can key on a tuple like this. Do you think it would be better to have it like `self.reqs[(req.method, req.url, req.json_response)]`? Non-blocking nitpick of course :)

manishtomar:
LGTM

otter/test/convergence/test_gathering.py
radix: I tend to like it, but I probably shouldn't because inconsistency is the root of all evil. I'll change it to using an explicit tuple.

PR #928, opened 2015-01-20
make add_to_cloud_metrics use service_request

radix:
This removes the last vestiges of request_func from metrics.py

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -188,7 +188,7 @@ def get_all_metrics(dispatcher, cass_groups, _print=False,
     return d.addCallback(lambda x: reduce(operator.add, x, []))
 
 
-def add_to_cloud_metrics(conf, region, total_desired,
+def add_to_cloud_metrics(ttl, region, total_desired,
                          total_actual, total_pending, log=None):
     """
     Add total number of desired, actual and pending servers of a region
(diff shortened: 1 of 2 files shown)


otter/metrics.py
```diff
@@ -277,15 +279,8 @@ def collect_metrics(reactor, config, client=None, authenticator=None,
...
 
     # Add to cloud metrics
-
-    # WARNING: This request func is configured to make requests against the
-    # metrics region. The same request_func can't be used for other purposes,
-    # like making requests to cloud servers.
-    req_func = get_request_func(authenticator, config['metrics']['tenant_id'],
-                                metrics_log, service_configs,
-                                config['metrics']['region'])
     eff = add_to_cloud_metrics(
```
manishtomar: This effect needs to be scoped with `config['metrics']['tenant_id']`

radix: Thanks, that's a very important catch! Fixed and pushed.

The tests for it aren't very nice; collect_metrics is really pretty nasty and is definitely deserving a refactoring to be more Effectful, but the fact that get_all_metrics is still returning a Deferred and not an Effect (because of the whole parallel limit thing) is a blocker. I may try to factor this stuff better over the next couple weeks.

manishtomar:
LGTM!

PR #934, opened 2015-01-21
Fix a bunch of lint in convergence's planning bits

lvh:


radix:
ah, good old https://code.launchpad.net/~radix/+junk/astdiff .

LGTM. :-)

PR #936, opened 2015-01-21
use service_request in convergence.effecting

radix:
This has a couple of fallouts:

- now no convergence code has request_func in it any more, most notably up through execute_convergence
- the RCv3 worker code had to be updated to apply a TenantScope, and use the full dispatcher.
- I decided to put the dispatcher in the "request bag" (aka the "request_func" object that supervisor passes around), since the supervisor has access to all the data necessary for creating a dispatcher when it's creating the request_func.
- I also had to put the tenant_id in the "request bag" so that the RCv3 worker code could wrap the effect in a TenantScope.

### DIFF
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -12,7 +12,7 @@ from otter.convergence.model import DesiredGroupState, CLBDescription
 from otter.convergence.planning import plan
 
 
-def execute_convergence(request_func, group_id, desired_group_state,
+def execute_convergence(group_id, desired_group_state,
                         get_all_convergence_data=get_all_convergence_data):
     """
     Execute convergence. This function will do following:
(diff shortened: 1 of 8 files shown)


lvh:
I'm reviewing this. (Can we use assignment to designate the reviewer?)

lvh:
The ticket I was talking about is #854, but it appears this PR doesn't ref it yet.

lvh:
It appears `_PureRequestStub` isn't even used anymore within this PR.

```
~/P/r/otter: ag _PureRequestStub                                                                                                                                                                    12:22:13
otter/test/convergence/test_effecting.py
26:class _PureRequestStub(object):
37:    Tests for :class:`_PureRequestStub`, the request func test double.
46:        characteristic_attrs = _PureRequestStub.characteristic_attributes
```

Removing it here or in a separate PR is up to your discretion :)

lvh:
LGTM; not merging so you get to address that comment however you like :-)

radix:
Yeah, I meant to check if `_PureRequestStub` was still used in this PR, but I forgot. I've removed it in the next branch that removes request_func entirely.

lvh:
Derp, I didn't even read the title of #937. My bad. So, LGTM!

PR #941, opened 2015-01-21
PR reviewer is assignee

lvh:
I suggested/did this on #936. @cyli liked it, so I'm putting it up as a PR in case someone else does too.


radix:
LGTM. merging without a passing build because I doubt a CONTRIBUTING.md change will fail the build :)

Issue #805, opened 2014-12-10
convergence.Request should specify retry logic


lvh:
Right now, convergence.Request has no way to specify retry logic. This might be fine: you can just go ahead and retry convergence if anything fails. However, this means making a ton of extra requests to upstream services, and we try to be nice about not hammering their API :) Plus, it makes sense to try and execute a plan to the fullest: we've just computed that this is what we need to do, the least we can do is retry to fix transient failures in executing it :)

Retry logic should decouple "can I retry" from the actual retrying policy (e.g., 5 times starting at two seconds with exponential backoff). This is so we don't run into the issue of every retrying call site specifying its own little bespoke artisanal handrolled retrying policy (see #791).

Steps to resolve:

- [ ] Solve #791 as specified on that ticket (add a default to [`retry_effect`](https://github.com/rackerlabs/otter/blob/master/otter/util/retry.py#L351)).
- [ ] Add a `can_retry` function (`Failure -> bool`, just like the one `retry_effect` takes) attribute to `convergence.Request`. By default, this is `None`.
- [ ] Add support for this to `convergence._conv_reqs_to_effect`: create the usual create-a-request effect, inspect the `can_retry` attr. If it's `None`, just use the effect verbatim; if not, wrap it with [`retry_effect`](https://github.com/rackerlabs/otter/blob/master/otter/util/retry.py#L351)

radix:
convergence.Request is gone now, and steps can specify their own retry logic in as_effect, so I'm closing this as invalid. Some discussion here: https://github.com/rackerlabs/otter/pull/943#issuecomment-71080192

Issue #791, opened 2014-11-21
Default pure_http retry policy


lvh:
Now that pure_http has awesome retries (thanks @radix), we're starting to see a large number of call sites specify their own retry logic. It's obviously fine that call sites _can_ specify their own retry logic (or even specify that they don't want any retry logic at all), but most sites want exactly the same retry logic. It does not make sense to have every call site specify the same  `retry_effect(request_func(...), retry_times(5), exponential_backoff_interval(2))`.

One way we could do this is simply give [`retry_effect`](https://github.com/rackerlabs/otter/blob/master/otter/util/retry.py#L351) a default `can_retry`, `next_interval` value. Alternatively, a partially applied version of that function with defaults. I'd normally gravitate towards partially applied, but I think that:
- overriding the defaults is sufficiently unusual,
- providing defaults does not prevent any code from overriding them,
- by calling `retry_effect` the caller is already opting in to _some_ retry behavior,

... that default values are fine.

This is also nice because we can change retry logic behavior more easily in the future, e.g. exponential backoff with jitter.

Original ask: https://github.com/rackerlabs/otter/pull/789/files#r20639086

PR #943, opened 2015-01-22
Change IStep.as_request to IStep.as_effect, get rid of steps.Request

### LINKED ISSUE #929: Replace IStep.as_request with IStep.as_effect, to allow more flexibility in what steps can do
This will allow steps to have effects other than *just* producing a fixed, static HTTP Request. Most pressing, we need CreateServer to be able to generate a random name to put into the launch config.

radix:
Fixes #929.

Pretty straightforward change -- all the as_request methods have been renamed to as_effect, and they call service_request instead of instantiating a Request.


manishtomar:
So, no retry in returned effects. Where do you think retries should be there?

radix:
Well, there was no retry in the previous code, so this just maintains the same behavior. Adding retries should be in a different PR.

manishtomar:
Sounds good :). Just wanted to clarify when and where they will be there. IMHO, the returned effects should have retry in them as they are aware of the kind request they are making and will be most equipped with kind of retry that needs to be done.

radix:
@manishtomar oh, okay, I totally misunderstood. yeah, makes sense to do it there :)

lvh:
retest this please

lvh:
This LGTM if Jenkins agrees. What's the ticket for adding the retries?

lvh:
FYI, there's two ostensibly relevant tickets:

- https://github.com/rackerlabs/otter/issues/791
- https://github.com/rackerlabs/otter/issues/805

I think #791 can stay as-is. It still makes sense for a default retry policy to exist. It's okay if that's explicit instead of an implicit kwarg, that's up to whoever owns the ticket.

#805, that's a different beast. Maybe it should be closed as invalid: you can just wrap the effects returned from as_effect with a retrying effect, right?

manishtomar:
> Maybe it should be closed as invalid: you can just wrap the effects returned from as_effect with a retrying effect, right?

The decision on whether the effect should be retries is dependent on the Step. For example, `CreateServer` step will retry with a get server ID to ensure server is not created whereas `DeleteServer` can just retry until it gets 204. Similarly, CLB retries should account for 422 codes and such.

radix:
Yeah, agreed. I think #805 should be marked invalid; I'll do that.

lvh:
@radix So, is this in progress? Only because of the dependent PR, or something else?

lvh:
(I'm asking because my RCv3 change can't start without this)

radix:
Only because of the dependent branch. I've removed the in progress label now.

PR #945, opened 2015-01-22
make CreateServer generate a random server name

### LINKED ISSUE #930: Implement CreateServer.as_effect such that it uses an Effect to generate a random name
This depends on #929.

To keep the convergence process pure (as in, referentially transparent), we need to encode the generation of a random name as an Effect. Maybe it'd be cool to have a RandomString intent, but IMO we should be fine with  Effect(Func(generate_random_name)) for now (as long as it's not a lambda, it should be trivial to test with equality in the unit tests).

radix:
EDIT: *really* in progress -- need to make sure LB metadata gets the server name too.

Depends on #943, fixes #930.

CreateServer.as_effect generates a random name to put into the launch config.

I have some more stuff coming to do the rest of the "preparation" of launch configs (i.e. adding metadata), so don't worry :)

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -2,10 +2,13 @@
 
 from characteristic import attributes
 
+from effect import Effect, Func
+
 from zope.interface import Interface, implementer
 
 from otter.constants import ServiceType
 from otter.http import has_code, service_request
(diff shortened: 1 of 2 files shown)


radix:
I might be completely redoing this, so closing this PR.

Issue #820, opened 2014-12-11
Handle nodes being deleted out-of-band for RCv3


lvh:
Suppose node `A` is attached to load balancer `B`. Node `A` is deleted out-of-band. Otter scales down, and wants to get rid of `A` (i.e. remove it from its load balancer and delete the node). The load balancer removal step errors out. In the CLB code path, this is still considered a success. 

For the Rackers, this is [AS-425](https://jira.rax.io/browse/AS-425).

lvh:
So, this ended up being trickier than I had hoped.

RCv3 bulk operations are atomic. So, when we get a failure because some of the machines aren't attached to LBs, we should try again with those machines removed.

lvh:
I'm also going to figure out what happens when you try to remove things from an RCv3 load balancer pool that doesn't exist in the first place.

PR #949, opened 2015-01-22
Decode JSON before passing it to the success_pred

lvh:
Fortunately, nothing relied on this behavior so far, so no test changes were required. However, #820 is going to change that real soon now.


radix:
(LGTM)

PR #954, opened 2015-01-23
Support RCv3 out-of-band delete

### LINKED ISSUE #820: Handle nodes being deleted out-of-band for RCv3

lvh:
Closes #820.

This is currently blocked on RCv3 getting back to us re: atomicity semantics of bulk operations.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -130,7 +130,7 @@ class ChangeCLBNode(object):
                   'weight': self.weight})
 
 
-def _rackconnect_bulk_request(lb_node_pairs, method, success_codes):
+def _rackconnect_bulk_request(lb_node_pairs, method, success_pred):
     """
     Creates a bulk request for RackConnect v3.0 load balancers.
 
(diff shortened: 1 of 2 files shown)


lvh:
So, according to RCv3, requests are validated *before* attempting to execute them, and bulk requests are atomic. That's unfortunate, because it might mean that the only way to recover from this mess is "whatever, converge again".

cyli:
@lvh Or retry one particular add with all the problematic servers removed.  It's probably fine to wait for the next converge cycle.

lvh:
Right, but currently we have no way of specifying "retry this *other* effect instead", and the effect doesn't know if this is attempt number 1 or number 37, so...

radix:
Yes you do: add an error callback that calls service_request and returns the result. You can even maintain a counter as a parameter you pass from callback to callback.

lvh:
Sorry, I keep forgetting that parametrization bit :)

lvh:
It appears the real resolution of this issue is going to be "perform some other effects". I'll update the parent ticket and close this PR for now.

PR #957, opened 2015-01-23
Generate a random server name when creating a server in Convergence

radix:
This branch introduces the same name-generation logic that the worker code has.

It does it in a pure way, by generating the random name in the Effect returned by CreateServer.as_effect.

I factored out a generate_server_name from the prepare_launch_config function in the worker code, and made the worker code import that function from convergence so the logic is shared.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -24,10 +24,14 @@ class IStep(Interface):
         """Return an Effect which performs this step."""
 
 
-def prepare_server_name(launch_config_args, name_suffix):
+def set_server_name(launch_config_args, name_suffix):
     """
     Append the given name_suffix to the name of the server in the launch
     config.
+
+    :param launch_config_args: The launch configuration args.
+    :param name_suffix: the suffix to append to the server name. If no name was
+        specified, it will be used as the name.
     """
     name = launch_config_args['server'].get('name')
     if name is not None:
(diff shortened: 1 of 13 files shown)


otter/convergence/steps.py
```diff
@@ -19,6 +24,19 @@ def as_effect():
         """Return an Effect which performs this step."""
 
 
+def prepare_server_name(launch_config_args, name_suffix):
+    """
+    Append the given name_suffix to the name of the server in the launch
+    config.
+    """
```
manishtomar: param docstrings will be helpful :)

otter/worker/launch_server_v1.py
```diff
@@ -566,17 +568,16 @@ def prepare_launch_config(scaling_group_uuid, launch_config):
     server_config['metadata'].update(generate_server_metadata(
         scaling_group_uuid, launch_config))
 
-    if server_config.get('name'):
-        server_name = server_config.get('name')
-        server_config['name'] = '{0}-{1}'.format(server_name, generate_server_name())
-    else:
-        server_config['name'] = generate_server_name()
+    suffix = generate_server_name()
```
manishtomar: Non-blocking nitpick: `generate_server_name` name confuses a little with `prepare_server_name`. Would you be fine with renaming `generate_server_name` to `generate_random_server_name`?

manishtomar:
LGTM modulo nitpicks

otter/worker/launch_server_v1.py
radix: I'll change it to `set_server_name`. It doesn't generate a random string, since that's passed in, so `random` isn't really accurate in the name.

radix: oh, I totally misunderstood you, I'm sorry.

radix: However, I think changing prepare_server_name to set_server_name should still make this less confusing and more descriptive, so I'll do that.

manishtomar:
LGTM

PR #962, opened 2015-01-24
Change 'status' in LB message to 'status_code'

cyli:
This is breaking the elasticsearch mappings, since this is a number and I guess group status or something is logged now as a string.

We should improve our ES practices/setup, but this is less work for now and a sufficient stopgap measure.


lvh:
LGTM, assuming tests pass.

lvh:
Test to see if comments get pushed

PR #965, opened 2015-01-26
Convergence integration testing -- initial commit.

sam-falvo:
Getting the ball rolling.

This code was primarily written by @lvh while pair-prg'ing with @sam-falvo and @kivattik, so code review should be trivial.

### DIFF
--- a/otter/test/integration/test_utils.py
+++ b/otter/test/integration/test_utils.py
@@ -158,6 +158,29 @@ class MeasureProgressTests(SynchronousTestCase):
             previous_state, current_state, desired_state)
         self.assertEqual(progress, 0)
 
+    def test_servers_going_from_build_to_error_with_reaping(self):
+        """
+        When some servers go from build to error, no progress was
+        made. That works correctly even if some of the errored
+        machines get reaped in the mean while.
+        """
+        previous_state = GroupState(
+            servers=self._create_servers(3, state=ServerState.BUILD),
+            lb_connections=pset([])
+        )
+        current_state = GroupState(
+            servers=(self._create_servers(1, state=ServerState.ACTIVE)
+                     | self._create_servers(1, state=ServerState.ERROR)),
+            lb_connections=pset([])
+        )
+        desired_state = DesiredGroupState(
+            launch_config=pmap(),
+            desired=5,
+        )
+        progress = measure_progress(
+            previous_state, current_state, desired_state)
+        self.assertEqual(progress, 0)
+
     def test_reaping_errored_servers(self):
         """
         Errored servers are removed; no progress is made.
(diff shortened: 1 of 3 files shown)


otter/test/integration/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    live_states = [ServerState.BUILD, ServerState.ACTIVE]
+    return len([s for s in servers if s.state in live_states])
+
+def _count_dead_servers(servers):
+    """
+    Count servers that are in error state.
+    """
+    return len([s for s in servers if s.state is ServerState.ERROR])
+
+def _sign(n):
```
lvh: This function is no longer used.

otter/test/integration/test_utils.py
```diff
@@ -0,0 +1,201 @@
+"""
+Tests for the utility functions for convergence black-box testing.
+"""
+from pyrsistent import pmap, pset
+
+from twisted.trial.unittest import SynchronousTestCase
+
+from otter.convergence.model import (
```
radix: Given that these are explicitly supposed to be "black-box" tests, I am surprised that internal model classes are being used.

sam-falvo:
When this PR passes review, PLEASE DO NOT MERGE.  This will be a work-in-progress branch, and I'll create a sub-branch from this one for incremental reviews.

otter/test/integration/test_utils.py
sam-falvo: I don't recall what we were going to use these for.  I think for re-using any JSON marshaling in Otter?  @lvh can you confirm?

otter/test/integration/test_utils.py
```diff
@@ -0,0 +1,201 @@
...
+        current_state = GroupState(
+            servers=self._create_servers(6),
+            lb_connections=pset([])
+        )
+        desired_state = DesiredGroupState(
+            launch_config=pmap(),
+            desired=5,
+        )
+        self.assertRaises(
+            AssertionError,
```
radix: This AssertionError check could be a bit better; the function under test raises AssertionError in multiple different cases, and this isn't ensuring that it's doing it for the right reason. I'm not sure how important that is in this case, though.

radix:
@sam-falvo If it doesn't cause the build to fail, I think it'd be better to just merge it to master as soon as it passes review. Maintaining a separate ephemeral master is just going to cause headaches. It's okay if you have to make changes to this code in later PRs.

otter/test/integration/test_utils.py
lvh: We're not doing any JSON marshalling.

Basically, we just needed some objects that describe all the state otter cares about, and re-using existing model classes is easier than writing new ones. These will never be the same instances as the functions in `otter.convergence` see, if that helps :)

lvh:
@radix Agreed w.r.t regular merges of small PRs.

cyli:
This LGTM - nice to have a just a rough metric to see if convergence is moving closer to its goal.

sam-falvo:
@wirehead said it was OK to merge.

Issue #964, opened 2015-01-26
Obverser to push logs to cloud feeds


manishtomar:
Part of #838.

There will be observer to push events to CF that will be tied to twisted log system. This way the code that wants to log will just do log.msg("Launching {no_of_servers} servers", no_of_servers=4, cloud_feed=True). The observer chain will log this normally on console (that gets shipped to ES) and also send event to CF due to cloud_feed=True. This is how audit log is currently setup.

Taken from https://github.com/rackerlabs/otter/issues/884#issuecomment-70963306

manishtomar:
Closing

PR #966, opened 2015-01-27
Setup cloud feeds

manishtomar:
Implements #964. This sets up cloud feeds observer that will sanitize and add event if it has `cloud_feeds=True`. This is completed however I am not able to test it due to CF returning 500. Kindly review it.


otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,158 @@
+"""
+Ingesting events to Cloud feeds
```
radix: incredibly minor nitpick: I would say "Sending" or "Feeding" events to Cloud Feeds, not "Ingesting". Ingesting is what the receiver does.

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,158 @@
...
+    except UnsuitableMessage as me:
+        log.err(me, ('Tried to add unsuitable message in cloud feeds: '
+                     '{unsuitable_message}'),
+                unsuitable_message=me.unsuitable_message)
+    else:
+        eff = Effect(TenantScope(cf.tenant_id,
+                                 add_event(event, error, timestamp,
+                                           cf.region, log)))
+        # TODO: Log error if event was not added
+        perform(
```
radix: the perform() call should be factored out into a separate function so you can test the effectful code separately.

otter/tap/api.py
```diff
@@ -223,6 +226,15 @@ def makeService(config):
...
         admin_service.setServiceParent(s)
 
+    # setup cloud feed,
+    cf_conf = config['cloudfeeds']
+    cf = CloudFeeds(reactor=reactor, authenticator=authenticator,
+                    region=region, tenant_id=cf_conf['tenant_id'],
+                    service_configs=service_configs)
+    # TODO: It would be nice to not store this as global state but
+    # observer factory is called by twistd before this step
+    set_cloud_feeds(cf)
```
radix: We should just do something like: log.addObserver(partial(observe_cloud_feeds, cf)). Granted, the whole log observer system is global, but it shouldn't require us to maintain our own global variables.

otter/log/cloudfeeds.py
manishtomar: `add_event` was meant to be that function. This function is composition of `sanitize_event` and `add_event`. I'll try thinking some more.

radix: Yeah, everything before the perform() is still close to pure, except:

- `event.pop()`, which can be replaced with `event = dicttoolz.dissoc(event, 'cloud_feed')`
- the `log.err` call, which is not a terrible thing to leave in an otherwise pure function, though you may want to leave it outside at your preference.

So I think add_event is fine, but you could create another function that factors some stuff out of `add_event_to_cloud_feed`. Mainly the TenantScope wrapping will be easier to test if it's in a pure function.

manishtomar: @radix Does https://gist.github.com/manishtomar/433fe9910da6cf5ce8a4 look good?

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,173 @@
...
+    cf_event = {}
+    error = False
+
+    # format message
+    event_copy = deepcopy(event)
+    PEP3101FormattingWrapper(lambda e: None)(event_copy)
+    msg = event_copy["message"]
+    try:
+        event["message"] = msg[0]
+    except KeyError:
```
lvh: Looks like that's going to be IndexError, not KeyError. Missing test case?

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,173 @@
...
+    request['entry']['content']['event']['product'].update(event)
+    request['entry']['content']['event']['id'] = str(_uuid())
+    return request
+
+
+def add_event(event, tenant_id, region, log,
+              _prep_req=prepare_request):
+    """
+    Add event to cloud feeds
+    # Review: Is taking _prep_req as argument really required?
```
lvh: What's the alternative?

otter/test/test_cloudfeeds.py
```diff
@@ -0,0 +1,288 @@
...
+        "message": "human"
+    }
+
+
+class SanitizeEventTests(SynchronousTestCase):
+    """
+    Tests for :func:`otter.cloudfeeds.sanitize_events`
+    """
+
+    def setUp(self):  # noqa
```
lvh: Wait, did someone turn on the name checker?

otter/log/cloudfeeds.py
manishtomar: Good catch. And yes, its a missing test case.

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,173 @@
...
+@attributes(['reactor', 'authenticator', 'tenant_id', 'region',
+             'service_configs', 'log', 'get_disp', 'add_event'],
+            defaults={'log': otter_log, 'get_disp': get_full_dispatcher,
+                      'add_event': add_event})
+class CloudFeedsObserver(object):
+    """
+    Log observer that pushes events to cloud feeds
+    """
+    def __init__(self):
+        # This is needed for attributes to work
```
lvh: Wait, what?

otter/test/test_cloudfeeds.py
manishtomar: I ran flake8 on the files and it gave this error. I am almost sure I ran `make flake8diff` too at it gave same error.

otter/log/cloudfeeds.py
manishtomar: Since `prepare_request` was not doing IO, I was thinking of leaving it as such but it unfortunately has random generation side-effect. Also, the request prepared was really quite big to to check in test. So, I decided to pass it as arg for easier testing.

lvh: It seems like the alternative then is passing in an arg, or patching it in, right? So no, it's not really necessary, but that doesn't mean it's a problem :)

radix: I would suggest what I did for the CreateServer.as_effect method. Use `Effect(Func(uuid4))` in your chain of effects to generate the name. Then in your tests you can assert `effect.intent == Func(uuid4)` and then provide your own testing value. voila :)

otter/test/test_cloudfeeds.py
radix: Were you in your otter virtualenv at the time?

manishtomar: Yes. Although I started this branch early and didn't merge master. Let me try doing that.

otter/log/cloudfeeds.py
manishtomar: Yeah. I couldn't get this to work without the `__init__` :( It kept giving some error.

otter/log/cloudfeeds.py
manishtomar: @radix That does sound like a good idea. Somehow `Effect` does not strike me when handling non-IO based side-effect :). I'll probably do that.

otter/test/test_cloudfeeds.py
lvh: You probably have a stale plugin still installed then. Try `pip list`, or recreate your virtualenv, or figure out what the package name is. (I think it's `pep8-naming`.)

otter/log/cloudfeeds.py
lvh: I think you should try it again:

```
~/P/r/otter: git diff                                                                                                                                                                               20:08:16
diff --git i/otter/log/cloudfeeds.py w/otter/log/cloudfeeds.py
index af7b4b4..77f7368 100644
--- i/otter/log/cloudfeeds.py
+++ w/otter/log/cloudfeeds.py
@@ -146,10 +146,6 @@ class CloudFeedsObserver(object):
     """
     Log observer that pushes events to cloud feeds
     """
-    def __init__(self):
-        # This is needed for attributes to work
-        pass
-
     def __call__(self, event_dict):
         """
         Process event and push it to Cloud feeds
~/P/r/otter: trial otter.test.test_cloudfeeds                                                                                                                                                       20:08:17
otter.test.test_cloudfeeds
  CloudFeedsObserverTests
    test_event_added ...                                                   [OK]
    test_no_cloud_feed ...                                                 [OK]
    test_perform_fails ...                                                 [OK]
    test_unsuitable_msg_logs ...                                           [OK]
  EventTests
    test_add_event ...                                                     [OK]
    test_prepare_request ...                                               [OK]
    test_prepare_request_error ...                                         [OK]
  SanitizeEventTests
    test_all_cf_keys ...                                                   [OK]
    test_error ...                                                         [OK]
    test_formats_message ...                                               [OK]
    test_subset_cf_keys ...                                                [OK]
    test_unsuitable_msg ...                                                [OK]

-------------------------------------------------------------------------------
Ran 12 tests in 0.006s

PASSED (successes=12)
```

lvh:
It tripped me up that the test package (`otter.test.test_cloudfeeds`) structure doesn't match the implementation package structure (`otter.log.cloudfeeds`). Was that intentional?

manishtomar:
> Was that intentional?

Currently, there is no `test/log` directory where I would've liked to put it and I didn't want to put these tests in `test/test_log.py`. So, I left it in `test/`. Should I create `test/log/test_cloudfeeds.py` and move `test_log.py` there?

radix:
@manihstomar +1 to creating test/log/test_cloudfeeds.py

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    request = deepcopy(req_fmt)
+    if error:
+        request['entry']['content']['event']['type'] = 'ERROR'
+    request['entry']['content']['event']['region'] = region
+    request['entry']['content']['event']['eventTime'] = timestamp
+    request['entry']['content']['event']['product'].update(event)
+
+    def set_id(_id):
+        request['entry']['content']['event']['id'] = _id
```
radix: This isn't quite what I imagined. While technically prepare_request, as a white box, is pure, I think it can be factored a bit more elgeantly without this mutation here.

What I suggest is this: make prepare_request take a parameter for the event ID. In add_event, create an Effect of Func(uuid4), and then attach prepare_request as a callback to it, which will get the event ID (and all the other parameters it currently takes) and return a new request dict. This makes prepare_request a simple function that doesn't care about effects and just produces a pure data structure.

otter/log/cloudfeeds.py
```diff
@@ -0,0 +1,177 @@
...
+        # in infinite recursion
+        log_keys = keyfilter(
+            lambda k: k not in ('message', 'cloud_feed'), event_dict)
+        log = self.log.bind(
+            system='otter.cloud_feed', cf_msg=event_dict['message'][0],
+            **log_keys)
+        try:
+            eff = self.add_event(event_dict, self.tenant_id, self.region, log)
+        except UnsuitableMessage as me:
+            log.err(None, ('Tried to add unsuitable message in cloud feeds: '
```
radix: You wanna add an `otter_event_type` key here? I suggest something like "unsuitable-cloud-feeds-message".

otter/tap/api.py
```diff
@@ -227,6 +228,17 @@ def makeService(config):
         admin_service = service(str(admin_port), admin_site)
         admin_service.setServiceParent(s)
 
+    # setup cloud feed
+    cf_conf = config.get('cloudfeeds', None)
+    if cf_conf is not None:
+        service_configs[ServiceType.CLOUD_FEEDS] = {
```
radix: Why are you modifying `service_configs` here instead of just making `get_service_configs` return it?

manishtomar: Because I want cloudfeeds to be optional. Especially in dev environment. It loses optional(ness) if it has to come from `get_service_configs` as that will always assume cloudfeed's config to be there. In fact, I feel `service_config` should be created in `makeService` depending on the values in config file instead of expecting all configs in `get_service_config`.

otter/log/cloudfeeds.py
manishtomar: Yeah. That sounds good. *sigh*

otter/tap/api.py
radix: It can be optional even in get_service_config -- it could have this same conditional logic to only include CLOUD_FEEDS if it's in the config file.

radix:
LGTM now! Too bad we can't test it against real CF yet. I'll leave it up to you if you want to merge now since it hasn't been tested (because of the misconfiguration in staging). I'm fine with merging it since it won't be enabled until it's added to the config.json file in prod.

PR #967, opened 2015-01-27
Support RCv3 out-of-band deletes

lvh:
Implements #820.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -230,7 +230,7 @@ class BulkRemoveFromRCv3(object):
         """
         eff = _rackconnect_bulk_request(self.lb_node_pairs, "DELETE",
                                         success_pred=_CONSTANTLY_TRUE)
-        return eff.on(_rcv3_check_bulk_delete)
+        return eff.on(_rcv3_check_bulk_delete).on(_maybe_execute_step)
 
 
 _CONSTANTLY_TRUE = lambda _: True
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -130,7 +130,7 @@ class ChangeCLBNode(object):
                   'weight': self.weight})
 
 
-def _rackconnect_bulk_request(lb_node_pairs, method, success_codes):
+def _rackconnect_bulk_request(lb_node_pairs, method, success_pred):
     """
     Creates a bulk request for RackConnect v3.0 load balancers.
 
@@ -138,8 +138,8 @@ def _rackconnect_bulk_request(lb_node_pairs, method, success_codes):
         connections to be made or broken.
     :param str method: The method of the request ``"DELETE"`` or
         ``"POST"``.
-    :param iterable success_codes: Status codes considered successful for this
-        request.
+    :param success_pred: Predicate that determines if a response was
+        successful.
     :return: A bulk RackConnect v3.0 request for the given load balancer,
         node pairs.
     :rtype: :class:`Request`
@@ -151,7 +151,7 @@ def _rackconnect_bulk_request(lb_node_pairs, method, success_codes):
         data=[{"cloud_server": {"id": node},
                "load_balancer_pool": {"id": lb}}
               for (lb, node) in lb_node_pairs],
-        success_pred=has_code(*success_codes))
+        success_pred=success_pred)
 
 
 @implementer(IStep)
@@ -174,7 +174,9 @@ class BulkAddToRCv3(object):
         Produce a :obj:`Effect` to add some nodes to some RCv3 load
         balancers.
         """
-        return _rackconnect_bulk_request(self.lb_node_pairs, "POST", (201,))
+        return _rackconnect_bulk_request(
+            self.lb_node_pairs, "POST",
+            success_pred=has_code(201))
 
 
 @implementer(IStep)
@@ -195,4 +197,6 @@ class BulkRemoveFromRCv3(object):
         Produce a :obj:`Effect` to remove some nodes from some RCv3 load
         balancers.
         """
-        return _rackconnect_bulk_request(self.lb_node_pairs, "DELETE", (204,))
+        return _rackconnect_bulk_request(
+            self.lb_node_pairs, "DELETE",
+            success_pred=has_code(204))
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -200,3 +200,16 @@ class BulkRemoveFromRCv3(object):
         return _rackconnect_bulk_request(
             self.lb_node_pairs, "DELETE",
             success_pred=has_code(204))
+
+
+def _rcv3_delete_successful(lb_node_pairs, response):
+    """
+    Checks if the RCv3 bulk deletion command was successful.
+
+    This means either the response code indicated unambiguous success
+    """
+    if response.code == 204:
+        return True
+
+    fatal_errors = []
+    return bool(fatal_errors)
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -241,7 +241,7 @@ _RCV3_NODE_NOT_A_MEMBER_PATTERN = re.compile(
     re.IGNORECASE)
 
 
-def _rcv3_check_bulk_delete(lb_node_pairs, response, body):
+def _rcv3_check_bulk_delete(lb_node_pairs, result):
     """
     Checks if the RCv3 bulk deletion command was successful.
 
@@ -254,6 +254,8 @@ def _rcv3_check_bulk_delete(lb_node_pairs, response, body):
     occurred. This is necessary because RCv3 bulk requests are
     atomic-ish.
     """
+    response, body = result
+
     if response.code == 204:
         return True
 
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -250,8 +250,8 @@ def _rcv3_check_bulk_delete(result):
     indicated unambiguous success, or the machine we're trying to
     remove is already removed anyway.
 
-    If the machine was already removed, returns an effect to try and
-    remove the remaining pairs, unless a fatal error *also*
+    If the machine was already removed, returns the next step to try
+    and remove the remaining pairs, unless a fatal error *also*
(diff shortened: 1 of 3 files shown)


lvh:
This is a WIP; I'm just putting it up to ask @radix for an opinion on the `REVIEW` comment in there.

otter/convergence/steps.py
```diff
@@ -222,4 +228,60 @@ def as_effect(self):
...
+        if not match:  # Unrecoverable error, bail!
+            return
+
+        node_id, lb_id = match.groups()
+        pairs_to_delete.append((lb_id, node_id))
+
+    # REVIEW: cc @radix The best thing to return here would be an
+    # effect; that way it would get executed immediately. However, I
+    # want something that's better to introspect for testing, and
+    # that's definitely the step, not the effect.
```
radix: I don't understand why you think the Effect isn't introspectable for testing, we do testing with Effects everywhere. The intent should support equality, so you just need to check to see if it's what you expect.

lvh: I'm not saying it's *not* introspectable, I'm saying it's better (easier) to introspect. It's a step with one attribute. A service request has a bunch of things on it which means that suddenly I'm asserting a lot of things besides "tries to remove this thing from this load balancer".

radix: Ok, the usual solution to this is to create a more abstract intent (just like we did with ServiceRequest over pure_http.Request). One thought I've been having is to turn all of our steps into real intents, and get rid of the IStep interface and as_effect method. They could just have performers that return the lower-level effects.

lvh: Okay, cool. For now, I'll just test against the effect.

lvh:
I think this is now ready for review.

otter/convergence/steps.py
```diff
@@ -222,4 +226,46 @@ def as_effect(self):
...
+        return eff.on(_rcv3_check_bulk_delete)
+
+
+_CONSTANTLY_TRUE = lambda _: True
+_UUID4_REGEX = ("[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}"
+                "-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}")
+_RCV3_NODE_NOT_A_MEMBER_PATTERN = re.compile(
+    "Node (?P<node_id>{uuid}) is not a member of Load Balancer Pool "
+    "(?P<lb_id>{uuid})".format(uuid=_UUID4_REGEX),
+    re.IGNORECASE)
```
manishtomar: :(

lvh: I don't like parsing error messages either, but this appears to be the only way to do it.

manishtomar: Yes yes. I realize that. I was just sad that this had to be done.

otter/test/convergence/test_steps.py
```diff
@@ -189,3 +199,91 @@ def test_remove_nodes_from_rcv3_load_balancers(self):
...
+             {'lb_id': 'D95AE0C4-6AB8-4873-B82F-F8433840CFF2',
+              'node_id': 'D6D3AA7C-DFA5-4E61-96EE-1D54AC1075D2'})
+        ]
+
+        for message, expected_group_dict in test_data:
+            res = match(message)
+            self.assertNotIdentical(res, None)
+            self.assertEqual(res.groupdict(), expected_group_dict)
+
+        for message in ["Load Balancer Pool {load_balancer_pool_id} is not "
```
radix: Did you accidentally leave out an additional string in this list? If not then I guess the `for` loop should be dropped

otter/convergence/steps.py
```diff
@@ -222,4 +226,46 @@ def as_effect(self):
...
+    remove is already removed anyway.
+
+    If the machine was already removed, returns the next step to try
+    and remove the remaining pairs, unless a fatal error *also*
+    occurred. This is necessary because RCv3 bulk requests are
+    atomic-ish.
+    """
+    response, body = result
+
+    if response.code == 204:  # All done!
```
radix: I'm a bit concerned about allowing _any_ HTTP code to pass and then checking them in _rcv3_check_bulk_delete, at least with the way this is currently checking them. For example, if a 500 error is raised, or a 404, then I think this code will end up getting an IndexError on the body["errors"] code below.

I would probably specify the success_pred to return True for any HTTP codes that are *expected*, so stuff like 500 errors or 404 errors just don't get processed here at all.

radix: And now I'm remembering why I put code-checking before json parsing in the pipeline in http.py. :-(

lvh: So, `has_code(204, 409)`. On it.

Is that reason empty bodies failing to parse? If so, I fixed that in a previous PR.

radix: Well, I was in particular thinking that 404s and 500s and whatnot are very unlikely to return JSON bodies. So having JSON parsing before HTTP code checking means we'll end up getting json parse errors instead of "HTTP code unexpected" errors

lvh: So, you're concerned about non-empty, non-JSON bodies?

I guess now the order doesn't really matter for me anymore anyway, because I can't do what I want in `success_pred`, so I won't be parsing the errors there. I'll just pretend 409 is a "success" which really just means "defer to the callback to tell you which effect to execute instead".

lvh:
I've changed the `success_pred` to `has_code(204, 409)`; this is now once again ready for review.

manishtomar:
On a different note (totally different PR), can `has_code` be renamed to `has_codes`?

lvh:
If you feel it should, file a ticket :)

manishtomar:
@lvh [Done](https://github.com/rackerlabs/otter/issues/983) :)

otter/test/convergence/test_steps.py
```diff
@@ -189,3 +198,91 @@ def test_remove_nodes_from_rcv3_load_balancers(self):
...
+
+class RCv3CheckBulkDeleteTests(SynchronousTestCase):
+    """
+    Tests for :func:`_rcv3_check_bulk_delete`.
+    """
+    def test_node_not_a_member_error_message_regex(self):
+        """
+        The error message regular expression correctly parses error
+        messages stating the node wasn't a member of a load
+        balancer. It correctly rejects messages saying anyhting else.
```
cyli: Nitpick: "anything"

otter/test/convergence/test_steps.py
```diff
@@ -189,3 +198,91 @@ def test_remove_nodes_from_rcv3_load_balancers(self):
...
+        for message, expected_group_dict in test_data:
+            res = match(message)
+            self.assertNotIdentical(res, None)
+            self.assertEqual(res.groupdict(), expected_group_dict)
+
+        for message in ["Load Balancer Pool {load_balancer_pool_id} is not "
+                        "in an ACTIVE state"]:
+            self.assertIdentical(match(message), None)
+
+    def _rcv3_check_bulk_delete(self, resp, body):
```
cyli: This function seems to just wrap (resp, body) in a tuple.  Can `_rcv3_check_bulk_delete` just be called directly from the test functions?

otter/convergence/steps.py
```diff
@@ -222,4 +226,47 @@ def as_effect(self):
...
+        return
+
+    pairs_to_delete = []
+    for error in body["errors"]:
+        match = _RCV3_NODE_NOT_A_MEMBER_PATTERN.match(error)
```
cyli: First, a question - do we actually want to wedge a group if the load balancer is not active or has been deleted?  In the case of the load balancer being deleted, it seems like our task may be complete (no load balancer, we can go ahead and delete the servers).  The way we handle CLBs currently is that if, during removing a server from a load balancer, we find out that the load balancer has been deleted we count that as a success.

Secondly, bulk delete can bulk delete from any number of load balancers.  This doesn't address the case of if there are operations across different RCv3 load balancers for a particular group.  It just bails as soon as a single non-matching pattern is found, whereas the deletes for the other load balancer may be doable. 

A couple examples of multiple load balancer per group may be (1) if the user just has 2 RCv3 load balancers for their group, or (2) if the user has several generations of servers, each on an different RCv3 load balancer, that they are now deleting.

The question of what should happen if a load balancer isn't active, or one load balancer may not be active, may be complex, so I'm also ok with filing a ticket about this and handling it later.

otter/convergence/steps.py
```diff
@@ -222,4 +226,47 @@ def as_effect(self):
...
+        if not match:  # Unrecoverable error, bail!
+            return
```
cyli: Note: this question was on a previous (wrong) line so I deleted that comment and moved it here.

First, a question - do we actually want to wedge a group if the load balancer is not active or has been deleted?  In the case of the load balancer being deleted, it seems like our task may be complete (no load balancer, we can go ahead and delete the servers).  The way we handle CLBs currently is that if, during removing a server from a load balancer, we find out that the load balancer has been deleted we count that as a success.

Secondly, bulk delete can bulk delete from any number of load balancers.  This doesn't address the case of if there are operations across different RCv3 load balancers for a particular group.  It just bails as soon as a single non-matching pattern is found, whereas the deletes for the other load balancer may be doable. 

A couple examples of multiple load balancer per group may be (1) if the user just has 2 RCv3 load balancers for their group, or (2) if the user has several generations of servers, each on an different RCv3 load balancer, that they are now deleting.

The question of what should happen if a load balancer isn't active, or one load balancer may not be active, may be complex, so I'm also ok with filing a ticket about this and handling it later.

otter/test/convergence/test_steps.py
lvh: Yeah, there used to be a reason for it existing, but that went away. I'll get rid of it.

otter/test/convergence/test_steps.py
lvh: Addressed in new commit, thanks!

otter/test/convergence/test_steps.py
lvh: Fixed, thanks!

lvh:
Closing this while I work on it some more.

PR #969, opened 2015-01-27
Rename desired to capacity in DesiredGroupState

### LINKED ISSUE #960: DesiredGroupState.desired is a poor attribute name
It ought to be `capacity`. After all, it's a capacity. `desired` isn't even a noun, and it's already in the class name.

cyli:
Fixes #960.

One of two renaming PRs - we seem to be using the name 'launch_config' interchangably with the actual launch config that a user submits, with both server and load balancer args, and the server part of the launch config.

I for one am confused, so I'd like to change it so that `launch_config`, `server_config`, and `lb_configs`/`desired_lbs` are  distinct and separate.  :)  (I am also not tied to those exact names)


lvh:
LGTM!

Big +1 on the fixing of launch_config and friends. I think your names are fine.

Issue #968, opened 2015-01-27
Converging load balancers should converge on that server's desired state


cyli:
It shouldn't move all servers to a new load balancer.  Only new ones.

Tasks:
- [x] Store the desired load balancer configs on the server metadata
- [x] Change gatherer so that the server's desired load balancer config is stored on the server
- [x] Change the existing code so that the desired load balancer state when converging is per server instead of global (launch config)

cyli:
Re-opening this - forgot that metadata is not stored on the servers yet:  https://github.com/rackerlabs/otter/issues/1050

PR #972, opened 2015-01-28
Pass load balancer configs in nova metadata

cyli:
This addresses the first task in #968.

Punting on draining timeout because I don't know where to get it from yet.

### DIFF
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -78,6 +78,9 @@ def get_desired_group_state(group_id, launch_config, desired):
     :param dict launch_config: Group's launch config as per
         :obj:`otter.json_schema.group_schemas.launch_config`
     :param int desired: Group's desired capacity
+
+    NOTE: Currently this ignores draining timeout settings, since it has
+    not been added to any schema yet.
     """
     server_lc = prepare_server_launch_config(
         group_id,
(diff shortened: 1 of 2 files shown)


otter/convergence/composition.py
```diff
@@ -86,8 +93,35 @@ def get_desired_group_state(group_id, launch_config, desired):
...
+
+    for config in lb_args:
+        if config.get('type') != 'RackConnectV3':
+            sanitized = keyfilter(lambda k: k in ('type', 'port'), config)
+            # provide a default type
+            sanitized.setdefault('type', 'CloudLoadBalancer')
+
+            server_args = server_args.set_in(
+                ('server', 'metadata',
+                 'rax:autoscale:lb:{0}'.format(config['loadBalancerId'])),
```
lvh: While this is in fact the correct name of our product, there's a lot of "auto_scaling" going on in the codebase. Maybe we should have a plan for transitioning those over to a consistent naming scheme?

lvh:
LGTM modulo that whole thing about consistent metadata key naming :)

otter/convergence/composition.py
cyli: I think other teams actively use this particular key (in particularly the control panel, although I don't know who else) - we can add a new key, and tell everyone that one is used now, and to stop using this one, and kill this one off after a while.

cyli: https://github.com/rackerlabs/otter/issues/973

cyli:
Oops, sorry, merge forward collapsed that convo.  Was there something else I should do with the metadata?  I filed https://github.com/rackerlabs/otter/issues/973

lvh:
Nah, that's definitely a separate ticket.

Issue #956, opened 2015-01-23
execute_convergence should have success, try again, failure semantics


lvh:
Currently, `execute_convergence` has return type `Effect bool`. The bool tells you if you should retry or not, and it picks that based on whether it executed any steps this cycle.

There's three reasons you might want to converge again or not converge anymore:

1. Convergence is done. (Stop converging.)
2. Convergence isn't going to happen; the group is wedged. (Stop converging.)
3. Convergence is happening, but we're not done yet. (Keep converging.)

Right now it returns `True` if some steps/effects were executed, `False` otherwise. However, this means that if some steps were executed, but the group is broken, it'll return `True`, and the caller will retry. That means we'll spend however many convergence cycles we're normally willing to wait for a non-broken group to converge for that broken group to converge, which we know is never going to happen.

If wedged, we should also set the group status to `ERROR`.

We can either add "figure out if wedged" behavior to `execute_convergence`, or just return `Effect [results]` instead of `Effect bool` and have some other function do it. This probably means changing all of the steps' effects. A step knows when it has failed, and it's not reasonable for that callback to know about the failure modes of all steps ever. Therefore, the return types of step effects should have something along the lines of:

```python
from twisted.python.constants import ValueConstant, Values
class Status(Values):
    """
    Constants representing success types.
    """
    OK = ValueConstant(1)
    RETRY = ValueConstant(2)
    FAIL = ValueConstant(3)
```

(These are value constants just so I can find the worst case in a group with `max`. It's fine if they're named instead.)

Suggestions:

- If a step's effect (partially) fails, it's allowed to optimistically execute some more effects. It's of type `Effect Status`, i.e., it eventually gives you a status.
- Step effects are of type `Effect (Status, [Effect])`. If none of its sibling effects (i.e. other effects passed to the same `parallel` call as part of an `execute_convergence` attempt) declare the group to be wedged, these effects are executed instead. This has the benefit of allowing us to encode graceful recovery of error conditions that we can figure out without another entire gathering cycle.

The return type of `execute_convergence` could stay `Effect bool`, or become `Effect Status`; I'm not sure it matters.

radix:
For the sake of factoring, I would suggest `execute_convergence` return `Effect Status`, which separates "how to do convergence" from "when to do convergence". Then, the "when to do convergence" layer will be responsible for marking the group as wedged when it sees a Status of `FAIL`.

[...]

PR #975, opened 2015-01-28
Step results

lvh:
First part of #956. Having the semantics means you can encode the semantics.

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -53,11 +53,29 @@ class CLBNodeType(Names):
 
 
 class ServerState(Names):
-    """Constants representing the state cloud servers can have"""
-    ACTIVE = NamedConstant()    # corresponds to Nova "ACTIVE"
-    ERROR = NamedConstant()     # corresponds to Nova "ERROR"
-    BUILD = NamedConstant()     # corresponds to Nova "BUILD" or "BUILDING"
-    DRAINING = NamedConstant()  # Autoscale is deleting the server
+    """
+    Constants representing the state of a Nova cloud server.
+    """
+
+    ACTIVE = NamedConstant()
+    """
+    Corresponds to Nova's ``ACTIVE`` state.
+    """
+
+    ERROR = NamedConstant()
+    """
+    Corresponds to Nova's ``ERROR`` state.
+    """
+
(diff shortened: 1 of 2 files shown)


lvh:
Oh, man. Lint for those comments :/

lvh:
I'm inclined to fix them in this PR, if nobody objects to using the Sphinx standard for documenting classattrs.

cyli:
Just a docstring under the attr right?  No objection.  :)

otter/test/convergence/test_model.py
```diff
@@ -177,3 +189,29 @@ def test_not_done_draining_before_timeout_if_no_connection_info(self):
...
         self.assertFalse(node.is_done_draining(now=15, timeout=30))
+
+
+class StepResultTests(SynchronousTestCase):
+    """
+    Tests for :class:`StepResult`.
+    """
+    def test_ordered(self):
+        """
+        Step results are ordered; FAILURE > RETRY > SUCCESS.
```
radix: I would hope we don't rely on this ordering. It seems like something that can easily cause confusion -- like, I don't know why they should be in that order and not another one.

lvh:
I have removed the ordering tests.

otter/convergence/model.py
```diff
@@ -14,31 +14,95 @@
 
 
 class CLBNodeCondition(Names):
-    """Constants representing the condition a load balancer node can be in"""
-    ENABLED = NamedConstant()   # Node can accept new connections.
-    DRAINING = NamedConstant()  # Node cannot accept any new connections.
-                                # Existing connections are forcibly terminated.
-    DISABLED = NamedConstant()  # Node cannot accept any new connections.
-                                # Existing connections are permitted to continue.
+    """
```
radix: I'm not going to make a blocking review comment, but for the record I think it would make more sense to put :ivar: directives into the class docstring. ( http://sphinx-doc.org/domains.html#info-field-lists ).

radix: .... or I guess `cvar` makes more sense

lvh: Made #980 to follow up.

cyli:
LGTM

PR #977, opened 2015-01-28
Run convergence from the controller

### LINKED ISSUE #925: Use convergence when executing policies for whitelisted tenants
When executing a policy, we should invoke convergence instead of the existing job-based code when a tenant has been whitelisted according to `otter.convergence.composition.tenant_is_enabled`.

radix:
This should fix #925. 

*Question*: What ZooKeeper path should we use for the convergence lock? I used `/locks/convergence/<group-id>` in this branch. The group-state modification lock is at `/locks/<group-id>`.

I'm not very happy about the implementation here, but it gets the job done.

- There's a new Converger (Service) class in otter.convergence.service. tap/api.py sets a global variable to hold an instance of it.
- controller looks up the global instance of Converger to run convergence.
- The tests use mocks and parameterization a lot.

The reason to do the global is basically just that it would take more work to thread the Converger object through all the REST code to ultimately get to the controller.


otter/convergence/service.py
```diff
@@ -0,0 +1,54 @@
...
+        """Get a ZooKeeper-backed lock for converging the group."""
+        path = '{}/convergence/{}'.format(LOCK_PATH, group_id)
+        return self._kz_client.Lock(path)
+
+    def converge(self, group_id, desired, launch_config,
+                 perform=perform,
+                 execute_convergence=execute_convergence):
+        """Converge a group to a capacity with a launch config."""
+        def exec_convergence():
+            desired_group_state = get_desired_group_state(
```
manishtomar: `Converger` should get desired capacity and launch config from CASS at the beginning of the convergence cycle as it can change in the process. @lvh was working on that in #956 (I think). Not sure if you want to do it in this PR.

radix: I figure we can add cycling to the Converger service after the Status codes being returned from convergence is done. (i.e. in another PR)

cyli:
@radix could you merge master in?  The "don't blow up when loadBalancers isn't specified in a convergey launch config" could also be applied to one more line.

radix:
@cyli done, thanks :)

radix:
Okay, I've done some testing on this branch and fixed a LOT of bugs (including bugs that were already there in convergence code). I've run it through its paces with a bunch of tests, and have shown that it at least creates servers correctly. Sadly none of our integration tests will work with the new codepath yet because they all rely on getting group state reflecting to check active/pending servers (which I've filed #985 about).

Anyway, I think this should be back up for review now. I realize it's gotten a bit big, since I had to fix a bunch of bugs to get the convergence stuff going... If anyone wants me to break it up, I'll do that.

lvh:
Breaking up sounds like a good idea, but we could also just make a concerted effort to merge this thing already :)

lvh:
On second thought, small PRs are pretty cool. I'll try to extract a few logical hunks into smaller PRs just to pare this down a little?

radix:
If you want to! I can also do it tomorrow.

Here's a few chunks that I can remember:

- don't assume a group entry exists in get_all_convergence_data (the change from itemgetter to a .get)
- get_clb_data fix to parse the data out of a {'loadBalancers'} dict instead of assuming you get a list
- CreateServer returns 202 not 200
- make DesiredGroupState.desired_lbs a persistent data structure (this was mostly just motivated by some unit tests where I wanted to put these objects in a dict key, but all model data should be persistent anyway)

cyli:
+1 on making `DesiredGroupState.desired_lbs` persistent - `NovaServer.desired_lbs` is a pmap, mainly to make tests pass, because we seem to be `pset`-ing things a lot in tests and such.

radix:
Moving this back to in progress while I break up some branches.

radix:
Back in review. All the other bugfixes have been fixed in other PRs and merged to master.

manishtomar:
> Question: What ZooKeeper path should we use for the convergence lock? I used /locks/convergence/<group-id> in this branch. The group-state modification lock is at /locks/<group-id>.

@radix Ideally I would prefer `/groups/<groupId>/{state_lock,converge_lock,converging}` where `lock` is group lock used to modify state, `converge_lock` used to lock when converging and `converging` being the dirty bit specifying that group is in the process of converging. However, I am not sure how it will impact existing `/locks/groupId` after deploying. It _may_ not affect convergence tenants.

radix:
I can change the new lock to `/groups/<groupId>/...`. I don't know what you mean by "converge_lock" vs "converging", though. which one is referring to the lock introduced in this branch?

manishtomar:
> I don't know what you mean by "converge_lock" vs "converging", though. which one is referring to the lock introduced in this branch?

`converge_lock` is the lock used when converging at this moment whereas `converging` is a bit that will remain until group has reached desired state. This will be used primarily during deploy to start converging on incomplete groups.

otter/controller.py
```diff
@@ -144,25 +147,31 @@ def converge(log, transaction_id, config, scaling_group, state, launch_config,
...
+        apply_delta(log, state.desired, state, config, policy)
+        get_converger().converge(
+            log, state.tenant_id, state.group_id, state.desired, launch_config)
         return None
-    elif delta > 0:
-        execute_log.msg("executing launch configs")
-        deferred = execute_launch_config(execute_log, transaction_id, state,
-                                         launch_config, scaling_group,
-                                         delta)
     else:
```
manishtomar: Is `else` required since there is `return None` above?

otter/controller.py
```diff
@@ -268,19 +277,18 @@ def check_cooldowns(log, state, config, policy, policy_id):
     return True
 
 
-def calculate_delta(log, state, config, policy):
+def apply_delta(log, current, state, config, policy):
     """
-    Calculate the desired change in the number of servers, keeping in mind the
-    minimum and maximum constraints.
+    Calculate a new desired number of servers based on a policy, assign that
```
manishtomar: "Calculate a new desired number of servers based on a policy _on current_"? Also, will be nice to document `current`.

otter/convergence/service.py
```diff
@@ -0,0 +1,58 @@
...
+        self._reactor = reactor
+        self._kz_client = kz_client
+        self._dispatcher = dispatcher
+
+    def _get_lock(self, group_id):
+        """Get a ZooKeeper-backed lock for converging the group."""
+        path = CONVERGENCE_LOCK_PATH.format(group_id=group_id)
+        return self._kz_client.Lock(path)
+
+    def converge(self, log, tenant_id, group_id, desired, launch_config,
```
manishtomar: Do you think this can be called `set_desired_state` or something similar? We already have lot of `converge` functions in our code :)

otter/convergence/service.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, reactor, kz_client, dispatcher):
+        self._reactor = reactor
+        self._kz_client = kz_client
+        self._dispatcher = dispatcher
+
+    def _get_lock(self, group_id):
+        """Get a ZooKeeper-backed lock for converging the group."""
+        path = CONVERGENCE_LOCK_PATH.format(group_id=group_id)
+        return self._kz_client.Lock(path)
```
manishtomar: I am sorry you might have to setup acquire timeout before returning zk lock. This is different from `acquire_timeout` called below in `with_lock`. See https://github.com/rackerlabs/otter/blob/master/otter/models/cass.py#L754

otter/convergence/service.py
```diff
@@ -0,0 +1,58 @@
...
+                 perform=perform,
+                 execute_convergence=execute_convergence):
+        """Converge a group to a capacity with a launch config."""
+        def exec_convergence():
+            desired_group_state = get_desired_group_state(
+                group_id, launch_config, desired)
+            eff = execute_convergence(group_id, desired_group_state)
+            eff = Effect(TenantScope(eff, tenant_id))
+            d = perform(self._dispatcher, eff)
+            return d.addErrback(log.err)
```
manishtomar: Could you add `d.addErrback(log.err, "Failed to execute convergence cycle")` as it really helps when analyzing logs?

otter/test/test_controller.py
```diff
@@ -1079,3 +1080,26 @@ def test_audit_log_scale_down(self):
...
+
+        self.addCleanup(set_converger, get_converger())
+        set_converger(converger_mock)
+
+        result = controller.converge(log, 'txn-id', group_config, self.group,
+                                     state, 'launch', policy,
+                                     config_value=config_data.get)
+        self.assertIs(result, None)
+        converger_mock.start_convergence.assert_called_once_with(
+            log, 'tenant', 'group-id', 5, 'launch')
```
manishtomar: Could you please also check if `execute_launch_config` is not called in this case? Also, will be nice to do vice versa in other tests :) i.e. `converger` not called on regular tenants.

manishtomar:
LGTM after the last comment :). I am really excited to see this. Finally working code!!

PR #978, opened 2015-01-28
Retrieve a server's desired LB state from metadata

cyli:
When NovaServer is created, parse the server metadata to store that server's desired loadbalancer state.

Second task of #968

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -190,7 +190,7 @@ def _lb_configs_from_metadata(server):
                     desired_lbs[lbid] = CLBDescription(
                         lb_id=lbid, port=config['port'])
 
-            except Exception:
+            except (KeyError, ValueError, TypeError):
                 pass
 
     return pmap(desired_lbs)
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -1,9 +1,11 @@
 """Code related to gathering data to inform convergence."""
-
+import json
 from operator import itemgetter
 from urllib import urlencode
 
 from effect import parallel
+
+from pyrsistent import pmap
 
 from toolz.curried import filter, groupby, map
 from toolz.dicttoolz import get_in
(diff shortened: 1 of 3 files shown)


otter/convergence/gathering.py
```diff
@@ -163,19 +165,45 @@ def _private_ipv4_addresses(server):
...
+        if k.startswith(key_prefix):
+            try:
+                config = json.loads(server['metadata'][k])
+
+                if config.get('type') != 'RackConnectV3':
+                    lbid = k[len(key_prefix):]
+                    desired_lbs[lbid] = CLBDescription(
+                        lb_id=lbid, port=config['port'])
+
+            except Exception:
```
lvh: `Exception` feels a bit wide. Looks like we have `KeyError` and that's about it? The slice doesn't except, you just get an empty string, right?

lvh: (This is a nitpick, feel free to ignore)

lvh:
LGTM modulo nit, which is up to your discretion

otter/convergence/gathering.py
cyli: KeyError and probably ValueError if the json doesn't load.  Possibly TypeError if the value is not a string, but we're pretty sure it will come back as a string if it's metadata.

cyli: Will fix since I need to merge forward anyway :)

otter/convergence/gathering.py
```diff
@@ -163,19 +165,45 @@ def _private_ipv4_addresses(server):
...
+        if k.startswith(key_prefix):
+            try:
+                config = json.loads(server['metadata'][k])
+
+                if config.get('type') != 'RackConnectV3':
+                    lbid = k[len(key_prefix):]
+                    desired_lbs[lbid] = CLBDescription(
+                        lb_id=lbid, port=config['port'])
+
+            except (KeyError, ValueError, TypeError):
```
lvh: I'm wondering if this is a "put group in error state" kind of condition, since it almost certainly means the user messed with the metadata. However, that's not a call we have to make right now, so I'm creating a ticket for it.

https://github.com/rackerlabs/otter/issues/981

lvh:
LGTM whenever Jenkins feels it's good.

PR #982, opened 2015-01-28
Gh 974 fix logging treq

sam-falvo:


otter/util/logging_treq.py
```diff
@@ -18,9 +19,14 @@ def _log_request(treq_call, url, **kwargs):
 
     :param callable f: a ``treq`` request method, such as ``treq.request``, or
         ``treq.get``, ``treq.post``, etc.
+    :param log: If provided, an instance of BoundLog.
+        Defaults to ``otter.log.log`` if not provided.
```
lvh: Looks like it defaults to `otter.log.default_log`.

otter/auth.py
```diff
@@ -61,6 +61,21 @@
 from otter.util.retry import repeating_interval, retry, retry_times
 
 
+class _DoNothingLogger(BoundLog):
```
cyli: I think we already have something like this - `otter.test.utils`:  https://github.com/rackerlabs/otter/blob/master/otter/test/utils.py#L299

Would that work for these purposes?  It does call mock, so you can assert stuff about logging if you wish, but it also will eat all logging messages and you do not have to check them.  :)

sam-falvo: I don't think that's a safe thing to do.  Production code should not depend on test code.

otter/util/logging_treq.py
```diff
@@ -18,9 +18,14 @@ def _log_request(treq_call, url, **kwargs):
 
     :param callable f: a ``treq`` request method, such as ``treq.request``, or
         ``treq.get``, ``treq.post``, etc.
+    :param log: If provided, an instance of BoundLog.
```
cyli: Yay, thank you for documenting this!

otter/auth.py
cyli: True, thanks :)

cyli:
LGTM!

PR #984, opened 2015-01-29
Pass pool kwarg through to logging_treq

sam-falvo:


otter/test/test_auth.py
```diff
@@ -77,10 +77,45 @@ def test_extract_token(self):
         resp = {'access': {'token': {'id': u'11111-111111-1111111-1111111'}}}
         self.assertEqual(extract_token(resp), '11111-111111-1111111-1111111')
 
-    def test_authenticate_user(self):
+    def test_authenticate_user_with_pool(self):
```
cyli: Possibly these two tests can be split into "Validate that authenticate_user works and processes treq responses correctly", and "Validate that the pool argument is passed to treq", in which case this particular test probably does not need to replicate everything `test_authenticate_user_without_pool` is doing.

Alternately, the actual testing of the treq response processing can be factored out into a helper function rather than copy-pasted?

Not going to block on this, just an observation for the future, etc.  :)

cyli:
I know you didn't touch that line, but the linting works on lines near whatever was actually changed too.  Looks like lint is failing on line 374 being too long?

otter/test/test_auth.py
```diff
@@ -92,7 +88,7 @@ def test_authenticate_user(self):
         self.treq.post.return_value = succeed(response)
 
         d = authenticate_user('http://identity/v2.0', 'user', 'pass',
-                              log=self.log)
+                              log=self.log, pool=pool)
```
cyli: Apologies, again for the bother :|  Would it be possible to test whether `pool is None`, and if so, just don't pass pool?  That way we can test that not passing a parameter won't break, as opposed to explicitly passing it as None?

This is much neater though, thank you for the refactor!!

cyli:
LGTM modulo some merge conflicts, thanks!  I guess needs merging forward?

It'd also be awesome if changes like this could be submitted against master next time, since it seems pretty useful and might be helpful to others too!   :)

PR #988, opened 2015-01-29
DeleteServer is successful on 204, not 200

lvh:
This was previously just the default. There are more steps with bogus `success_pred` functions because they rely on the default. I'm not sure we should even have the default...


cyli:
Good catch!  LGTM.

PR #990, opened 2015-01-29
Move converge tests

cyli:
Apologies, this PR might be hard to review since the formatting and reflowing makes it harder to compare (also diff seems like it produced suboptimal output for reading :|), but this is a purely mechanical transformation (modulo some lint) that makes any convergence tests that specifically test helper functions (the load balancer helper functions specifically) call `converge` itself rather than those helper functions.

This way, we both ensure that those helper functions work and that they're called with the right arguments.  Also, if we ever have to refactor the helper functions or change their signatures, the tests need not change (hint, this is coming next :) )


lvh:
Beautiful! Big fan.

PR #991, opened 2015-01-29
Fix lb-configs in metadata so that one LB ID can have multiple LB configurations

cyli:
In my previous two PRs, I mapped LB ids to one load balancer config.

They should be mapped to a list load balancer configs instead.


lvh:
Looks like some of @radix's changes may have conflicted with this, sorry :(

lvh:
LGTM!

PR #992, opened 2015-01-29
Fix load balancer list parsing

radix:
get_clb_contents was dealing with the response from "GET .../loadbalancers" incorrectly, by assuming it was a list, but in fact the response is a dict of the form {"loadBalancers": [...]}.

I've also deleted a redundant copy of the unit tests for json_to_LBConfigs, which were moved to composition.


lvh:
LGTM!

PR #993, opened 2015-01-29
don't blow up during gathering when there are no servers in a group

radix:
If no servers were found for a group, gathering would blow up with a KeyError.


lvh:
LGTM!

PR #995, opened 2015-01-29
Don't blow up when loadBalancers is not specified in launch config

radix:
if launch config doesn't have a `loadBalancers` key in `args`, get_desired_group_state would blow up.

I also changed `DesiredGroupState` to be a persistent data structure in this branch, mostly because otherwise I would have had conflicting branches which would be annoying.

### DIFF
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -67,7 +67,7 @@ def json_to_LBConfigs(lbs_json):
         if lb.get('type') != 'RackConnectV3':
             lbd[lb['loadBalancerId']].append(CLBDescription(
                 lb_id=str(lb['loadBalancerId']), port=lb['port']))
-    return lbd
+    return dict(lbd)
 
 
 def get_desired_group_state(group_id, launch_config, desired):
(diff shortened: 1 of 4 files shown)


otter/convergence/composition.py
```diff
@@ -67,7 +67,7 @@ def json_to_LBConfigs(lbs_json):
         if lb.get('type') != 'RackConnectV3':
             lbd[lb['loadBalancerId']].append(CLBDescription(
                 lb_id=str(lb['loadBalancerId']), port=lb['port']))
-    return lbd
+    return dict(lbd)
```
lvh: Looks like you're just trying to get rid of the `defaultdict` behavior, right? I guess maybe one day this could be a pmap...

lvh:
LGTM modulo comment which is down to your discretion.

otter/convergence/composition.py
radix: Yeah, the problem was that freeze(defaultdict) (which happens later) doesn't work.

I guess I can change this to turn it into persistent data structure sooner -- I'll do that.

lvh: Awesome :) Feel free to do that in a different PR if it ends up being messy (e.g. something is relying on it being mutable)

radix:
@lvh I've made changes, can you look?

otter/convergence/model.py
```diff
@@ -124,7 +124,7 @@ class NovaServer(object):
 
 
 @attributes(['server_config', 'capacity',
-             Attribute('desired_lbs', default_factory=dict, instance_of=dict),
+             Attribute('desired_lbs', default_factory=dict),
```
lvh: Is there a reason this isn't `default_factory=pset, instance_of=PSet`? Total nitpick, feel free to ignore.

lvh:
LGTM modulo nit :)

lvh:
(also the build, of course)

otter/convergence/model.py
radix: oh, cool. I'll do that.

radix:
ok, I made even more changes now @lvh. I made it so DesiredGroupState requires a pmap and all the tests now pass in pmaps.

lvh:
LGTM!

PR #996, opened 2015-01-29
don't blow up when convergence-tenants is unspecified in the config

radix:
Trivial bugfix.

### DIFF
--- a/otter/convergence/composition.py
+++ b/otter/convergence/composition.py
@@ -49,7 +49,9 @@ def tenant_is_enabled(tenant_id, get_config_value):
     :param callable get_config_value: config key -> config value.
     """
     enabled_tenant_ids = get_config_value("convergence-tenants")
-    return (tenant_id in enabled_tenant_ids)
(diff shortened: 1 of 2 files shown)


otter/convergence/composition.py
```diff
@@ -49,7 +49,9 @@ def tenant_is_enabled(tenant_id, get_config_value):
     :param callable get_config_value: config key -> config value.
     """
     enabled_tenant_ids = get_config_value("convergence-tenants")
-    return (tenant_id in enabled_tenant_ids)
+    if enabled_tenant_ids is not None:
+        return (tenant_id in enabled_tenant_ids)
+    return False
```
lvh: I feel bad for wanting to specify a default of `[]` in `get_config_value`.

lvh:
LGTM modulo nitpick which may or may not be valid and is entirely up to you :)

radix:
Mostly I just don't feel like messing around with any code in otter.util.config, which is where the real config_value function is that's passed to this function.

PR #997, opened 2015-01-29
Add a test for RCv3 node that isn't a member of the LB we're trying to remove it from

lvh:


radix:
LGTM. I'll click the merge button when the build passes.

PR #998, opened 2015-01-29
Use server desired load balancers rather than the launch config one

cyli:
When converging.  This way we don't move old servers to new load balancers.

Last task of #968.


lvh:
LGTM!

PR #999, opened 2015-01-29
Document _rcv3_check_bulk_delete params

lvh:
Thanks, @cyli!

Return value is not yet documented because it's about to change as per #956.


cyli:
LGTM, thanks!

Issue #989, opened 2015-01-29
Fix CLB bulk APIs


lvh:
Fix CLB bulk APIs:

- [x] The AddNodesToCLB url is wrong, and also succeed on 413 because we want to just try again next convergence cycle.  Ditto with 422 in the case of duplicate nodes.
- [x] There is a bulk delete nodes API on CLB - add that one.
    - The URL is wrong - it is missing 'nodes' between the load balancer ID and the node ID
    - Try again on 413, 422 PENDING_UPDATE, and succeed on load balancers being deleted
- [ ] Optimize the bulk delete calls

### ORIGINAL DESCRIPTION:  

[The documentation](http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/Nodes-d1e2173.html) does not mention [the API we appear to use in convergence](https://github.com/rackerlabs/otter/blob/master/otter/convergence/steps.py#L114-L122), so I'm not sure it's correct.

lvh:
@cyli and I just walked through the documentation and are now convinced it's not correct, so I guess we get to replace that with a step that does actually map to an API call?

cyli:
Ok, the documentation says "Add Node", but the example json contains 3 nodes.  So it's not super clear - going to test this out and see what it actually does.

cyli:
Tested it - adding multiple nodes does work.

cyli:
Actually the URL is wrong - it needs to go to `/loadbalancers/lbid/nodes`, but that I think is the only issue.

cyli:
Also, CLB has bulk delete operations.  Going to change the title and description of this issue.

PR #1002, opened 2015-01-29
Fix the AddNodesToCLB step

cyli:
- the url is `/loadbalancer/<lbid>/nodes`
- succeed on 413 because we want to retry on the next convergence cycle
- succeed on 422 (only duplicate nodes) because we can't figure out which nodes are duplicate - wait for next convergence cycle.
- add tests

Part of #989

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -112,6 +112,10 @@ class AddNodesToCLB(object):
     """
     Multiple nodes must be added to a load balancer.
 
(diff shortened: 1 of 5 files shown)


cyli:
Oops, also the case of http status code 422 when duplicate nodes are detected.  Which should just succeed and try again next convergence cycle.

This is different than 422 the load balancer has been deleted or is in pending delete, which should fail.  :|

lvh:
LGTM!

PR #1004, opened 2015-01-29
Makefile improvements for building documentation

lvh:
This also moves more behavior into the Makefile so that Jenkins can leverage it effectively, while having as little behavior as possible living in Jenkins.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ cleandocs:
 	rm -rf htmldoc
 	rm -rf docbook/target
 
-docs: sphinxdocs
+docs: sphinxdocs docbook
 
 sphinxdocs:
 	cp -r ${DOCDIR} _builddoc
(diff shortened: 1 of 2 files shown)


cyli:
retest this please
(Changed jenkins to no longer run maven manually and just call make docbook.)

lvh:
Oh, docs doesn't build the docbook target yet. Oops.

cyli:
@lvh Actually if we build them separately we can have separate step plots on jenkins.  Although I could always make it do sphinxdoc and then docbook separately.

cyli:
retest this please

lvh:
Sure, they're still available as separate build targets if you'd like. That said, `make docs` should build all the docs, right? Also, Jenkins only appears to call `make docs` right now, so previously that only did Sphinx.

cyli:
@lvh sure :)

cyli:
LGTM!

PR #1008, opened 2015-01-29
CLB add nodes should succeed on 422 PENDING_UDPATE too

cyli:
This is also a success case for adding nodes to CLB, since it just means we need to try later also.  Nice catch @manishtomar!


lvh:
LGTM!

PR #1011, opened 2015-01-30
Kill unitgration

### LINKED ISSUE #902: Get rid of Keymaster cruft
Keymaster is an old piece of cruft that isn't used anymore.

I first tried to do this in #894, but that cut out too much.

lvh:
Also keymaster.

Refs #894. Closes #902.


manishtomar:
LGTM

PR #1014, opened 2015-01-30
Pass query parameters to treq through the Request intent.

cyli:
For bulk CLB deletes.


lvh:
LGTM assuming the build works.

PR #1017, opened 2015-01-30
Removed Key Configurations and Working with the AS API sections from the...

MariaAbrahms:
... Dev guide; consolidated information into the AS Concepts chapter and the API Ref (WADL) chapter. Added new concepts file as an xinclude to both books. Updated rev histories in both books.


wirehead:
LGTM +1.

PR #1019, opened 2015-01-30
Delete cron scale-up policy before creating cron scale-down policy

### LINKED ISSUE #958: Timing issue with scheduler test `test_system_min_max_entities_cron_style`
This is `system.schedule_policies.test_system_scheduler_update_group.UpdateSchedulerScalingPolicy.test_system_min_max_entities_cron_style` 

The failure was here:  https://as-jenkins.rax.io/job/dev-ci-integration-tests/17/console

The test creates a group with a min of 1 and max of 2.  It creates a policy that executes every minute that scales up by 3, expecting and checking that the group to scale up to 2 (due to constraints).

Then it creates a policy that scales down by 2, expecting and checking that the group scale down to 1 (due to constraints).

The problem is that from the logs, what happened was that by the time the test checked that the group had scaled down to 1, the first policy (that scales up by 3) had executed again, because it runs every minute. 

So this test can fail due to timing issues.  Proposal: delete first scale-up policy before creating the second scale-down policy.  (@lekhajee? - just want to make sure that this doesn't violate the original intent of the test)

cyli:
Or they may interact and cause test to fail.

Fixes #958 

Ran this locally - can't tell if it eliminates the random failures, but at least the test doesn't fail 100%.

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_scheduler_update_group.py
+++ b/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_scheduler_update_group.py
@@ -57,13 +57,15 @@ class UpdateSchedulerScalingPolicy(AutoscaleFixture):
         maxentities = 2
         group = self._create_group(
             cooldown=0, minentities=minentities, maxentities=maxentities)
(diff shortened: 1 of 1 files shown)


lvh:
LGTM!

PR #1020, opened 2015-01-30
Make mimic put server in building for 3 seconds for one LB test.

### LINKED ISSUE #1015: System integration LBaaS timing cloudcafe bug
`test_system_integration_lbaas.AutoscaleLbaasFixture.test_negative_create_group_with_invalid_load_balancer` sometimes fails because:

- it creates a group with minEntities = 1 and an invalid load balancer
- it waits for the server to build, attempt to be added to the load balancer, fail, and for desired capacity to go to zero, and passes.

However, it starts of checking that the desired capacity = 1, else it fails.  If it is 1, then it waits for the desired capacity to go to zero.

However, since Otter is running against Mimic, it is sometimes so fast that when the tests check the state, `desired_capacity` is already 1, so it fails.

Solution: make that server build for > 0 seconds before going to active.

cyli:
Ensure that `test_negative_scaling_group_with_invalid_load_balancer` creates a server that will remain in building for >0 seconds when running on mimic, since that's what it expects (since it expects the server to show up, and then get deleted when adding to the load balancer fails).

Fixes #1015

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_lbaas.py
+++ b/autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_lbaas.py
@@ -356,7 +356,8 @@ class AutoscaleLbaasFixture(AutoscaleFixture):
         """
         load_balancer_list = [self.lb_other_region]
         for each_load_balancer in load_balancer_list:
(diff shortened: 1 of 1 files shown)


lvh:
LGTM!

PR #1021, opened 2015-01-30
Add an optimizer for bulk deletes from CLB.

### LINKED ISSUE #989: Fix CLB bulk APIs

cyli:
Last task of #989, and hence fixes #989.

Review:  Since the code for the optimizer for both the add and removes are basically the same, I factored them out into a function that generates the optimizer.

I'm not sure if this is actually less readable or worse than just having two functions though.  Opinions?

### DIFF
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -267,20 +267,32 @@ def _optimizer(step_type):
     return _add_to_optimizers
 
 
-@_optimizer(AddNodesToCLB)
-def _optimize_lb_adds(lb_add_steps):
+def _register_bulk_clb_optimizer(step_class, attr_name):
     """
-    Merge together multiple :obj:`AddNodesToCLB`, per load balancer.
+    Merge together multiple CLB bulk steps per load balancer.  This function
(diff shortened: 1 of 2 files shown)


lvh:
Looks OK to me. Not the hugest fan of the **dict_literal, but can't think of something that's obviously better, so...

PR #1022, opened 2015-01-30
Package Identity authentication code.

sam-falvo:


lvh:
Remember that `TestHelper(self).start()` pattern we talked about a while ago? This might be helpful here, since you don't want to push out that gross pool contract on to your caller.

If, for some reason, you only want one such pool, you could have a `PoolHelper`, but I'm not sure if/why you'd want that.

sam-falvo:
I do now.  :-)  I'll cross that bridge when I get to that point, but that is definitely the goal.

otter/test/integration/lib.py
```diff
@@ -0,0 +1,53 @@
...
+writing integration tests in the context of the Otter project.
+"""
+
+
+class IdentityV2(object):
+    """This class provides a way to configure commonly used parameters
+    exactly once for any number of Identity-related API calls.
+    """
+
+    def __init__(self, auth, username, password, endpoint, pool=None):
```
manishtomar: Using [`attributes`](https://characteristic.readthedocs.org/en/stable/api.html#characteristic.attributes) decorator can help removing some of the boilerplate code. You can see examples [here](https://github.com/rackerlabs/otter/blob/master/otter/convergence/model.py).

otter/test/integration/lib.py
```diff
@@ -0,0 +1,61 @@
...
+        test/test_treq_integration.py#L60-L74 for more information.
+    """
+
+    def authenticate_user(self):
+        return self.auth.authenticate_user(
+            self.endpoint, self.username, self.password, pool=self.pool
+        )
+
+
+def find_endpoint(catalog, kind, region):
```
lvh: otter.auth [appears to have the same functionality](https://github.com/rackerlabs/otter/blob/master/otter/auth.py#L419-L451).

otter/test/integration/lib.py
```diff
@@ -0,0 +1,97 @@
...
+        :param TestResources rcs: A :class:`TestResources` instance used to
+            record the identity results.
+
+        :return: A Deferred which, when fired, returns a copy of the resources
+            given.  The :attr:`access` field will be set to the Python
+            dictionary representation of the Identity authentication results.
+        """
+
+        def record_result(r):
+            rcs.access = r
```
cyli: If we don't want the access data to be modified by accident (I think probably we don't?), maybe this could be persistent (immutable) dict instead.  `from pyrsistent import freeze`, and `r.access = freeze(r)` would make the dictionary immutable.

otter/test/integration/test_lib.py
```diff
@@ -0,0 +1,137 @@
...
+from twisted.internet.defer import Deferred
+from twisted.trial.unittest import SynchronousTestCase
+
+from otter.test.integration import lib
+
+
+class IdentityV2Tests(SynchronousTestCase):
+    """Tests for the :class:`IdentityV2` class."""
+
+    def test_missing_params(self):
```
cyli: Non-blocking - this seems like it might be more of a test of `characteristic` instead of the class?  Although I can see how you might want to test that the attributes were created correctly (without a default).

otter/test/integration/test_lib.py
```diff
@@ -0,0 +1,137 @@
...
+
+            def authenticate_user(self, endpt, user, passwd, pool=None):
+                self.pool = pool
+                return Deferred().addCallback(lambda: {})
+
+        stub = Stub()
+        lib.IdentityV2(
+            auth=stub, username="username",
+            password="password", endpoint="endpoint"
+        ).authenticate_user(rcs)
```
cyli: Non-blocking:  Possibly authenticate user doesn't need to be called here, since this is just testing the constructor?  Unless the intent is to test that `authenticate_user` does not override the pool, in which case `authenticate_user` never actually finishes because that `Deferred` is never callbacked.

Also handy:  Unless you want to specifically have the Deferred have no result yet (e.g. you are making an `assertNoResult` somewhere), using `defer.succeed({})` might be easier because you don't have to manually callback the Deferred.  It's shorthand for `d = Deferred(); d.callback({})`  :)

otter/test/integration/test_lib.py
```diff
@@ -0,0 +1,137 @@
...
+        self.assertEquals(stub.pool, 42)
+
+    def test_records_results(self):
+        """The IdentityV2 instance you create should cache its
+        results when it receives them.
+        """
+        class Stub(object):
+            def authenticate_user(self, *unused_args, **unused_kwargs):
+                return Deferred().addCallback(lambda _: "cached")
```
cyli: Same as above.  Here are equivalent statements to this:

- `return Deferred(); ... <d>.callback("cached")`
- `return succeed("cached")`  (without calling `.callback(...)` later)

The last one is generally preferred unless it's important for the Deferred to not return right away.

otter/test/integration/lib.py
```diff
@@ -0,0 +1,97 @@
...
+        return self.auth.authenticate_user(
+            self.endpoint, self.username, self.password, pool=self.pool
+        ).addCallback(record_result)
+
+
+def find_endpoint(catalog, kind, region):
+    """Locate an endpoint in a service catalog, as returned by IdentityV2.
+
+    :param dict catalog: The Identity service catalog.
+    :param str kind: The "type" of service to look for.
```
cyli: Non-blocking nitpick - please feel free to ignore this, but 'service_type' might be a more descriptive variable.

otter/test/integration/test_lib.py
sam-falvo: I include this test to provide concrete documentation of expected behavior.  You're right that, semantically, it's basically testing characteristic.  However, unit tests double as documentation (executable specifications), and this test falls under that particular use-case.

otter/test/integration/test_lib.py
sam-falvo: Thanks for the bug catch.  Yes, they should be called back.  The intent is to exercise the complete (albeit simplified) control flow to make sure that the pool doesn't get distorted.  I did need to call authenticate_user() though, because that's the only way stub.authenticate_user() gets invoked, and thus, the only (black-box) means of discovering what my pool setting was supplied as.

cyli:
Aside from freezing that data structure, LGTM :)

PR #1026, opened 2015-02-02
a type checker in otter

radix:
(in-progress while metrics.py is being fixed)

This adds type-checking to resolve_effect by allowing intent classes to declare a `intent_result_type` attribute which is passed as the second argument to `isinstance` (meaning it can be a tuple to allow for specifying union types).


lvh:
I feel an arbitrary predicate would make more sense than limiting it to just types.

radix:
@lvh good idea! thanks :)

otter/test/utils.py
```diff
@@ -592,6 +594,21 @@ def alist_get(data, key):
...
 
 
+def resolve_effect(effect, result, is_error=False):
+    """
+    Just like :func:`effect.testing.resolve_effect`, except it performs a
+    type-check on ``result`` based on a declared type in the intent's
+    ``intent_result_type`` attribute.
+    """
+    if not is_error:
+        pred = getattr(effect.intent, 'intent_result_pred', None)
```
cyli: Non-blocking - all the ones without the attribute are the ones that come from `effect`, right?  `ParallelEffect`, `Delay`, etc?  I'm wondering if it's worth checking if the intent are those in particular, to make sure that all of the intents defined in otter all have the predicate?

It may be too much trouble, but even if not, I'm fine with doing that in another PR.

cyli:
LGTM to me other than the comment above!  Please merge at will however you choose to address said comment.  :)

radix:
great idea, @cyli. I added an assertion that any intent defined within otter must have the predicate defined, which led me to discover I forgot to define a predicate for `Retry`, which the uncovered the fact that metrics.py (and its tests) need to be updated to the new service_request result format.

radix:
hold on, no, I think this is doing the wrong thing. I think metrics.py is fine... need to do some thinkin'.

radix:
Ok, so I guess this is harder than I thought it would be :-)

The problem was with my idea to have "wrapper intents" (like TenantScope and Retry) pass through their predicates to the wrapped intents. This was working sometimes but not others. It turns out it was breaking specifically for the cases where the wrapped-effect had callbacks attached -- and so, we can't _just_ rely on the checks of the intent. The value that you must give when resolving a TenantScope shouldn't just be in the shape of the result of the intent that it wraps, but the *final* result of the intent after being passed through all the callbacks.

A simple example of this is Effect(Retry(service_request(...).on(lambda r: r[1]))). This gives you an Effect of str, *not* an Effect of tuple. But my naive implementation was only considering the wrapped intent, so it would want you to resolve the Retry with a tuple, so it's wrong.

Of course I could just not implement any predicates for the wrapper-intents like TenantScope or Retry -- but this totally guts our ability to detect type errors in a lot of cases.

Sadly, I can't think of a useful way to solve this problem short of using a fully-fledged polymorphic type system, and either declaring the type signatures on all our callbacks or having type inference. Needless to say, this is far-fetched :)

So I think I'll just revert it back to the state where we check a predicate if it's there, and otherwise do nothing. In the cases where we're resolving an effect based on ServiceRequest directly, we'll get some error checking, but it won't be as thorough as I'd hoped.

cyli:
LGTM

PR #1027, opened 2015-02-03
changes to check Nova counts after scaling

kivattik:

PR #1030, opened 2015-02-04
On delete success, response is still None, so accept that in _rcv3.

cyli:
There is still a `'NoneType' object is not iterable` bug, but now it occurs in `_rcv3.py`.

If a delete is successful, `None` is still returned rather than a tuple of (response, body), for instance [here](https://github.com/rackerlabs/otter/blob/master/otter/convergence/steps.py#L386) and also if there is no retry.

I wasn't sure if we wanted the contract for `Step.as_effect()` to always be to return `(response, body)` no matter what, so I made the change just in `_rcv3.py`.  Happy to go the other way.

This fixes the new test fixes introduced in https://github.com/rackerlabs/otter/pull/1027.

### DIFF
--- a/otter/test/worker/test_rcv3.py
+++ b/otter/test/worker/test_rcv3.py
@@ -53,6 +53,9 @@ class RCv3Tests(SynchronousTestCase):
         self.dispatcher = object()
         self.request_bag = _RequestBag(dispatcher=self.dispatcher,
                                        tenant_id='thetenantid')
+        self.post_response = (StubResponse(201, {}),
+                              _rcv3_add_response_body("lb_id", "server_id"))
+        self.del_response = None
 
     def _fake_perform(self, dispatcher, effect):
         """
@@ -82,14 +85,11 @@ class RCv3Tests(SynchronousTestCase):
         if req.method == "POST":
             self.assertEqual(req.success_pred, has_code(201))
             # http://docs.rcv3.apiary.io/#post-%2Fv3%2F{tenant_id}%2Fload_balancer_pools%2Fnodes
-            response = (StubResponse(201, {}),
-                        _rcv3_add_response_body("lb_id", "server_id"))
+            return succeed([self.post_response])
         elif req.method == "DELETE":
             self.assertEqual(req.success_pred, has_code(204, 409))
             # http://docs.rcv3.apiary.io/#delete-%2Fv3%2F{tenant_id}%2Fload_balancer_pools%2Fnode
-            response = None
-
-        return succeed([response])
+            return succeed([self.del_response])
 
     def test_add_to_rcv3(self):
         """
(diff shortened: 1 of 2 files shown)


cyli:
retest this please  (weird pip issue)

radix:
@cyli the contract for the Effect returned from `IStep.as_effect` is "returns something ignored or throws an exception", as far as the actual convergence code goes. It will eventually (in #956) return some sort of status information. Nothing should rely on it returning a `(treq_response, content)` result -- though clearly that's what `_rcv3.py` does now.

This suggests to me that the interfacing between `_rcv3.py` and the step code will need to change a little bit -- since `_rcv3.py` (for some reason) cares about the HTTP response body, and `IStep.as_effect` is likely to stop returning that, we'll need to refactor some of the code so that it can get what it wants -- probably by having _rcv3.py stop using the steps, and instead just call a factored-out function that returns an Effect of RCv3 interactions (which doesn't abstract over the service_request return value).

Alternatively, maybe the _rcv3 desire for the HTTP response body is actually not warranted. I tried to trace the path that data takes through the `launch_server_v1` and `supervisor` code but I got lost :P Maybe we can just drop it?


... All that said, it's probably fine to get this merged to get things working for now, but it will need changed pretty soon.

cyli:
@radix:  Cool, thanks - so the reason it wants the body is so the result of adding a server to a load balancer, be that CLB or RCv3, can be stored in the DB for the worker to clean up when it wants to delete the server.

`otter.worker._rcv3.add_to_rc3` is called by `otter.worker.launch_server_v1.add_to_load_balancer`, which returns the response from either rcv3 or clb.

That in turn is called by `otter.worker.launch_server_v1.add_to_load_balancers`, which zips up the results of all the load balancer adds.  This in turn is called by `otter.worker.launch_server_v1.launch_server`, which returns a tuple of the server blob and all the load balancer info.

This gets returned to `otter.supervisor.SupervisorService.execute_config` which eventually stores all that information in the DB.

:|  Not a good scene.  But still necessary for worker.  Really looking forward to getting rid of that...

cyli:

cyli:
Both [integration](https://as-jenkins.rax.io/job/otter-pr-integration/38/) and [regular unit tests](https://as-jenkins.rax.io/job/otter-pr/2752/) pass.

radix:
LGTM, but could you add a TODO comment in _rcv3.py pointing at #956, and maybe a comment saying how the None case is only expected for deletions, and not for additions?

cyli:
@radix I added a comment both in the code and in #956.  But I also added an assertion, because I think if we mess up the factoring, adding to the rcv3 load balancer should just fail and all our rcv3 tests should fail.  Would you mind having another look?

cyli:

radix:
LGTM! Merge after tests pass!

Issue #1028, opened 2015-02-03
Clean up some cloudcafe test running configuration


cyli:
Tagging every single test as good for convergence or good for mimic is long and tedious, and Cloudcafe can't negatively select tests.

Add our own thing to autoskip tests based on information in the cloudcafe configs.

- [x] AutoSkip all RCv3 tests if there is no RCv3 configuration in the cloudcafe configs
- [x] skip test if mimic=True or if convergence is true - either make own decorator `@skip_if(mimic=True)` or `@skip_if(convergence=True)` or use `unittest.skipIf` and `unittest.skipUnless`
- [x] Check the config environment attribute rather than config name to see whether to inject the hard-coded autoscale URL (needed for more ci environments)

PR #1032, opened 2015-02-04
cloudcafe dev/preprod/ci testing change

cyli:
For local otter testing, don't get the autoscale endpoint from the service catalog (since we want to test our own version).

Currently this checks the name of the config file (which is stored in an environment variable by cloudcafe) for the word 'dev' or 'preprod'. 

Check an explicit configuration parameter instead for what the environment is, and inject the autoscale endpoint if the environment is not production or staging.

This makes it easier to add new CI environments without having to explicitly change the testing code.

Addresses part of #1028

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -71,13 +71,14 @@ class AutoscaleFixture(BaseTestFixture):
 
         cls.tenant_id = cls.autoscale_config.tenant_id
 
-        # If not production or staging, always use the configured server
-        # endpoint instead of what's in the service catalog.
         if cls.autoscale_config.environment in ('production', 'staging'):
             autoscale_service = access_data.get_service(
                 cls.autoscale_config.autoscale_endpoint_name)
             cls.url = autoscale_service.get_endpoint(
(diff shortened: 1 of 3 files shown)


autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -72,16 +71,17 @@ def setUpClass(cls):
 
         cls.tenant_id = cls.autoscale_config.tenant_id
 
-        env = os.environ['OSTNG_CONFIG_FILE']
-        if ('preprod' in env.lower()) or ('dev' in env.lower()):
-            cls.url = str(cls.autoscale_config.server_endpoint) + \
-                '/' + str(cls.tenant_id)
-            print(" ------ Using dev or pre-prod otter --------")
-        else:
+        # If not production or staging, always use the configured server
```
manishtomar: Minor nitpick: Do you think moving this comment to below `else:` statement will be easier to read?

manishtomar:
LGTM modulo nitpick :) (which is upto you to fix)

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: Will do, thanks for the suggestion!

PR #1033, opened 2015-02-04
Skip all of the RCv3 tests if the client is None 

cyli:
Which happens when there is not enough configuration information for rcv3 in the cloudcafe config.

Just makes this nicer if the rcv3 tests ever get run by accident on an account without rcv3.  Also makes it so that test-running commands can be the same whether or not rcv3 is supported.

I couldn't figure out how to skip the class when already in the `setUpClass` function - raising `unittest.SkipTest` just failed the whole thing, and `unittest.skipTest()` didn't work.

Addresses part of #1028 

### Update:
I  moved a bunch of stuff from `setUpClass` to the global module state in `fixtures`, and just used `@unittest.skipUnless` to check for that module state instead.

While this still skips all the tests individually when the tests are run, it seems neater, and also reduces the amount of stuff in `setUpClass`.

Would probably be better if the rcv3 client is just made in the rcv3 system tests instead of in fixtures directly, but then the rcv3 system tests may need to create a new behaviors instance then - would prefer to neaten that up in a different refactor in the future, if necessary.

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -54,6 +54,7 @@ def _make_client(access_data, service_name, region, client_cls, debug_name):
         return client_cls(url, access_data.token.id_, 'json', 'json')
     else:
         print("This account does not support {0}.".format(debug_name))
+        return None
 
 
 def _set_up_clients():
@@ -73,30 +74,30 @@ def _set_up_clients():
             "Unable to authenticate against identity to get auth token and "
             "service catalog.")
 
-    autoscale_client = _make_client(
+    _autoscale_client = _make_client(
         access_data,
         autoscale_config.autoscale_endpoint_name,
         autoscale_config.region,
         AutoscalingAPIClient,
         "Autoscale")
 
-    server_client = _make_client(
+    _server_client = _make_client(
         access_data,
         autoscale_config.server_endpoint_name,
         autoscale_config.server_region_override or autoscale_config.region,
         ServersClient,
         "Nova Compute")
 
-    lbaas_client = _make_client(
+    _lbaas_client = _make_client(
         access_data,
         autoscale_config.load_balancer_endpoint_name,
         autoscale_config.lbaas_region_override or autoscale_config.region,
         LbaasAPIClient,
         "Cloud Load Balancers")
 
-    rcv3_client = None
+    _rcv3_client = None
     if _rcv3_cloud_network and _rcv3_load_balancer_pool:
-        rcv3_client = _make_client(
+        _rcv3_client = _make_client(
             access_data,
             autoscale_config.rcv3_endpoint_name,
             autoscale_config.rcv3_region_override or autoscale_config.region,
@@ -111,7 +112,7 @@ def _set_up_clients():
         raise Exception(
             "Unable to instantiate all necessary clients.")
 
-    return (autoscale_client, server_client, lbaas_client, rcv3_client)
+    return (_autoscale_client, _server_client, _lbaas_client, _rcv3_client)
 
 
 # Global testing state - unfortunate, but only needs to be done once and also
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -58,7 +58,12 @@ def _make_client(access_data, service_name, region, client_cls, debug_name):
 
 def _set_up_clients():
     """
-    Function
+    Read the user creds from the configuration file in and constructs all the
+    service clients.  If it can't authenticate, or it cannot construct the
+    autoscale/server/lbaas clients, then it fails.
+
+    The RCv3 client is not created if the account does not have access to RCv3
+    or if RCv3 configuration parameters are not present or invalid.
     """
     user_config = UserConfig()
     access_data = AuthProvider.get_access_data(endpoint_config, user_config)
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -63,6 +63,11 @@ def _set_up_clients():
     user_config = UserConfig()
     access_data = AuthProvider.get_access_data(endpoint_config, user_config)
 
+    if access_data is None:
+        raise Exception(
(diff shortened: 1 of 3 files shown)


autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
```diff
@@ -57,6 +57,16 @@ def setUpClass(cls):  # nopep8 - ignore N802
         """
         super(AutoscaleRackConnectFixture, cls).setUpClass()
 
+        if cls.rcv3_client is None:
+            # don't bother with the rest of the set-up; skip all tests
+            skip = unittest.skip("This account does not support RCv3.")
+
+            for test in inspect.getmembers(cls, predicate=inspect.ismethod):
+                # test is a tuple of name, object
+                if test[0].startswith("test_"):
```
manishtomar: This is an interesting trick to skip tests at runtime. I wonder if there is another cleaner way to do the same. This is merely non-blocking thought :)

manishtomar:
LGTM!

autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_rcv3.py
cyli: You can call unittest.skipTest() during `setUp`, apparently, but it didn't seem to work in `setUpClass` either.  :|

I have an idea of how to skip the whole class, but it involves a lot more refactoring.  I guess let me try that...

cyli:

cyli:
This has the unfortunate side effect of not logging the initial auth call's request and response, because I guess the TestCase fixtures inherited from cloudcafe are the thing that set up logging handlers.

This also means that auth only happens once per test run now, instead of per TestCase class.

cyli:

cyli:

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -25,7 +25,104 @@
...
 autoscale_config = AutoscaleConfig()
+endpoint_config = UserAuthConfig()
+
+# Get optional RCV3 values.  These might not be present in the config
+# file.
+try:
+    _rcv3_load_balancer_pool = json.loads(
+        autoscale_config.rcv3_load_balancer_pool)
+except Exception:
+    _rcv3_load_balancer_pool = None
```
manishtomar: A print/log saying "Couldn't find RCv3 configs" can help

cyli:
[Integration](https://as-jenkins.rax.io/job/dev-ci-integration-tests/91) and [unit](https://as-jenkins.rax.io/job/otter-pr/2770/) tests pass.

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: I added, but in a different place (where i'm actually checking for it), if that's ok.  :)

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -25,7 +25,106 @@
...
+
+    if url is not None:
+        return client_cls(url, access_data.token.id_, 'json', 'json')
+    else:
+        print("This account does not support {0}.".format(debug_name))
+
+
+def _set_up_clients():
+    """
+    Function
```
manishtomar: Ha ha ha :) This is the best docstring I've seen :smile:

autoscale_cloudroast/test_repo/autoscale/fixtures.py
manishtomar: +1 :)

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: Hahaha!  Oops.  Will fix :D

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -25,7 +25,111 @@
...
+    lbaas_client = _make_client(
+        access_data,
+        autoscale_config.load_balancer_endpoint_name,
+        autoscale_config.lbaas_region_override or autoscale_config.region,
+        LbaasAPIClient,
+        "Cloud Load Balancers")
+
+    rcv3_client = None
+    if _rcv3_cloud_network and _rcv3_load_balancer_pool:
+        print("Not enough test configuration for RCv3 provided. "
```
manishtomar: Did you intend to put this in `else` clause?

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -25,7 +25,111 @@
...
+
+    else:
+        service = access_data.get_service(service_name)
+        if service is not None:
+            url = service.get_endpoint(region).public_url
+
+    if url is not None:
+        return client_cls(url, access_data.token.id_, 'json', 'json')
+    else:
+        print("This account does not support {0}.".format(debug_name))
```
manishtomar: Non-blocking nitpick: A `return None` after this will be nice to be explicit.

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -25,7 +25,111 @@
...
+    """
+    user_config = UserConfig()
+    access_data = AuthProvider.get_access_data(endpoint_config, user_config)
+
+    if access_data is None:
+        raise Exception(
+            "Unable to authenticate against identity to get auth token and "
+            "service catalog.")
+
+    autoscale_client = _make_client(
```
manishtomar: Can you have different names of client variables that don't match global variable names? This can be `_autoscale_client` to emphasize the localness

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: Nice catch, yes, thanks!

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: Done.  :)

manishtomar:
LGTM :) Merging

PR #1034, opened 2015-02-04
Automatically skip mimic tests

cyli:
Based on the config, not tags, using `@unittest.skipIf`.

Convergence-incompatible tests can also be skipped this way.


manishtomar:
LGTM

PR #1036, opened 2015-02-05
Add automatic destruction of scaling groups.

### LINKED ISSUE #1035: Clean up scaling group resources.

sam-falvo:
Fixes #1035 .

### DIFF
--- a/otter/test/integration/lib.py
+++ b/otter/test/integration/lib.py
@@ -10,7 +10,7 @@ from pyrsistent import freeze
 
 import treq
 
-from otter.util.http import check_success
+from otter.util.http import check_success, headers
 
 
 @attributes([
@@ -119,10 +119,7 @@ class ScalingGroup(object):
 
         return (treq.delete(
             "%s/groups/%s" % (str(rcs.endpoints["otter"]), self.group_id),
-            headers={
-                'X-Auth-Token': str(rcs.token),
-                'Accept': 'application/xml',
-            },
+            headers=headers(str(rcs.token)),
             pool=self.pool
         ).addCallback(check_success, [204, 404]))
 
@@ -145,17 +142,20 @@ class ScalingGroup(object):
                 "%s/groups/%s/state" % (
                     str(rcs.endpoints["otter"]), self.group_id
                 ),
-                headers={
-                    'X-Auth-Token': str(rcs.token),
-                    'Accept': 'application/xml',
-                },
+                headers=headers(str(rcs.token)),
                 pool=self.pool
             ).addCallback(check_success, [200, 404])
             .addCallback(decide)
         )
 
     def start(self, rcs, test):
-        """Create a scaling group."""
+        """Create a scaling group.
+        
+        :param TestResources rcs: A set of OpenStack resources encapsulated
+            in a TestResources instance.
+        
+        :return: The same instance of TestResources.
+        """
 
         test.addCleanup(self.stop, rcs)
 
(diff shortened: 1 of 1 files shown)


otter/test/integration/lib.py
```diff
@@ -81,6 +87,118 @@ def record_result(r):
...
+        """Clean up a scaling group.  Although safe to call yourself, you
+        should think twice about it.  Let :method:`start` handle registering
+        this function for you.
+
+        At the present time, this function DOES NOT stop to verify
+        servers are removed.  (This is because I haven't created
+        any tests which create them yet.)
+        """
+
+        return (self.get_scaling_group_state(rcs)
```
cyli: Non-blocking:  this is totally fine, but destroying a scaling group in autoscale is idempotent (it's a bug if it isn't).
So probably you can just delete it without checking state first, and either a 404 or a 204 are ok?

otter/test/integration/lib.py
```diff
@@ -81,6 +87,100 @@ def record_result(r):
...
+                'Accept': 'application/xml',
+            },
+            pool=self.pool
+        ).addCallback(check_success, [204, 404]))
+
+    def get_scaling_group_state(self, rcs):
+        """Retrieve the state of the scaling group.
+
+        :return: A :class:`Deferred` which, upon firing, returns the result
+            code and, optionally, scaling group state as a 2-tuple, in that
```
cyli: Just a doc nitpick - the body result doesn't seem to be optional.  :)

otter/test/integration/lib.py
```diff
@@ -81,6 +87,100 @@ def record_result(r):
...
+                    'X-Auth-Token': str(rcs.token),
+                    'Accept': 'application/xml',
+                },
+                pool=self.pool
+            ).addCallback(check_success, [200, 404])
+            .addCallback(decide)
+        )
+
+    def start(self, rcs, test):
+        """Create a scaling group."""
```
cyli: Doc nitpick - might be nice to document the expected return value (the result of creating a group, if creating a group was successful, and an API failure if the creation was unsuccessful).

otter/test/integration/lib.py
```diff
@@ -81,6 +87,100 @@ def record_result(r):
...
+            self.group_id = str(resp["group"]["id"])
+            return rcs
+
+        return (
+            treq.post(
+                "%s/groups" % str(rcs.endpoints["otter"]),
+                json.dumps(self.group_config),
+                headers={
+                    'X-Auth-Token': str(rcs.token),
+                    'Content-Type': 'application/xml',
```
cyli: Should headers be `'key': ['values']` instead of `'key': 'value'`?  I vaguely remember us trying it out with just a string yesterday and it not working.

Also, this should probably be JSON instead of xml (here and in the other functions), but autoscale ignores the content-type and accept headers anyway.

`otter.util.http.headers` also provides a nice method of generating headers given just the token (it adds the content-type, accept, and a user agent).

radix: Repro for header values as strings:

```
>>> from __future__ import print_function; import treq; from twisted.internet import reactor
>>> d = treq.get('https://httpbin.org/get', headers={'X-Foo': 'Bar'}).addCallback(treq.content).addCallback(print).addErrback(print)
>>> reactor.run()
{
  "args": {},
  "headers": {
    "Accept-Encoding": "gzip",
    "Host": "httpbin.org",
    "X-Foo": "Bar"
  },
  "origin": "108.245.70.37",
  "url": "https://httpbin.org/get"
}
```

cyli: Thanks :)

sam-falvo: <_<

>_>

You saw nothing.

cyli:
LGTM - please merge once lint issues are fixed.  :)

PR #1041, opened 2015-02-09
CreateServer success semantics

lvh:
This is a very simple PR, just to test the waters. Quite simply, if the service request results in any exception, `RETRY`, else `SUCCESS`, and no reasons either way. We can add smarter behavior (e.g. fail fast, or more reasons) in other PRs.

Refs #956.

### DIFF
--- a/otter/test/convergence/test_steps.py
+++ b/otter/test/convergence/test_steps.py
@@ -9,7 +9,7 @@ from pyrsistent import freeze, pset
 from twisted.trial.unittest import SynchronousTestCase
 
 from otter.constants import ServiceType
-from otter.convergence.model import CLBDescription
+from otter.convergence.model import CLBDescription, StepResult
 from otter.convergence.steps import (
     AddNodesToCLB,
     BulkAddToRCv3,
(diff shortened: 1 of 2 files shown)


lvh:
jenkins run integration tests please

otter/test/convergence/test_steps.py
```diff
@@ -43,19 +44,32 @@ def test_create_server(self):
...
             service_request(
                 ServiceType.CLOUD_SERVERS,
                 'POST',
                 'servers',
                 data={'server': {'name': 'myserver-random-name',
                                  'flavorRef': '1'}},
-                success_pred=has_code(202)))
+                success_pred=has_code(202)).intent)
+
+        self.assertEqual(
```
cyli: Non-blocking - wondering if failure should be a different test?  I realize it's pretty short - changing the docstring to be more clear about exactly what this test does, and maybe providing a message to `assertEqual` would probably be sufficient too.

cyli:
This LGTM insofar as I understand https://github.com/rackerlabs/otter/issues/956  - wondering if issues should be openned about next steps?  (fail fast, as you said, etc)?

lvh:
Yep!

PR #1042, opened 2015-02-09
Use LB Description/Node interface methods and attributes in convergence

cyli:
Rather than the CLB-specific ones.

Getting ready for working on #821


lvh:
LGTM!

PR #1046, opened 2015-02-09
Fix some of the remove/draining CLB tests

cyli:
To ensure that the tests weren't just passing because no desired load balancers on the server were passed.


cyli:
Note - the other build is because of jenkins ugliness - the travis one should be the only source of truth

radix:
LGTM! :)

PR #1047, opened 2015-02-09
Update Makefile for Travis

cyli:
Travis also uses the merge commit and needs to use HEAD^1, so update the Makefile to support that.

Also, make docbook if there are docbook changes.


cyli:
Note - the other build is because of jenkins ugliness - the travis one should be the only source of truth

manishtomar:
+1

PR #1051, opened 2015-02-10
Clean up groups.py

lvh:


otter/rest/groups.py
```diff
@@ -160,14 +177,15 @@ def format_list(group_states):
...
-    # --------------------------- CRD a scaling group -----------------------------
+    # -------------------------- CRD a scaling group -------------------------
     # (CRD = CRUD - U, because updating happens at suburls - so you can update
     # different parts)
 
-    # TODO: Currently, the response does not include scaling policy ids, because
-    # we are just repeating whatever the request body was, with an ID and links
-    # attached.  If we are going to create the scaling policies here too, we should
-    # probably also return their ids and links, just like the manifest.
+    # TODO: Currently, the response does not include scaling policy ids,
```
cyli: Apologies, I know all you did was reflow this, but this TODO is no longer applicable (we return ids and links).  Can it be removed in this PR, or should there just be another PR?

lvh: Sure!

lvh: Also plus one data point for issue trackers :)

otter/test/rest/test_groups.py
```diff
@@ -833,26 +865,33 @@ def test_view_manifest_with_webhooks(self):
                 }
             ]
         ]
-        webhooks_internal_links = [
-            [[{"href": '/v1.0/11111/groups/one/policies/5/webhooks/3/', "rel": "self"},
+        webhooks_links = [
```
cyli: This seems to change the test - it now declares `webhooks_links`, and then overwrites them with another copy.

This test is intended to test that the manifest lists all the webhooks for the policies (which was declared in `webhook_internal_links`), but also list a link to the webhooks resource for that policy (which was `webhooks_link`).

otter/test/rest/test_groups.py
```diff
@@ -860,7 +899,7 @@ def test_view_manifest_with_webhooks(self):
         for i in [0, 1]:
             exp_policies[i]['webhooks'] = deepcopy(webhooks[i])
             for j, webhook in enumerate(exp_policies[i]['webhooks']):
-                exp_policies[i]['webhooks'][j]['links'] = webhooks_internal_links[i][j]
```
cyli: As above, this should probably remain `webhooks_internal_links` (or be renamed to a different variable name :))

lvh:
Thanks, @cyli, good catch! I've fixed that with a manual line break instead of renaming the variable.

otter/test/rest/test_groups.py
lvh: Github is still showing this comment because it's attached to the deleted line, and the line is still deleted. However, the problem (name change) has been addressed.

cyli:
LGTM!

PR #1052, opened 2015-02-10
DeleteServer success semantics

lvh:
Refs #956.

Other tickets will need to be created in order to cover all of the edge cases for this.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -93,11 +93,19 @@ class DeleteServer(object):
 
     def as_effect(self):
         """Produce a :obj:`Effect` to delete a server."""
-        return service_request(
+        eff = service_request(
             ServiceType.CLOUD_SERVERS,
             'DELETE',
             append_segments('servers', self.server_id),
             success_pred=has_code(204))
+
+        def report_success(result):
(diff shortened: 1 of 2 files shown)


otter/convergence/steps.py
```diff
@@ -93,12 +93,20 @@ class DeleteServer(object):
...
     def as_effect(self):
         """Produce a :obj:`Effect` to delete a server."""
-        return service_request(
+        eff = service_request(
             ServiceType.CLOUD_SERVERS,
             'DELETE',
             append_segments('servers', self.server_id),
             success_pred=has_code(204))
 
+        def report_success(result):
```
manishtomar: Can this section be written as a decorator and `as_effect` can be decorated with that since most of the functions will have similar behavior?

lvh: Perhaps, eventually, but not in this PR. It is clear that we don't actually know what they're going to end up looking like; most steps need very distinct things. I want to know the actual requirements before I refactor it into something we can't actually use.

lvh:
@manishtomar Other than your comment, does this look good to you?

manishtomar:
LGTM :)

Issue #1054, opened 2015-02-10
Cover remaining error response codes for CLB update


lvh:
[Documentation](http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173.html)

Error response codes: loadbalancerFault (400, 500), serviceUnavailable (503), badRequest (400), unauthorized (401), overLimit (413), itemNotFound (404)

It's kind of ridiculous that 400 is occasionally a load balancer fault and occasionally just a bad request. We don't have to deal with unauthorized; that's the authenticator's problem.

manishtomar:
> It's kind of ridiculous that 400 is occasionally a load balancer fault and occasionally just a bad request.

Its most likely a documentation issue. As far as I've seen, 400 comes only on bad request. But we should confirm from them.

manishtomar:
Asked question to CLB team

manishtomar:
400 code mentioned in docs is actually 422 IMMUTABLE fault which is being handled by `cloud_client.py`. Closing.

PR #1055, opened 2015-02-10
ChangeCLBNode is successful on 202, not 200

lvh:
This doesn't cover all the codes; that's #1054. This is just for ensuring super-basic correctness.


radix:
LGTM :-)

PR #1060, opened 2015-02-11
moving `next_cron_occurrence` to models.interface

manishtomar:
This is because cass importing scheduler causes cyclic import issues when having CQL intent in cass.py (that will be done in next PR). 

FYI: Cron scheduler tests pass against mimic.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -35,8 +35,8 @@ from otter.models.interface import (
     PoliciesOverLimitError,
     ScalingGroupOverLimitError,
     UnrecognizedCapabilityError,
-    WebhooksOverLimitError)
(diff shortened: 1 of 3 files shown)


cyli:
LGTM, although wondering if there should be tests for `next_cron_occurrence`?  I couldn't find any ones specifically to test that function, but it does seem to be mostly whether `croniter` works.

I'll leave it up to you - please merge if you think it's fine.  :)

manishtomar:
Thanks @cyli. I think we can skip tests for now :) It's also mainly if `croniter` works. OTOH, a black box test that gives cron string and expects the correct date would be nice. I've created #1068 to write it.

Issue #1059, opened 2015-02-10
Do whatever it takes to get test_system_server_details_name_and_metadata to work


radix:
(updated based on Ying's comment)

it lives in autoscale_cloudroast/test_repo/autoscale/system/group/test_system_launch_config.py.

It does these things:

- creates a group with 1 min entities
- checks to see if pending + active according to the get group state call is 1
- looks in nova for the servers and makes sure they have the right metadata (no lb metadata checks though), the right image, the right flavor, and the right name

cyli:
I don't think that one actually checks nova - I think it just checks autoscale.  Possibly `test_system_server_details_name_and_metadata`, which:

- creates a group with 1 min entities
- checks to see if pending + active according to the get group state call is 1
- looks in nova for the servers and makes sure they have the right metadata (no lb metadata checks though), the right image, the right flavor, and the right name

radix:
Thanks Ying, there's still a lot I don't know about those tests :) I'll update the description.

radix:
This is fixed now.

PR #1061, opened 2015-02-11
Tag one of the system group tests convergence=p1 (for phase 1)

cyli:
Silly change.  :)  To make it easier to run #1059


radix:
Is there a reason you tagged it with the phase instead of just something like "yes"? is there some value in separating the tests run by phase? anyway LGTM :)

cyli:
@radix Hmm... I guess not.  I'll change it to "yes".  :)

PR #1062, opened 2015-02-11
fixes for some bugs in to_nova_server

radix:
- server status is in the "status" key, not "state"
- parsed json data gives us unicode, not str, so loosen the type check in NovaServer.servicenet_address.


cyli:
LGTM!

Issue #793, opened 2014-11-26
RackConnectV3 Integration Testing


kivattik:
Create a set of tests for RackConnect V3

Tasks:
- [ ] 1.) Create configuration to run against RackConnect account
  - [ ] 1a) Document new config file elements and the changes made to support them
        (In elaboration comment - list new values, their default values and dependencies) 
         ex. rcv3_cloud_network, default not present, add to config.py and fixtures.py) 
  - [x] 1b.) Create config file for Mimic
  - [x] 1c.) Create config file for Hardware
  - [ ] 1d.) Add config file generation to autoscaling-chef
  - [x] 1e.) Modify autoscale cloudcafe to read in new config values

- [ ] 2.) Implement RCV3 API client in autoscale_cloudcafe
  *RackConnect API methods to implement*
  - [x] 2a.) List all load balancer pools on the account
    - [x] RackConnectLBPool response object and marshalling
    - [x] RackConnectLBPools response object and marshalling
  - [x] 2b.) Get info on a specific LB Pool (uses RackConnectLBPool response object)
  - [x] 2c.) List nodes on a LB Pool
    - [x] RackConnectLBNode response object
    - [x] RackConnectLBNodes response object
  - [x] 2d.) Get Node Info Detail  - This is a superset of Get Node Info
    - [x] Add response objects
  - [x] 2e.) Remove Node
    - [x] TBD request/response objects
  - [ ] 2f.) _Get Node Info [Optional]_ This is not likely needed since all info is also in detail view
  - [ ] 2g.) _List Nodes with detail [Optional]_ We shouldn't need any of the detail info for testing
  - [ ] 2h.) _Add Node [Optional]_ Not needed for testing at this scope

- [ ] 3.) Test Cases

*Create Group*

- [x] 1.) Create group with the following:
        - min_entities = 0
        - Single RackConnectV3 type load balancer with valid pre-created pool_id
        - RackConnect internal network only

        Verify:
        - Create status is OK
        - Create response is 201
        - launchConfig for the group contains the correct LB uuid & type

- [x] 2.) Create group with the following:
        - min_entities = 1
        - Single RackConnectV3 loadbalancer with valid uuid
        - RackConnect internal network only

        Actions:
        - In setUp - get initial node_counts on lb pool
        - wait for expected active servers (OR) FAIL
        - get current node_counts

        Verify:
        - verify cloud_server node count increased by one
        - verify total count increased by one
        - verify external count did not change

[...]

PR #1064, opened 2015-02-11
Deal with missing RCv3 load balancer pools

### LINKED ISSUE #1012: Deal with missing RCv3 load balancer pools
RCv3 has a currently undocumented error condition for bulk add/delete operations, when a load balancer pool being referenced doesn't exist. RCv3 tells us this error message is:

`Load Balancer Pool {lb pool id} does not exist`

Otter should handle it.

lvh:
Closes #1012.


radix:
This LGTM. I added a comment on #793 about dealing with this case in tests.

PR #1065, opened 2015-02-11
Bare RCv3 request API

lvh:
In worker, RCv3 steps want to see the parsed request body. In convergence, it should return `Status, [reasons]` as per #956. Previously, there were some hacks in the RCv3 worker code to work around this, but that would stop working as soon as we try to make the return type of the RCv3 bulk add step also be `Effect (Status [reasons])`. This removes that hack by introducing a new private API, returning the bare request effect (of type `Effect (Response, ParsedResponseBody)`) without the callbacks attached.

This should be merged after #1064.


radix:

radix:
LGTM!

Issue #1039, opened 2015-02-06
Move all webhook lookups to the new table


manishtomar:
Currently, webhook lookup from hash tries from new `webhook_keys` table and defaults to using index on `policy_webhooks` table. Using the index is incorrect usage as per [docs](http://www.datastax.com/documentation/cql/3.0/cql/ddl/ddl_when_use_index_c.html?scroll=concept_ds_sgh_yzz_zj__highCardCol). We need to
- [ ] Adding all webhooks in `webhook_keys` table
- [ ] Stop using `policy_webhooks` in code to lookup based on hash
- [ ] Delete `webhooks_by_token` index

PR #1069, opened 2015-02-11
Webhook migrate script

manishtomar:
First task of #1039. This creates `CQLQueryExecute` intent along with its performer and dispatcher to migrate the webhooks. Would like to know if having separate `otter.effect_dispatcher.get_cql_dispatcher` is a good idea @radix ?

I've tested this from dev vm by copying data from SYD prod.

### DIFF
--- a/otter/effect_dispatcher.py
+++ b/otter/effect_dispatcher.py
@@ -2,15 +2,10 @@
 
 from functools import partial
 
-from cql.connection import connect
-
 from effect import (
     ComposedDispatcher,
-    ParallelEffects,
     TypeDispatcher,
-    base_dispatcher,
-    sync_perform,
-    sync_performer)
+    base_dispatcher)
 from effect.twisted import make_twisted_dispatcher
 
 from .auth import (
--- a/otter/effect_dispatcher.py
+++ b/otter/effect_dispatcher.py
@@ -2,7 +2,9 @@
 
 from functools import partial
 
-from effect import ComposedDispatcher, TypeDispatcher, base_dispatcher
+from cql.connection import connect
+
+from effect import ComposedDispatcher, TypeDispatcher, base_dispatcher, sync_performer
 from effect.twisted import make_twisted_dispatcher
 
 from .auth import (
@@ -12,6 +14,7 @@ from .auth import (
     perform_invalidate_token,
 )
 from .http import TenantScope, perform_tenant_scope
+from .models.cass import CQLQueryExecute
 from .util.pure_http import Request, perform_request
 from .util.retry import Retry, perform_retry
 
@@ -48,3 +51,35 @@ def get_full_dispatcher(reactor, authenticator, log, service_config):
                                  service_config)}),
         get_simple_dispatcher(reactor),
     ])
+
+
+def get_sync_cql_dispatcher(cursor):
+    """
+    Get dispatcher with `CQLQueryExecute`'s synchronous performer in it
+
+    :param cql: CQL cursor
+    """
+    return ComposedDispatcher([
+        base_dispatcher,
+        TypeDispatcher({
+            CQLQueryExecute: partial(perform_query_sync, cursor),
+            ParallelEffects: perform_serial
+        })
+    ])
+
+@sync_performer
+def perform_serial(disp, intent):
+    """
+    Performs parallel effects serially. Useful when testing or simple cases
+    """
+    return map(partial(perform, disp), intent.effects)
+
+
+@sync_performer
+def perform_query_sync(cursor, disp, intent):
+    """
+    Perform CQLQueryExecute synchronously using cursor
+    """
+    return cursor.execute(
+        intent.query, intent.params,
+        consistency_level=intent.consistency_level)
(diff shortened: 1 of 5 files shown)


otter/effect_dispatcher.py
```diff
@@ -32,7 +36,7 @@ def get_simple_dispatcher(reactor):
             Authenticate: perform_authenticate,
             InvalidateToken: perform_invalidate_token,
             Request: perform_request,
-            Retry: perform_retry,
+            Retry: perform_retry
```
radix: That trailing comma was intentional (exactly so diffs would be simpler), could you revert this change?

radix:
get_cql_dispatcher seems fine to me :)

otter/test/models/test_cass_models.py
```diff
@@ -106,6 +114,28 @@ def _cassandrify_data(list_of_dicts):
...
+    """
+    Tests for :func:`perform_cql_query` function
+    """
+
+    def test_perform_query_sync(self):
+        """
+        Calls given connection's execute
+        """
+        conn = mock.Mock(spec=CQLClient)
+        conn.execute.return_value = 'ret'
```
radix: Could you change this to be a defer.succeed('ret')? this test could succeed without the @deferred_performer decorator, but changing the return value to succeed('ret') should make it test that it's really dealing with deferreds correctly.

otter/models/cass.py
```diff
@@ -1486,6 +1510,46 @@ def _do_webhook_lookup(webhook_rec):
...
+                                params={},
+                                consistency_level=ConsistencyLevel.ONE))])
+        return eff.on(
+            lambda (whooks, wkeys): set(freeze(whooks)) - set(freeze(wkeys)))
+
+    def add_webhook_keys(self, webhook_keys):
+        """
+        Add webhook keys to webhook_keys table
+
+        :return: Effect of CQLQueryExecute
```
radix: The "Effect of *" idiom should generally be used to describe the *result* of the Effect, not the intent of the effect. So, in this case it'd be an Effect of... whatever connection.execute() will return in this case. (None?)

otter/models/cass.py
```diff
@@ -1486,6 +1510,46 @@ def _do_webhook_lookup(webhook_rec):
         d.addCallback(_do_webhook_lookup)
         return d
 
+    def get_webhook_index_only(self):
+        """
+        Get webhook info that is there in webhook index but is not there in
+        webhook_keys table
+
+        :return: Effect of CQLQueryExecute
```
radix: Same `Effect of *` comment here.

otter/test/models/test_cass_models.py
```diff
@@ -2527,6 +2557,68 @@ def test_add_scaling_policy_cron(self):
...
+
+    def setUp(self):
+        """
+        Sample CassScalingGroup object
+        """
+        self.store = CassScalingGroupCollection(None, None)
+
+    def test_webhook_index_only(self):
+        """
+        Test `get_webhook_index_only`
```
glyph: What aspect of `get_webhook_index_only` is this testing?  `"w2": "2"` is not the most self-documenting of examples :).

manishtomar: Well.. I would just have to repeat the docstring of `get_webhook_index_only` here to make sense. Do you think its ok to put `Refer get_webhook_index_only docs` here?

glyph: I prefer to avoid that; the test is always testing something more nuanced than the public documentation.  In this case, one obvious distinction is that `get_webhook_index_only`'s docstring does not note that the effects take place in parallel, yet this test cares and will fail if the effect's intent is sequential rather than parallel.

manishtomar: Hmm.. OK. I didn't think from that aspect. I'll change the docstring.

otter/test/models/test_cass_models.py
```diff
@@ -106,6 +114,28 @@ def _cassandrify_data(list_of_dicts):
     return _de_identify(list_of_dicts)
 
 
+class PerformTests(SynchronousTestCase):
+    """
+    Tests for :func:`perform_cql_query` function
+    """
+
+    def test_perform_cql_query(self):
```
cyli: +1 for actually testing perform :D

cyli:
I feel like I'm somewhat shaky on effect, so I don't feel super comfortable reviewing the dispatcher + tests, but the rest of it LGTM.  @radix maybe?

My non-blocking concern is I'm not sure how many webhooks are there in the webhooks table - I dunno if grabbing all of it at once would cause any problems.

If you've already tested against our biggest DCs then it it's fine as is, otherwise maybe we can do this on a per-tenant basis?  (Although that can be done in a different PR).

radix:
The effect/dispatcher/performer stuff looks fine to me. I'm +1 on merging once cyli's concern about data size is addressed.

manishtomar:
> My non-blocking concern is I'm not sure how many webhooks are there in the webhooks table

I've done that against IAD and ORD (AFAIR). So, I think we are good in that sense.

PR #1072, opened 2015-02-11
Integration attempt scale up

sam-falvo:
Branch name is a bit of a misnomer; this is library support to enable scale-up integration testing via Trial.

### DIFF
--- a/otter/test/integration/lib.py
+++ b/otter/test/integration/lib.py
@@ -184,6 +184,15 @@ class ScalingGroup(object):
     Attribute('scaling_group', instance_of=ScalingGroup),
 ])
 class ScalingPolicy(object):
+    """ScalingPolicy class instances represent individual policies which your
+    integration tests can execute at their convenience.
+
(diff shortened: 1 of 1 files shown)


otter/test/integration/lib.py
```diff
@@ -177,6 +179,127 @@ def record_results(resp):
...
+    :param ScalingGroup scaling_group: The scaling group to which this policy
+        applies.
+    """
+
+    def __init__(self):
+        self.policy = [{
+            "name": "integration-test-policy",
+            "cooldown": 0,
+            "type": "webhook",
+            "change": self.scale_by
```
lvh: This doesn't appear to cover scaling by percentages, or absolute capacities. Is that intentional?

lvh:
LGTM modulo comment.

otter/test/integration/lib.py
sam-falvo: Yes.  YAGNI.  When integration tests covering those features are written, we can add them in at that time.

PR #1078, opened 2015-02-12
honor the stored desiredCapacity in group state for convergence-enabled tenants

### LINKED ISSUE #1076: getting group state on a convergence-enabled tenant should return "desiredCapacity" based on stored value
right now desiredCapacity is always returned as active + pending, but active + pending will become basically just a cache of our most recent data-gathering, and thus may not be accurate. So, for convergence-enabled tenants, we should return the stored number.

radix:
fixes #1076.

### DIFF
--- a/otter/rest/groups.py
+++ b/otter/rest/groups.py
@@ -9,6 +9,7 @@ from functools import partial
 from twisted.internet import defer
 
 from otter import controller
+from otter.convergence.composition import tenant_is_enabled
 from otter.json_schema.group_schemas import (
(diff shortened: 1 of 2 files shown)


cyli:
This may be overkill, but would it be ok to add one more assertion to that test that a non-convergence tenant, given otter running on the same config, gets a different response?

radix:
Good idea @cyli, I'll do that!

cyli:
LGTM!  Thanks!  :D

Issue #1049, opened 2015-02-10
Do not bother passing around dicts for desired load balancers


cyli:
From working on #1042, it looks like maybe the desired load balancers being keyed by the `lb_id` in our models doesn't do anything for us? Just a collection of descriptions seems sufficient, and may make more sense when we include RCv3 stuff.

Not super important - we can just grab the values() out for now, but would be neater just to have sets.

radix:
+1 :)

PR #1079, opened 2015-02-12
Save `active` servers to group state during convergence

### LINKED ISSUE #985: Getting group status should reflect changes in Convergence for whitelisted tenants
There are a couple of options here.

1. Getting group status, for convergence-enabled customers, shouldn't show 'active' or 'pending', just 'desired', since we don't keep track of that information any more.
2. Getting group status for convergence-enabled customers should make a request to nova to determine the active/pending servers. This would maintain backwards compatibility but could be a major stress to Nova since we have to scan the entire server list every time a group is fetched.

radix:
This fixes #985 and #1059.

This changes the convergence service so it updates the `active` member of GroupState (in cassandra) when performing convergence. Some notes:

- `pending` doesn't need updated: see my recently merged #1078.
- I added a `convergence=yes` tag to an additional integration test which should pass with this change.
- I rewrote/moved execute_convergence. Since it ended up with group-statey knowledge, it seemed more appropriate to put it into service.py.


otter/convergence/service.py
```diff
@@ -1,17 +1,88 @@
...
+    """
+    return {'id': server.id}
+
+
+def determine_active(servers, lb_nodes):
+    """
+    Given the current NovaServers and LB nodes, determine which servers are
+    completely built.
+
+    :param desired_lbs: As per DesiredGroupState.desired_lbs.
```
manishtomar: This parameter is probably not required.

otter/util/fp.py
```diff
@@ -78,3 +79,10 @@ def predicate_any(*preds):
     with or operator
     """
     return lambda *a, **kw: any(p(*a, **kw) for p in preds)
+
+
+def obj_assoc(o, **k):
```
manishtomar: How about `updated_object` as the name?

radix: so, I was going to say that I named it after the function in the proposed `objtoolz` package (to be added to toolz), except I apparently got the name backwards:

https://github.com/pytoolz/toolz/pull/195

they use `assoc_obj` instead of `obj_assoc`. I'd prefer to stick with their names so we can switch to that library once it's included and released.

otter/convergence/service.py
```diff
@@ -29,22 +99,24 @@ def _get_lock(self, group_id):
         lock.acquire = partial(lock.acquire, timeout=120)
         return lock
 
-    def start_convergence(self, log, tenant_id, group_id, desired,
+    def start_convergence(self, log, scaling_group, group_state,
                           launch_config,
                           perform=perform,
                           execute_convergence=execute_convergence):
         """Converge a group to a capacity with a launch config."""
```
manishtomar: If you don't mind, could you add types for params docstring? Especially for `scaling_group` and `group_state` as they can be confusing. I realize they should probably not be there in the first place :/

radix:

otter/convergence/service.py
```diff
@@ -1,17 +1,87 @@
...
+
+    :param servers: sequence of :obj:`NovaServer`.
+    :param lb_nodes: sequence of :obj:`ILBNode`.
+
+    :return: list of servers that are active.
+    """
+
+    def all_met(server, current_lb_nodes):
+        """Determine if a server is in all the LBs it wants to be in."""
+        desired_lbs = set(concat(server.desired_lbs.values()))
```
manishtomar: Non-blocking thought: Do you think it will be better to implement #1049 and change this?

otter/convergence/service.py
```diff
@@ -1,17 +1,95 @@
...
+            if desired.equivalent_definition(node.description)])
+        return desired_lbs == met_desireds
+
+    return [s for s in servers
+            if s.state == ServerState.ACTIVE
+            and all_met(s, [node for node in lb_nodes if node.matches(s)])]
+
+
+def execute_convergence(
+        scaling_group, desired, launch_config, now, log,
```
manishtomar: Non-blocking nitpick: `desired` can be renamed to `desired_capacity` or `capacity` to be consistent with name in `DesiredGroupState` and also documented :) (Its completely fine to leave it as such)

radix: good point, I'll make the change :)

otter/convergence/service.py
```diff
@@ -1,17 +1,95 @@
...
+
+    :param IScalingGroup scaling_group: The scaling group object.
+    :param GroupState group_state: The group state.
+    :param launch_config: An otter launch config.
+    :param now: The current time in seconds.
+    :param log: A bound logger.
+    :param get_all_convergence_data: The :func`get_all_convergence_data` to use
+        for testing.
+
+    :return: An Effect of List of Step Effect Results. (???)
```
manishtomar: How about `Effect [(StepResult, [String])]`? :)

otter/convergence/service.py
radix: Well, when #1049 gets implemented, this should get updated too. I don't want to block this change on that.

otter/convergence/service.py
radix: Oh right, I forgot about that. The problem is that we still don't have the (StepResult, [String]) thing happening. I'll tweak this docstring though.

otter/test/convergence/test_service.py
```diff
@@ -57,11 +73,169 @@ def test_converge_error_log(self):
...
+        self._assert_active(mgs_eff, expected_active)
+        p_effs = resolve_effect(mgs_eff, None)
+        self.assertEqual(p_effs, parallel([]))
+
+    def test_success(self):
+        """
+        Executes optimized steps if state of world does not match desired and
+        returns the result of all the steps.
+        """
+        # The scenario: We have two servers but they're not in the LBs
```
manishtomar: +1 for the comments :)

manishtomar:
LGTM! Excited to see this :)

radix:
Finally!

Issue #844, opened 2014-12-17
Implement function that retries convergence until done


lvh:
#817 gives us a clean function for a single convergence cycle: gather -> plan -> effect. We additionally need something that tries to converge until the plan is empty, or until an unrecoverable error occurs.

This probably needs #818 to be particularly successful.

lvh:
I should clarify: `ConvergeLater` would still be valuable, potentially, as a hint towards how much you should wait before retrying. However, that's very definitely not a blocker; for now we can just retry with fairly lengthy times in between.

If the times are sufficiently lengthy, this might provide an answer to the conundrum we had on #866 with respect to how to implement request-level throttling.

lvh:
The effect-wrapped bool returned by `execute_convergence` specifies if you should retry or not. That bool really just checks if the list of effects that were parallel-executed is empty or not, so currently the semantics are "empty plan means you're done". So, it looks like we need to implement `ConvergeLater` no matter what.

manishtomar:
+1 on `ConvergeLater`

lvh:
Cc @radix.

One of the arguments to `execute_convergence` is the desired state of the world. When repeatedly executing convergence, we should try to get a new version of the desired state of the world in between `execute_convergence` runs. After all, if a customer is at 10 machines, asks for 100, we bring them up to 20 and then they ask for 50 instead, we shouldn't gradually take them to 100; we should stop when we hit 50.

This means that the real thing that we're retrying is a combination of:

1. Get the new desired state.
2. Execute convergence to attempt to reach that state.

Since our fancy retry logic takes effects, that means that there should be an effect for doing both of these steps that take the same argument as the relevant functions. So:

1. An intent `GetDesiredState`
2. An intent `ExecuteConvergence`

How do I combine these intents into a single effect?

radix:
Ah, I see.

That would just be: `Effect(Func(get_desired_state)).on(execute_convergence)`.

For the record, you don't actually need to make these things specific intents. As my code example shows, your "get desired state" step can just be a Func intent calling get_desired_state. And ExecuteConvergence is just what our execute_convergence function already returns. Not that making them intents is bad.

lvh:
And, to be clear, that'd be an effect of bool, right? (Well, effect of `X`, where `execute_convergence` returns an effect of `X`).

radix:
yep. in order to get that to work with the retry machinery, you could add another callback that raises an exception when the result is True.

[...]

PR #1080, opened 2015-02-12
Add a ModifyGroupState intent

radix:
this adds a simple intent for using ScalingGroup.modify_state.

### DIFF
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -11,6 +11,6 @@ class ModifyGroupState(object):
 
 
 @deferred_performer
-def perform_modify_group_state(dispatcher, mgs):
+def perform_modify_group_state(dispatcher, mgs_intent):
     """Perform a :obj:`ModifyGroupState`."""
-    return mgs.scaling_group.modify_state(mgs.modifier)
+    return mgs_intent.scaling_group.modify_state(mgs_intent.modifier)
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -12,5 +12,5 @@ class ModifyGroupState(object):
 
 @deferred_performer
 def perform_modify_group_state(dispatcher, mgs):
-    """Perform an :obj:`UpdateGroupState`."""
+    """Perform a :obj:`ModifyGroupState`."""
     return mgs.scaling_group.modify_state(mgs.modifier)
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -1,0 +1,16 @@
+from characteristic import attributes
+
+from effect.twisted import deferred_performer
+
+
+@attributes(['scaling_group', 'modifier'])
+class ModifyGroupState(object):
+    """
+    An Effect intent which indicates that a group state should be updated.
+    """
+
+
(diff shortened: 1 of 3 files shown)


otter/models/intents.py
```diff
@@ -0,0 +1,16 @@
...
+
+@attributes(['scaling_group', 'modifier'])
+class ModifyGroupState(object):
+    """
+    An Effect intent which indicates that a group state should be updated.
+    """
+
+
+@deferred_performer
+def perform_modify_group_state(dispatcher, mgs):
```
manishtomar: Non-blocking nitpick: Could you rename `mgs` to `mgs_intent`?

otter/models/intents.py
```diff
@@ -0,0 +1,16 @@
+from characteristic import attributes
+
+from effect.twisted import deferred_performer
+
+
+@attributes(['scaling_group', 'modifier'])
+class ModifyGroupState(object):
```
manishtomar: Have you considered a generic `GroupModelRequest` intent as described [here](https://github.com/rackerlabs/otter/issues/844#issuecomment-70881324)? I feel it can be good path to move any code in direction of effect.

radix: On first impression I'm not really keen on intents like that. My concern is that an intent like that would be _too_ flexible and encourage writing over-effectful code. It's on a continuum with very specific stuff like `ModifyGroupState` at one end, and `Func` on the other.

You could consider that intents should just be like effectful functions (or "procedures", in Pascal lingo). We shouldn't be afraid to write intents for the specific side-effecting operations we need to perform -- and on the other side of the coin, it would be kind of strange to have an actual procedure like `group_model_request` which took a method name and args and kwargs which we used any time we wanted to invoke a group operations.

However, I do understand that defining these intents can be tedious (not only needing a class and a function, but also a change to another file, effect_dispatcher.py). But if this is the main concern, instead of creating very general intents like `GroupModelRequest`, we could instead abstract the _definition_ of intents like these so that we can create them more efficiently -- for example, if we end up creating a bunch of intents that map 1:1 with ScalingGroup methods, imagine that we could actually have an `otter/models/intents.py` that looks like this:

```
performers = {}
def make_group_intent(method_name, arguments):
    """Create an intent class and performer, and register the performer in performers"""
    ...

ModifyGroupState = make_group_intent('modify_state', ['modifier'])
ViewGroupConfig = make_group_intent('view_config', [])

dispatcher = TypeDispatcher(performers)
```

Then, in effect_dispatcher.py, we could just compose in this `otter.models.intents.dispatcher` so we wouldn't need to update that file when we add a new intent here.

Anyway, just a thought exercise.

manishtomar: The `GroupModelRequest` is kind of a bridge between specific intents and `Func`. The only difference is instead of storing the function, you store the function name. But I do like the abstraction you gave of creating specific intents. Out of curiosity, if we were implementing model with effect today, how would you define the intent? This comment is non-blocking though :). I am fine with current intent.

manishtomar:
LGTM!

PR #1081, opened 2015-02-12
coveralls support added

manishtomar:
One issue is that unit tests will be run twice; once with `make unit` and another one with `make coverage`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,13 @@ python:
 install:
     - pip install -r dev_requirements.txt
     - pip install -r requirements.txt
+    - pip install coveralls
 
 script:
     - make lint
     - make docbook
     - make unit
+    - make coverage
+
+after_success:
+    - coveralls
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,8 +6,13 @@ python:
 install:
     - pip install -r dev_requirements.txt
     - pip install -r requirements.txt
+    - pip install coveralls
 
 script:
     - make lint
     - make docbook
     - make unit
+    - make coverage
```
cyli: We can probably just run trial with coverage instead of both make unit and make coverage?  And `make coverage` also builds the html, which I think coverage doesn't need and maybe isn't that useful on travis anyway.

Mimic [just runs `coverage run trial` and `coverage report`](https://github.com/rackerlabs/mimic/blob/master/tox.ini#L8) for instance, rather than just run trial for its tests.

manishtomar: Yeah. Let me try to factor the Makefile to do this.

manishtomar:
I screwed up the PR. I'll close this and open another one shortly.

PR #1083, opened 2015-02-12
RCv3 bulk add success semantics

### LINKED ISSUE #1082: Regain error recovery for RCv3 add/delete in worker code
Currently, these things get the raw result (`response, body`). That's fine, except that the success_pred is now `has_code(20x, 409)`, because there's supposed to be a callback that parses 409s in order to recover; at least in convergence.

This means that 409s are now considered success, which is clearly wrong.

Perhaps the resolution is to run the callback, check if the result is `Success, []`, and, if so, return the body; otherwise, raise an exception?

lvh:
This covers the case where a machine has already been added. This should probably also only happen after I resolve #1082 with #1088, otherwise we might break RCv3.

### DIFF
--- a/otter/test/convergence/test_steps.py
+++ b/otter/test/convergence/test_steps.py
@@ -701,8 +701,8 @@ class RCv3CheckBulkDeleteTests(SynchronousTestCase):
 
     def test_node_not_a_member(self):
         """
-        If the node is not a member of the load balancer pool it's being
-        removed from, the response was successful.
+        If the nodes are already not member of the load balancer pools
+        they're being removed from, the response was successful.
         """
         node_id = '825b8c72-9951-4aff-9cd8-fa3ca5551c90'
         lb_id = '2b0e17b6-0429-4056-b86c-e670ad5de853'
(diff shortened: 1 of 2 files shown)


otter/test/convergence/test_steps.py
```diff
@@ -494,6 +516,83 @@ def test_no_such_lb_message(self):
...
+        (partial_check_bulk_add, _), = eff.callbacks
+        self.assertEqual(partial_check_bulk_add.func,
+                         _rcv3_check_bulk_add)
+        expected_pairs = pset([(lb_b_id, node_b_id)])
+        self.assertEqual(partial_check_bulk_add.args, (expected_pairs,))
+        self.assertEqual(partial_check_bulk_add.keywords, None)
+
+    def test_node_already_a_member(self):
+        """
+        If the node was already a member, the request is successful.
```
cyli: Non-blocking nitpick:  to clarify the distinction between this test and the previous test, maybe "if all nodes are already members"?

cyli:
LGTM if travis unwedges itself and the tests pass :)

cyli:
Oops, looks like the worker tests had a test for `has_code(201)`, which probably needs to be updated to `has_code(201, 409)` ?

lvh:

lvh:

lvh:

lvh:
I am ignoring coveralls since it's still wrong; it's complaining about a bunch of code I didn't even touch.

lvh:

radix:

cyli:

cyli:

PR #1084, opened 2015-02-12
coveralls support added

manishtomar:
Calling only `make coverage` instead of `make unit` as that runs the tests. Factored the Makefile to do the same


lvh:
LGTM, although Travis looks like it's choking.

PR #1088, opened 2015-02-12
RCv3 worker error handling

### LINKED ISSUE #1082: Regain error recovery for RCv3 add/delete in worker code

lvh:
This closes #1082.

This takes the bare effect and strips off the 409 success_pred.

I figured you don't actually want to use the 409-recovery callback, because that way you lose the original body, which is important for worker.

This is kind of a hack, but fortunately it's in code that's going away soon. I kinda want to put the ugly in `otter.convergence.steps` since that already has `bare_request` anyway, but I figure it doesn't matter too much.

This requires ~~#1085~~ first.

### DIFF
--- a/otter/worker/_rcv3.py
+++ b/otter/worker/_rcv3.py
@@ -11,6 +11,7 @@ from effect.twisted import perform
 
 from otter.convergence.steps import BulkAddToRCv3, BulkRemoveFromRCv3
 from otter.http import TenantScope
+from otter.util.pure_http import has_code
 
 
 def _generic_rcv3_request(step_class, request_bag, lb_id, server_id):
@@ -27,6 +28,11 @@ def _generic_rcv3_request(step_class, request_bag, lb_id, server_id):
         request has no body.
     """
     effect = step_class(lb_node_pairs=[(lb_id, server_id)])._bare_effect()
+
+    svc_req = effect.intent
+    codes = set(svc_req.success_pred.codes) - set([409])
+    svc_req.succes_pred = has_code(*codes)
+
     # Unfortunate that we have to TenantScope here, but here's where we're
     # performing.
     scoped = Effect(TenantScope(effect, request_bag.tenant_id))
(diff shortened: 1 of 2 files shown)


lvh:

otter/worker/_rcv3.py
```diff
@@ -27,6 +28,11 @@ def _generic_rcv3_request(step_class, request_bag, lb_id, server_id):
         request has no body.
     """
     effect = step_class(lb_node_pairs=[(lb_id, server_id)])._bare_effect()
+
+    svc_req = effect.intent
```
cyli: If we pop this off of both, won't then deleting fail if the node had already been deleted?

lvh: Yep. Alas, 409 is really wide, so allowing 409 means also allowing a bunch of other failures. What do you think the appropriate way forward is?

cyli: I think for deletes only, most of the 409s are ok?  If the node isn't a member, that's fine.  If the load balancer pool doesn't exist anymore, that's fine.  If the load balancer pool isn't active... uh, I guess the RCv3 team said they'd deal with errors like that, right?

lvh: Okiedo, I'll change it to only be for adds.

cyli: Thanks.  :)  Appreciate @kivattik is currently adding a test that would fail if 409's were to prevent deletes from happening (the test is otter should still be able to scale down if the user had already removed the server from the load balancer pool)

cyli:
:|  Travis is wedged.

But LGTM if travis OKs it :)

lvh:
Coveralls is just straight-up wrong about this; try adding a raise there and see what happens.

PR #1091, opened 2015-02-13
running trial without subprocesses

manishtomar:
This fixes coverage and in effect coveralls as it cannot track subprocess without some [changes](http://nedbatchelder.com/code/coverage/subprocess.html)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ flake8diff:
 flake8full:
 	flake8 ${PYDIRS}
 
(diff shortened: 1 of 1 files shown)


radix:
this only needs to change for the `coverage` target, right? can you make it so the normal test runs still use -j but only coverage doesn't?

manishtomar:
Yeah. Can do that. Will do.

lvh:
LGTM assuming Travis is happy

radix:
woo! 98%!

PR #1093, opened 2015-02-14
policy info from webhook hash using the webhook_keys table only

manishtomar:
This needs to be merged after all the webhooks have been migrated. I've merged the script to do that in #1069 but it is yet to be run on all the regions. I've put this up for any early reviews.


cyli:

cyli:
LGTM!

PR #1095, opened 2015-02-16
Single tenant auth

ttata:


stretchservers:

manishtomar:
@ttata Hey thanks for the PR! I've been thinking of doing the same for quite sometime :)

lvh:
We have a Github label to mark PRs that are in progress, so I edited the title and added the label :)

otter/test/test_auth.py
```diff
@@ -827,12 +856,11 @@ def test_composition(self):
...
         self.assertEqual(ra._retry_interval, 5)
 
-        ia = ra._authenticator
-        self.assertIsInstance(ia, ImpersonatingAuthenticator)
-        self.assertEqual(ia._identity_admin_user, 'uname')
-        self.assertEqual(ia._identity_admin_password, 'pwd')
-        self.assertEqual(ia._url, 'htp')
-        self.assertEqual(ia._admin_url, 'ad')
+        st = ra._authenticator
+        self.assertIsInstance(st, SingleTenantAuthenticator)
```
lvh: These appear to be the impersonating authenticator tests, which should remain unchanged?

ttata: This is the composition test which verifies the authenticator is generated correctly. Shouldn't the the single tenant authenticator be the one used?

lvh: Nope; we're not changing the default authenticator. We're giving people who have regular user credentials (as opposed to admin credentials) the option to run otter, too.

There should be a test that checks composition of the single tenant authenticator, yes, but the existing impersonating authenticator tests shouldn't go away, and the single-tenant authenticator shouldn't be the default.

ttata: Gotcha. I will rework this.

radix:
Is the intent to allow running otter in a stand-alone mode for a single tenant? Are you planning using this authenticator in the API server? I guess that could be done with a change to `generate_authenticator` in the same module FYI, by looking for a config option to specify the authenticator to use.

otter/auth.py
```diff
@@ -509,6 +601,17 @@ def generate_authenticator(reactor, config):
     """
     # FIXME: Pick an arbitrary cache ttl value based on absolutely no science.
     cache_ttl = config.get('cache_ttl', 300)
+    if config['strategy'] == 'single_tenant':
```
lvh: For backwards compatibility reasons, it'd be a good idea to change this to `config.get('strategy', 'impersonation')`.

lvh:

otter/auth.py
ttata: thanks!

lvh:

otter/auth.py
```diff
@@ -284,6 +284,42 @@ def __hash__(self):
...
+    def __init__(self, _identity_user, _identity_password, url):
+        self._identity_user = _identity_user
+        self._identity_password = _identity_password
+        self._url = url
+
+    @wait(ignore_kwargs=['log'])
+    def _auth_me(self, tenant_id, log=None):
+        if log:
+            log.msg('Authenticating as new tenant')
+        d = authenticate_single_tenant(self._url,
```
manishtomar: There is an existing function `authenticate_user` in this file that does the same job as `authenticate_single_tenant`. Any reasons for not using that?

otter/auth.py
```diff
@@ -284,6 +284,42 @@ def __hash__(self):
                      self._url, self._admin_url))
 
 
+@implementer(IAuthenticator)
+class SingleTenantAuthenticator(object):
+    """
+    An authentication handler that authenticates using a single tenant id.
+    """
+    def __init__(self, _identity_user, _identity_password, url):
```
manishtomar: Since this class authenticates only one tenant I think it should take that tenant ID as argument here and ensure only that is the only tenant allowed in `authenticate_tenant`

otter/auth.py
lvh: Given @ttata is new to the codebase, I think it's pretty reasonable for him to not be aware of certain functions that exist within it.

otter/auth.py
```diff
@@ -284,6 +284,42 @@ def __hash__(self):
...
+            log.msg('Authenticating as new tenant')
+        d = authenticate_single_tenant(self._url,
+                                       self._identity_user,
+                                       self._identity_password,
+                                       tenant_id,
+                                       log=log)
+        d.addCallback(extract_service_catalog)
+        return d
+
+    def authenticate_tenant(self, tenant_id, log=None):
```
manishtomar: I think you can decorate this function directly with `@wait` instead of `_auth_me` above.

otter/auth.py
lvh: If the user screws this up, wouldn't they just get authentication errors anyway?

(Also, more importantly, if someone has a custom identity endpoint that lets certain creds authenticate for any tenant, this change would prevent them from using that for ostensibly little to no benefit.)

otter/auth.py
ttata: thanks!

otter/auth.py
manishtomar: > If the user screws this up, wouldn't they just get authentication errors anyway?

Not really. Say the credentials are for tenant "t1" and someone calls `authenticate_tenant` on "t2", it will succeed and return token for "t1" which is unacceptable. 

> (Also, more importantly, if someone has a custom identity endpoint that lets certain creds authenticate for any tenant, this change would prevent them from using that for ostensibly little to no benefit.)

True. However, this code is authenticating to a keystone service which I am not sure allows that behavior. Even otherwise, I think another class can be implemented for that behavior.

otter/auth.py
manishtomar: Yes. That is true. I am sorry I should've reworded it better form :)

ttata: Is there an elegant way to include the "tenantId" and "expire-in-seconds" args to the json in the single tenant case, but exclude them in the impersonation case? Are there any ramifications to sending null values in that request?

otter/auth.py
manishtomar: Just to clarify you don't need `_auth_me` above :)

otter/auth.py
manishtomar: Actually, I am sorry didn't notice that auth request is sending `tenant_id` in it. This will ensure that auth on any other tenant fails. Also, the [docs](http://docs.rackspace.com/auth/api/v2.0/auth-client-devguide/content/POST_authenticate_v2.0_tokens_Token_Calls.html) discuss ways where tenant can be useful. So, kindly ignore my concern about passing the `tenant_id` to the constructor :)

lvh: Damnit, I was typing that up :-)

Also, "not actually Keystone" is a reasonable constraint. Up until very recently (and still, at present?) this was the case at a little hosting company called Rackspace ;)

lvh:

lvh:
Awesome! Thanks @ttata!!

Issue #955, opened 2015-01-23
Parse 409 errors from bulk adding to RCv3


cyli:
@kivattik and @sam-falvo discovered that if the server doesn't have the right network (the same one as the load balancer pool), adding to RCv3 will fail.

The problematic server may need to be deleted, if it's a new server.  Again if it's a new server the group needs to go into error state, because the launch config is wrong.  (We should probably try to validate this during launch config submission, but it might be possible for the network to go away in the interim too, so validation may not be much use).

If it's an old server, I guess we do nothing?  :|

lvh:
How does it fail? I'm actually writing a ticket right now suggesting that the return type of  `execute_convergence` is wrong because it doesn't let you specify "nope, this is wedged, stop retrying". If we can specify that from any step, at least we can give up right now.

Perhaps the signature of the resulting effects is just wrong, and it should instead be an `Effect [Effect]`, where the child effects specify what this effect suggests should happen next.

radix:
Effects intrinsically allow specifying what happens next dynamically by allowing callbacks to return more Effects. What were you thinking those Effects in the list would do and what their result would be?

I would imagine the simplest way to implement "wedging" would be to have a special exception type that a Step's Effect could result in, that would set the group state to ERROR (and a reason, of course).

lvh:
@radix Yeah, but if I return effects from a callback, they're going to get executed unconditionally, which is a waste of time if any of the sibling effects of that convergence cycle end up resulting in a hard failure, right?

radix:
Yeah... I guess I don't fully understand your idea. I guess we should discuss it on the "implement erroring of groups in convergence" ticket

lvh:
Does that ticket exist yet, or is that the one I'm writing?

radix:
the one you're writing :)

manishtomar:
> If it's an old server, I guess we do nothing? :|

@cyli This is w.r.t convergence only right? If so, then I am guessing yes - we do nothing. However, we should log this based on outcome of #884.

cyli:
@manishtomar Yes, that's true.  :)  And yes, it should definitely log.

sam-falvo:
I could well be blowing smoke with this message, but I just had an idea.

[...]

PR #1096, opened 2015-02-16
RCv3 bulk add success semantics

lvh:
This adds support for inactive LBs and nonexistent LBs. In both cases, the request is set to failure. (It's possible that we get to retry these failed groups later, but for now I think failure is the most correct thing that describes what happened).

This refs #955, but doesn't resolve it yet. There's still the matter of invalid cloud servers and "unprocessable" cloud servers.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -377,13 +377,20 @@ def _rcv3_check_bulk_add(attempted_pairs, result):
     if response.code == 201:  # All done!
         return StepResult.SUCCESS, []
 
+    failure_reasons = []
     to_retry = pset(attempted_pairs)
     for error in body["errors"]:
         match = _RCV3_NODE_ALREADY_A_MEMBER_PATTERN.match(error)
         if match is not None:
             to_retry -= pset([match.groups()[::-1]])
 
-    if to_retry:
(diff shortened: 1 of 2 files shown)


lvh:

cyli:
Just to make sure I remember, does the failure step means wedge the group and put into error state?

lvh:
Yes, for now.

cyli:
Non-blocking nitpick - do we want to say which server needed which load balancer?  Users might get confused if we eventually log those errors, if it's a load balancer that's no longer on their launch config.

lvh:
Yeah, that's a great idea. Crossed my mind while writing it, no idea why I forgot about it. I've already been building on this branch for something else; since you mentioned it's non-blocking, I'll file a ticket for that and do it next?

cyli:
Sounds good.  This LGTM if the integration tests pass - for some reason it's not running :|

cyli:

lvh:
Jenkins has basically stopped listening to me

cyli:
jenkins doesn't listen to anyone at this point :|

cyli:
oh there it goes

lvh:
(#1098 is that ticket)

Issue #973, opened 2015-01-28
Consistent Nova metadata naming scheme


cyli:
Now that we're adding load balancer configs back into nova metadata, and possibly draining time (because that's per group not per load balancer), might be good to pick a naming scheme for these variables.

`rax:autoscale:lb:{lbid}` has been proposed for the load balancers.  Is that a good namespacing scheme?  So all our keys in the future will be `rax:autoscale:????`?

We can also possibly transition over the scaling group ID metadata if we just support both for long enough to tell everyone to stop using the first.

lvh:
Reach also cares about this. Unfortunately that probably means we should:

1. Write both for a while on new servers.
2. Find existing servers; write both.
3. Upgrade Reach to use the new one.
4. Stop writing the old key style.

Also, nitpick, the inconsistency with a colon or an underscore after the `autoscale` has always bothered me :)

cyli:
We've done items 1-2 above.  I'm not sure about 3+4, because there might be groups with really old servers that we won't update anytime soon.

We'd need to convert every server that we know about on every old group before we can do 3 and 4.

PR #1102, opened 2015-02-17
Nova metadata consistency

### LINKED ISSUE #1048: Change the nova metadata for lbs to include lb type
Currently convergence expects `rax:autoscale:lb:{lbid}`.

Certainly CLBs and RackConnectV3 IDs do not currently overlap, since one is an integer and the other is a UUID, but this seems brittle to future load balancer type support.

Should be changed to `rax:autoscale:lb:{type}:{lbid}`

cyli:
This fixes #1048.

It also addresses part of #973.  (I hope)

So one thing that may not be cool is that the type is already in the metadata key.  Wondering if I should eliminate it from the value too, but then for stuff like RackConnect the value would be empty.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -1,12 +1,7 @@
 """Code related to gathering data to inform convergence."""
-import json
-import re
-from collections import defaultdict
 from urllib import urlencode
 
 from effect import parallel
-
-from pyrsistent import pmap
 
 from toolz.curried import filter, groupby, keyfilter, map
 from toolz.dicttoolz import get_in
@@ -20,7 +15,6 @@ from otter.convergence.model import (
     CLBNodeType,
     NovaServer,
     ServerState)
-from otter.convergence.steps import _UUID4_REGEX
 from otter.http import service_request
 from otter.indexer import atom
 from otter.util.http import append_segments
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -8,7 +8,7 @@ from effect import parallel
 from pyrsistent import pmap
 
 from toolz.curried import filter, groupby, map
-from toolz.dicttoolz import get_in
+from toolz.dicttoolz import get_in, keyfilter
 from toolz.functoolz import compose, identity
 
 from otter.constants import ServiceType
@@ -205,23 +205,28 @@ def _lb_configs_from_metadata(server):
 
     # throw away any value that is malformed
 
-    def parse_config(config, key):
-        if config.get('type') != 'RackConnectV3':
-            lbid = k[len(_key_prefix):]
+    def parse_config(lb_type, lb_id, configs):
+        for config in configs:
             try:
-                desired_lbs[lbid].append(
-                    CLBDescription(lb_id=lbid, port=config['port']))
+                yield CLBDescription(lb_id=lbid, port=config['port'])
             except (KeyError, TypeError):
                 pass
 
-    for k in server.get('metadata', {}):
-        if k.startswith(_key_prefix):
-            try:
-                configs = json.loads(server['metadata'][k])
-                for config in configs:
-                    parse_config(config, k)
-            except (ValueError, AttributeError):
-                pass
+    lb_metadata = keyfilter(lambda key: key.startswith(_key_prefix),
+                            server.get('metadata', {}))
+
+    for k, v in lb_metadata.iteritems():
+        lb_type, lbid = k[len(_key_prefix):].split(':')
+        if lb_type == 'RackConnectV3':
+            continue
+
+        try:
+            configs = json.loads(v)
+        except (ValueError, AttributeError):
+            pass
+        else:
+            if isinstance(configs, list):
+                desired_lbs[lbid] = list(parse_config(lb_type, lbid, configs))
 
     return pmap(desired_lbs)
 
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -196,40 +196,6 @@ def _servicenet_address(server):
                  if ip.startswith("10.")), "")
 
 
-_key_prefix = "a"
-
-_LB_KEY_PATTERN = re.compile(
-    "rax:autoscale:lb:(CloudLoadBalancer:\d+|RackConnectV3:{uuid})"
-    .format(uuid=_UUID4_REGEX), re.IGNORECASE)
-
-
-def _lb_configs_from_metadata(server):
-    """
-    Construct a mapping of load balancer ID to :class:`ILBDescription` based
-    on the server metadata.
-    """
-    desired_lbs = defaultdict(list)
-
-    for k, v in server.get('metadata', {}).iteritems():
-        m = _LB_KEY_PATTERN.match(k)
-        if m is None:
-            continue
-
-        lb_type, lbid = m.groups()[0].split(':', 1)
-        if lb_type.lower() == 'cloudloadbalancer':
-            # if malformed, skiped the whole key
-            try:
-                configs = json.loads(v)
-                if isinstance(configs, list):
-                    desired_lbs[lbid] = [
-                        CLBDescription(lb_id=lbid, port=c['port'])
-                        for c in configs]
-            except (ValueError, KeyError, TypeError):
-                pass
-
-    return pmap(desired_lbs)
-
-
 def to_nova_server(server_json):
     """
     Convert from JSON format to :obj:`NovaServer` instance.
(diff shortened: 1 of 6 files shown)


otter/convergence/gathering.py
```diff
@@ -205,23 +205,28 @@ def _lb_configs_from_metadata(server):
 
     # throw away any value that is malformed
 
-    def parse_config(config, key):
-        if config.get('type') != 'RackConnectV3':
-            lbid = k[len(_key_prefix):]
+    def parse_config(lb_type, lb_id, configs):
```
lvh: I must admit I was a little confused reading this code, and I'm not sure how to improve it. lb_type is unused, but that's only because RCv3 LBs are manually thrown out further down, right? Are we going to eventually do this for RCv3 too?

otter/convergence/gathering.py
```diff
@@ -205,23 +205,28 @@ def _lb_configs_from_metadata(server):
...
-                configs = json.loads(server['metadata'][k])
-                for config in configs:
-                    parse_config(config, k)
-            except (ValueError, AttributeError):
-                pass
+    lb_metadata = keyfilter(lambda key: key.startswith(_key_prefix),
+                            server.get('metadata', {}))
+
+    for k, v in lb_metadata.iteritems():
+        lb_type, lbid = k[len(_key_prefix):].split(':')
```
lvh: FULL RECOGNITION BEFORE PROCESSING

(nonblocking nitpick, but yeah I do think we can do better than this for recognizing metadata keys, specifically I think this is a good example of a case where a regex would do a better job)

lvh: For those of you who don't recognize it: http://www.cs.dartmouth.edu/~sergey/langsec/occupy/

otter/convergence/gathering.py
cyli: Yes, it's because they're thrown out.  I was going to either create CLBDescription or RCv3Description here, but we don't actually have the RCv3Descriptions or nodes yet as models, so nothing can be done with the RCv3 data yet.

otter/convergence/gathering.py
cyli: Good point, changing

otter/convergence/gathering.py
cyli: If you prefer, we can block this PR until that's in place.

lvh: No, that's fine, it's non-blocking; we already have an RCv3 convergence ticket. However, we should link this conversation there; I'll go do that.

lvh:
LGTM, but over time we probably wanna refactor the metadata parsing thing into something a bit more polished, with a bunch of named capture groups etc :-)

cyli:
@lvh I was wondering if maybe generating/parsing metadata stuff should be on NovaServer, since it should know what kind of metadata it should make and expect?  So it isn't split between composition and gathering, both of which need to know about the exact keys.

lvh:
Yeah, that sounds great! Furthermore, the actual metadata parsing bit should be separated out, since this can be shared across Rackspace services (we agree on the metadata pattern now).

lvh:
(That separation doesn't have to happen from the start, but it probably should be a design concern for whoever writes it.)

cyli:
@lvh I'm not sure how the many-services parsing thing should work, since it seems to just be pulling out `:rax:yourservice`, but I'm happy to do that factoring of our metadata stuff in this PR if you like/agree.

lvh:
It would literally just be `rax_metadata_keys(metadata, "autoscale") -> {"lb": [{...}, {...}]}`, assuming we standardize on the indices thing :)

cyli:
@lvh.  Oh.  Yeah that's easy.  I'm going to put this PR into in-progress then.

otter/convergence/gathering.py
```diff
@@ -193,7 +195,11 @@ def _servicenet_address(server):
                  if ip.startswith("10.")), "")
 
 
-_key_prefix = 'rax:autoscale:lb:'
+_key_prefix = "a"
```
radix: Looks like this can be deleted?

radix:
I think Reach relies on the name of some of these keys, like the group ID one maybe? I like the idea of standardizing but that's kind of a pain. Maybe we could just create a wiki page describing  changes and then send it off to Reach when we're closer to deploying convergence?

lvh:
I've talked to Reach about standardizing within Rackspace on the `rax:` keyspace, and they're on board with that. I don't think they care about the LB metadata keys, but I could be mistaken.

cyli:
@radix @lvh: I remember an email @radix sent confirming that they did not care about the lb keys...

radix:
yeah, the only one I notice that may be a problem is the change to the group ID key.

cyli:
@radix: we're not removing it yet, going to be writing both for a while.  And then probably reading both after we stop writing.

cyli:
As a note to the reviewer, I plan on doing https://github.com/rackerlabs/otter/issues/1049 next since the dicts are becoming more pointless.

cyli:
Ok, complete redo.  :|  Apologies @lvh.

cyli:
At this point, wondering if NovaServer should just have a `from_server_details` classmethod constructor that is basically `gathering.to_nova_server`.  Since the model is literally "serialized server info from nova".

radix:
+1 cyli. I'm also tempted to say that the attributes we expose should also just be properties (or even explicit methods) that look up the data in the stashed json structure.

cyli:
@radix Created an issue to discuss how to do that, but this PR seems big enough for now.  Would it be ok to just review as is?

otter/convergence/model.py
```diff
@@ -126,6 +154,69 @@ def __init__(self):
...
+        Generate autoscale-specific Nova server metadata given the group ID and
+        an iterable of :class:`ILBDescription` providers.
+
+        NOTE: Currently this ignores RCv3 settings and draining timeout
+        settings, since they haven't been implemented yet.
+
+        :return: a metadata `dict` containing the group ID and LB information
+        """
+        metadata = {
+            'rax:auto_scaling_group_id': group_id,
```
radix: I'm not really sure why we want to write both these keys. If we're writing the new ID, then all of our services should know how to look for the new keys. Is it just for backwards compatibility with Reach? Can't we just get them to start looking for the new keys before we deploy convergence?

otter/convergence/model.py
```diff
@@ -126,6 +154,69 @@ def __init__(self):
         assert self.state in ServerState.iterconstants(), \
             "%r is not a ServerState" % (self.state,)
 
+    @classmethod
+    def group_id_from_metadata(cls, metadata):
```
radix: I kinda feel like this should just be a toplevel function (like get_service_metadata) since it doesn't do anything with NovaServer directly

otter/convergence/model.py
```diff
@@ -126,6 +154,69 @@ def __init__(self):
...
+        The old key was ``rax:auto_scaling_group_id``, but the new key is
+        ``rax:autoscale:group:id``.  Try pulling from the new key first, and
+        if it doesn't exist, pull from the old.
+        """
+        return metadata.get(
+            "rax:autoscale:group:id",
+            metadata.get("rax:auto_scaling_group_id", None))
+
+    @classmethod
+    def lbs_from_metadata(cls, metadata):
```
radix: Likewise here (think it should be a toplevel function)

otter/convergence/model.py
cyli: Yes, it's for backwards compatibility.  Yes, we can only write one in convergence so long as we can get Reach to read both keys for a while (we will have to do the same).

Unless we want to put the effort into also migrating the current worker, and also migrate all existing server keys.

radix:
This looks good, just had a few minor comments (and a question about backwards compatibility)

otter/convergence/model.py
radix: Alright, that's cool. :) Thanks for explaining.

cyli: I don't mind asking them - it's just not something I would have thought of.  :)  So... what do we want to do?  Just both sides read both, or actually do a migration?

otter/convergence/model.py
cyli: Sure - if we implement https://github.com/rackerlabs/otter/issues/1106 though, should it be an instance function?  We can lazy-parse (parse an attribute only when it's needed)

otter/convergence/model.py
cyli: Sure, likewise with https://github.com/rackerlabs/otter/issues/1106 as above though.

otter/convergence/model.py
radix: Oh, yeah, you can leave it here for now.

otter/convergence/model.py
radix: Sorry, I meant that your conclusion the first time was fine. Branch LGTM now, but I'll let you merge.

otter/convergence/model.py
cyli: Ok yeah, since we're undecided on whether NovaServer should have properties or attributes, I'll leave them here for now and then maybe move them.  :|

PR #1103, opened 2015-02-17
Webhook table only

manishtomar:
Option in the script to display webhook from index that are not in keys table. This can be run before migration and after migration to check if it is completed.

### DIFF
--- a/scripts/load_cql.py
+++ b/scripts/load_cql.py
@@ -6,6 +6,7 @@ Loads cql into Cassandra
 from __future__ import print_function
 
 import argparse
+from pprint import pprint
 import sys
 import re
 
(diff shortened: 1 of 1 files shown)


scripts/load_cql.py
```diff
@@ -158,6 +163,17 @@ def setup_connection(reactor, args):
...
 
 
+def webhook_index(reactor, args):
+    """
+    Show webhook indexes that is not there table connection
+    """
+    store = CassScalingGroupCollection(None, None)
+    eff = store.get_webhook_index_only()
+    conn = setup_connection(reactor, args)
+    return perform(get_cql_dispatcher(reactor, conn), eff).addCallback(
```
cyli: Nonblocking - in case there are any errors, so that we can see this before we continue, do we want to do `addBoth(pprint)`

cyli:
LGTM aside from comment above.  :)

scripts/load_cql.py
manishtomar: It it errors, it will be printed on `stderr` as it will be returned to `task.react`.

cyli: Ok, sounds good.

PR #1105, opened 2015-02-18
Support for fixed url in service

manishtomar:
if service config contains a url that is prefixed, otherwise URL is taken from service catalog. Need early review please. @radix ?


radix:
looks good, this is just what I imagined :)

radix:
Looks good! :-)

PR #1111, opened 2015-02-19
Integration test credential check

sam-falvo:
Enhances config_check.py to look for accidental credentials in shell scripts matching "*/config.sh" glob pattern.

### DIFF
--- a/scripts/config_check.py
+++ b/scripts/config_check.py
@@ -11,13 +11,15 @@ changed_files = subprocess.check_output(
     ["git", "diff", "--cached", "--name-only"]).split('\n')
 
 
-def look_for_password(dictionary):
+def look_for_password_in_json(dictionary):
     """
     Check for dangerous plaintext passwords.
 
     Search through a dictionary to see if there is a password value that
     is not "REPLACE_WITH_REAL_PASSWORD" or a username value that is not
     "REPLACE_WITH_REAL_USERNAME".
+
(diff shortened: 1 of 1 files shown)


scripts/config_check.py
```diff
@@ -30,13 +33,55 @@ def look_for_password(dictionary):
...
+    """
+    Check for dangerous plaintext passwords in a text file organized by lines.
+    Essentially performs a grep for USERNAME= and PASSWORD= and makes sure they
+    have well-known standard values REPLACE_WITH_REAL_USERNAME, etc.
+
+    :param list lines: A list of lines in the script file.
+    """
+
+    un_template = 'REPLACE_WITH_REAL_USERNAME'
+    pw_template = 'REPLACE_WITH_REAL_TEMPLATE'
```
cyli: Non-blocking nitpick:  out of curiosity, why "REPLACE_WITH_REAL_TEMPLATE" instead of "REPLACE_WITH_REAL_PASSWORD"?

Also, if it's "REPACE_WITH_REAL_PASSWORD", the same template variables could be used for both this function and `look_for_pasword_in_json`.

cyli:
Nice - prints out the filename and error too!  Tested this locally with `config.sh` in root otter directory and then some subdirectory.  Successfully raises an error.

Other than the nitpick above, LGTM.

lvh:
Thanks for addressing this!

PR #1112, opened 2015-02-19
Factor out the partitioning logic from scheduler into its own service

radix:
This makes the partitioner logic a self-contained service so we can use it elsewhere, notably in the convergence code.

Sadly this is a pretty big change, but it's as small in scope as I could get it. It'll probably be useful to just browse the code in the branch to get an idea of the new SchedulerService code.

### DIFF
--- a/otter/scheduler.py
+++ b/otter/scheduler.py
@@ -7,7 +7,7 @@ in the first place.
 from datetime import datetime
 from functools import partial
 
-from twisted.application.internet import TimerService
+from twisted.application.service import MultiService
 from twisted.internet import defer
 
 from otter.controller import (
@@ -19,81 +19,46 @@ from otter.util.deferredutils import ignore_and_log
 from otter.util.hashkey import generate_transaction_id
 
 
-class SchedulerService(TimerService):
+class SchedulerService(MultiService, object):
     """
     Service to trigger scheduled events
     """
 
-    def __init__(self, batchsize, interval, store, kz_client,
-                 zk_partition_path, time_boundary, buckets, clock=None, threshold=60):
+    def __init__(self, batchsize, store, partitioner_factory, threshold=60):
         """
         Initialize the scheduler service
 
         :param int batchsize: number of events to fetch on each iteration
-        :param int interval: time between each iteration
-        :param kz_client: `TxKazooClient` instance
-        :param buckets: an iterable containing the buckets which contains scheduled events
-        :param zk_partition_path: Partiton path used by kz_client to partition the buckets
-        :param time_boundary: Time to wait for partition to become stable
-        :param clock: An instance of IReactorTime provider that defaults to reactor if not provided
+        :param store: cassandra store
+        :param partitioner_factory: :obj:`Partitioner`
         """
-        TimerService.__init__(self, interval, self.check_events, batchsize)
+        super(SchedulerService, self).__init__()
         self.store = store
-        self.clock = clock
-        self.kz_client = kz_client
-        self.buckets = buckets
-        self.zk_partition_path = zk_partition_path
-        self.time_boundary = time_boundary
-        self.kz_partition = None
         self.threshold = threshold
         self.log = otter_log.bind(system='otter.scheduler')
-
-    def startService(self):
-        """
-        Start this service. This will start buckets partitioning
-        """
-        self.kz_partition = self.kz_client.SetPartitioner(
-            self.zk_partition_path, set=set(self.buckets),
-            time_boundary=self.time_boundary)
-        TimerService.startService(self)
-
-    def stopService(self):
-        """
-        Stop this service. This will release buckets partitions it holds
-        """
-        TimerService.stopService(self)
-        if self.kz_partition.acquired:
-            return self.kz_partition.finish()
+        self.partitioner = partitioner_factory(
+            self.log, partial(self._check_events, batchsize))
+        self.partitioner.setServiceParent(self)
 
     def reset(self, path):
         """
-        Reset the scheduler with new path
+        Reset the scheduler with a new path.
         """
-        if self.zk_partition_path != path:
-            self.zk_partition_path = path
-            self.kz_partition = self.kz_client.SetPartitioner(
-                self.zk_partition_path, set=set(self.buckets),
-                time_boundary=self.time_boundary)
-        else:
-            raise ValueError('same path')
+        return self.partitioner.reset_path(path)
 
     def health_check(self):
         """
-        Checks if scheduler service is healthy by comparing oldest event w.r.t current
-        time. If oldtest event is older than a threshold, then it is not healthy
+        Checks if scheduler service is healthy by comparing oldest event w.r.t
+        current time. If oldtest event is older than a threshold, then it is
+        not healthy
 
-        :return: Deferred that fires with tuple (Bool, `dict` of extra debug info)
+        :return: Deferred that fires with tuple (Bool, `dict` of extra debug
+        info)
         """
         if not self.running:
             return defer.succeed((False, {'reason': 'Not running'}))
 
-        if not self.kz_partition.acquired:
-            # TODO: Until there is check added for not being allocted for long time
-            # it is fine to assume service is not healthy when it is allocating since
-            # allocating should happen only on deploy or network issues
-            return defer.succeed((False, {'reason': 'Not acquired'}))
-
-        def check_older_events(events):
+        def check_older_events(events, info):
             now = datetime.utcnow()
             old_events = []
             for event in events:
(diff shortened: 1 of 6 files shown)


radix:

radix:
Since it's hidden up in a previous revision's overall-failing node, I'll link to the successful run of the integration tests here in a comment: https://as-jenkins.rax.io/job/dev-ci-integration-tests/164/

otter/scheduler.py
```diff
@@ -19,81 +19,47 @@
...
-        :param buckets: an iterable containing the buckets which contains scheduled events
-        :param zk_partition_path: Partiton path used by kz_client to partition the buckets
-        :param time_boundary: Time to wait for partition to become stable
-        :param clock: An instance of IReactorTime provider that defaults to reactor if not provided
+        :param store: cassandra store
+        :param partitioner_factory: Callable of (log, callback) ->
+            :obj:`Partitioner`
         """
-        TimerService.__init__(self, interval, self.check_events, batchsize)
+        super(SchedulerService, self).__init__()
```
lvh: Wait, is this okay? I thought MultiService was old-style.

lvh: (Notably, see how `object` appears *after* `MultiService`, which is the special case in which maybe sometimes it works.)

otter/scheduler.py
```diff
@@ -19,81 +19,47 @@
 from otter.util.hashkey import generate_transaction_id
 
 
-class SchedulerService(TimerService):
+class SchedulerService(MultiService, object):
```
lvh: Why is this `MultiService`? We only appear to be using `setServiceParent`, not `addService`.

lvh:
This mostly LGTM modulo comments. Do you need any help getting the Twisted bugfix out so that we have a working threadpool?

otter/scheduler.py
radix: `child.setServiceParent(parent)` calls `parent.addService(child)`. Circular references! Only MultiServices can be parents (or anything else implementing IServiceCollection).

otter/scheduler.py
radix: I'm not really sure what you meant by that parenthetical -- my response to your first question would be "I inherited from `object` as well, so super() will work here"

radix:
I'm not working on the threadpool changes in Twisted -- @glyph has a branch for it, and I'd be happy to review it when it's ready for review. This PR shouldn't change anything in how we use threadpools so the status quo should be maintained.

otter/util/zkpartitioner.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        super(Partitioner, self).__init__()
+        self.kz_client = kz_client
+        self.partitioner_path = partitioner_path
+        self.buckets = buckets
+        self.log = log
+        self.got_buckets = got_buckets
+        self.time_boundary = time_boundary
+        ts = TimerService(interval, self.check_partition)
+        ts.setServiceParent(self)
```
manishtomar: Instead of inheriting from `MultiService` and doing this, what are you thoughts on inheriting `Service` and keep `TimerService` as an object that gets started and stopped on this service's `startService` and `stopService`? I've done similar thing in [metrics](https://github.com/rackerlabs/otter/blob/master/otter/metrics.py#L317-L353)

radix: I think this is better specifically because it allows us to not care about it in startService and stopService, and instead just let the MultiService machinery ensure it gets started and stopped when appropriate.

otter/scheduler.py
lvh: I thought that didn't *quite* work, as foom elucidates in ["Super considered harmful"](https://fuhm.net/super-harmful/).

otter/scheduler.py
```diff
@@ -19,81 +19,47 @@
...
-        else:
-            raise ValueError('same path')
+        return self.partitioner.reset_path(path)
 
     def health_check(self):
         """
-        Checks if scheduler service is healthy by comparing oldest event w.r.t current
-        time. If oldtest event is older than a threshold, then it is not healthy
+        Checks if scheduler service is healthy by comparing oldest event w.r.t
+        current time. If oldtest event is older than a threshold, then it is
```
lvh: typo: oldtest

manishtomar:
I am also reviewing this. Kindly do not merge for sometime :)

otter/util/zkpartitioner.py
```diff
@@ -0,0 +1,133 @@
...
+    local buckets have been determined.
+    """
+    def __init__(self, kz_client, interval, partitioner_path, buckets,
+                 time_boundary, log, got_buckets,
+                 clock=None):
+        """
+        :param log: a bound log
+        :param kz_client: txKazoo client
+        :param partitioner_path: ZooKeeper path, used for partitioning
+        :param list buckets: buckets to distribute between nodes. Ideally
```
manishtomar: This should be `param set buckets`. Also, I didn't understand why this should be `sequence of str`

otter/util/zkpartitioner.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class Partitioner(MultiService, object):
+    """
+    A Twisted service which uses a Kazoo :obj:`SetPartitioner` to allocate
+    logical ``buckets`` between nodes.
+
+    Multiple servers can instantiate these with the same ``partitioner_path``
+    and ``buckets``, and each server will get a disjoint subset of
+    ``range(buckets)`` allocated to them.
```
manishtomar: `range(buckets)` just happens to be the _set_ that this class partitions but in theory it can partition any set of strings.

otter/util/zkpartitioner.py
```diff
@@ -0,0 +1,133 @@
...
+        """Re-initialize the partitioner to use a new path."""
+        if self.partitioner_path != path:
+            self.partitioner_path = path
+            self.partitioner = self._new_partitioner()
+        else:
+            raise ValueError('same path')
+
+    def check_partition(self):
+        """
+        Step through the SetPartitioner state machine, and once everything is
```
manishtomar: +1. Nice comments :)

otter/util/zkpartitioner.py
```diff
@@ -0,0 +1,133 @@
...
+        # Create the partitioner *before* up-calling, because otherwise the
+        # TimerService may call `check_partition` before self.partitioner is
+        # created.
+        self.partitioner = self._new_partitioner()
+        super(Partitioner, self).startService()
+
+    def stopService(self):
+        """Release the buckets."""
+        if self.partitioner.acquired:
+            return self.partitioner.finish()
```
manishtomar: You might want to call `super(Partitioner, self).stopService()` to ensure `ts` is stopped and also `self.running` is updated.

otter/util/zkpartitioner.py
radix: Thanks for the catch. While developing I had experimented with passing in a number of buckets instead of the actual buckets, but then switched back, and this docstring got lost. thanks :)

otter/scheduler.py
radix: Well, I think this is one of the cases where it works, but I can switch it back to an old-style class.

otter/util/zkpartitioner.py
radix: good catch. I'll also add tests for this

radix:
All this junk with stopService and startService and upcalling is tempting me to decompose even further and separating the MultiService from the Partitioner from the TimerService.

otter/util/zkpartitioner.py
radix: It doesn't actually need to be a set -- SetPartitioner doesn't require the arguments to be a set, but any iterable. I'll update this to make it more general/clear.

glyph:
My branch for it is sort of stalled.  It went through review, and exarkun asked for some documentation and a bunch of other stuff which I was not feeling encouraged to go work on; nobody could deterministically reproduce Otter's threadpool issues, so I wasn't completely sure it would fix the problem, and I was quickly distracted by other things.

I should probably just write the docs and get it merged.

otter/test/util/test_zkpartitioner.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Called once when starting and now again when got bad state
+        self.assertEqual(self.kz_client.SetPartitioner.call_args_list,
+                         [mock.call(self.path,
+                                    set=self.buckets,
+                                    time_boundary=self.time_boundary)] * 2)
+        self.assertEqual(self.partitioner.partitioner, new_kz_partition)
+        self.assertEqual(self.buckets_received, [])
+
+    def test_acquired(self):
```
manishtomar: Do you think one more test `test_acquired_again` that checks if buckets is received again after 10 seconds will help?

radix: done :)

manishtomar:
LGTM once build passes :) 

Will be helpful if you can run all scheduler tests before merging. They are not run as integration tests.

radix:
@manishtomar - I tried running all the scheduler tests, but one test is being flaky. Do you know about it? I ran it three times and got three different results.

First an error on line 105:
```
======================================================================
FAIL: test_system_update_cron_style_scheduler_to_execute_in_the_future (test_repo.autoscale.system.schedule_policies.test_system_update_scheduler.UpdateSchedulerTests)
Create an cron style scheduler policy to execute the next minute,         update the policy to execute every 2 minutes and         verify the server count on the group is as expected when the policy is updated,         wait one minute after the policy update and verify the older policy (every minute)         is not executed.
----------------------------------------------------------------------
  File "/opt/otter/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_update_scheduler.py", line 105, in test_system_update_cron_style_scheduler_to_execute_in_the_future
    (cron_style_policy['change'] + 2))
  File "/opt/otter/autoscale_cloudroast/test_repo/autoscale/fixtures.py", line 265, in verify_group_state
    desired_capacity, group_id))
AssertionError: 5 != 3 : Active + Pending servers (5) != (3) minentities on the group 81657698-5458-454b-9827-fbabda5b8672
```

Then an error on line 101:
```
======================================================================
FAIL: test_system_update_cron_style_scheduler_to_execute_in_the_future (test_repo.autoscale.system.schedule_policies.test_system_update_scheduler.UpdateSchedulerTests)
Create an cron style scheduler policy to execute the next minute,         update the policy to execute every 2 minutes and         verify the server count on the group is as expected when the policy is updated,         wait one minute after the policy update and verify the older policy (every minute)         is not executed.
----------------------------------------------------------------------
  File "/opt/otter/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_update_scheduler.py", line 101, in test_system_update_cron_style_scheduler_to_execute_in_the_future
    cron_style_policy['change'])
  File "/opt/otter/autoscale_cloudroast/test_repo/autoscale/fixtures.py", line 265, in verify_group_state
    desired_capacity, group_id))
AssertionError: 2 != 1 : Active + Pending servers (2) != (1) minentities on the group 9327e906-8618-480c-9265-a30ac10aa021

----------------------------------------------------------------------
```

And on the third time I ran it, it passed.

manishtomar:
@radix These are unfortunately known [issues](https://github.com/rackerlabs/otter/labels/bug). Please feel free to merge this :)

PR #1113, opened 2015-02-19
single tenant auth was not returning token + catalog tuple

ttata:
it was only returning service catalog :-/


stretchservers:

lvh:
Oops, good catch. My kingdom for a type system...

PR #1114, opened 2015-02-19
Finish first integration test

sam-falvo:


otter/integration/lib/autoscale.py
```diff
@@ -72,6 +84,74 @@ def decide(resp):
...
+
+        class Looper(object):
+            """Inner-class to support looping until either of two conditions
+            holds: (1) we see at least `servers_desired` servers in existence,
+            or (2) a timeout is reached.
+            """
+
+            def __init__(self, sg):
+                self.elapsed_time = 0
+                self.sg = sg
```
lvh: What is an "sg"?

otter/integration/lib/autoscale.py
```diff
@@ -72,6 +84,74 @@ def decide(resp):
...
+                else:
+                    raise TimeoutError(
+                        "Spent %ds, polling every %ds, timeout." % (
+                            self.elapsed_time, period
+                        )
+                    )
+
+            def ensure_200_result(self, state_results):
+                if state_results[0] != 200:
+                    raise Exception("Scaling group not found.")
```
lvh: This error message could use improvement. On a variety of errors that aren't about not finding the scaling group, it'd say the scaling group wasn't found, right?

sam-falvo: 200 and 404 are listed as "valid results" for the get_scaling_group_details call.  Any other result will throw an exception.  In this case, I'm also discarding the 404 valid result, since it's not valid for our needs here.

otter/integration/lib/autoscale.py
```diff
@@ -72,6 +84,75 @@ def decide(resp):
...
+            the clock to use for scheduling things.  Defaults to `reactor`
+            if not specified.
+
+        :return: If the operation succeeds, the same instance of TestResources.
+            Otherwise, an exception is raised.
+        """
+
+        clock = clock or reactor
+
+        class Looper(object):
```
manishtomar: You could do 
```python
retry(check_servers, can_retry=transient_errors_except(UnknownScalingGroup), next_interval=repeating_interval(10))
```
where `UnknownScalingGroup` exception is raised on 404 instead of implementing your own `Looper`.

cyli:
Please ignore integration test failure - that is due to a different bug in master.

otter/integration/lib/autoscale.py
sam-falvo: Is this blocking?  I'd prefer to explore this in a separate PR, especially since this PR works and is getting long in the tooth.

cyli: I'd love to see this explored/addressed soonish, since it'd be nice to reuse existing code, and also Looper may run into issues if autoscale does not reply in a timely manner (it counts the delays between tries, so if autoscale just hangs on a reply, the tests will also hang).

But yes we can address this in a later PR.  :)  Especially since the tests are for the wait for N servers function.

cyli:
I've run the test against mimic, (with some adjustment because autoscale is not in the mimic service catalog), and it does work, as do the unit tests on lib.

I'm cool with merging this in as a first successful trial-based integration test spike, but it'd be awesome to do a cleanup/refactor run after this too.  :)

sam-falvo:
The reason 200 and 404 are "successful" values is because 404 is an expected result in one of my tests.  Just not this one.  :)

Issue #1115, opened 2015-02-19
Logging errors


manishtomar:
My current idea is to have class like [`UpstreamError`](https://github.com/rackerlabs/otter/blob/master/otter/util/http.py#L78-L112) that will store `system` as `ServiceType`. This class has `details` property that will return `dict` of internal details and this will be added to the log. 

It will be helpful to decide on this now as it will help in #1130. One option is that instead of returning `Effect (Status, [String])` from `as_effect` it can return `Effect (Status, [Exception])` where exception will be mostly `UpstreamError`. It can be other exceptions also. The advantage is that exception will contain parsed message along with structured info in `.details`. Any other ideas?

radix:
@manishtomar I think we might be overusing `system` -- I'd prefer to keep the service type in a separate dedicated field like `service_type`.

radix:
@manishtomar - I'm a bit confused about what exactly this ticket is about. I see two things:

1. the log system should automatically extract data from `UpstreamError` to include it in log messages in a structured way
2. we should put exceptions in our `reasons` list

The second point is already done-ish (that is, some steps are including exceptions in their reasons list now), but the first isn't. Should this issue be just about the idea of smart logging of exception objects? Or do we have another ticket for that? (The closest I can find is #1107, but that's more general).

manishtomar:
Closing this as we've started putting exceptions in steps result and #1189 is being used to discuss the structure of exceptions to be supported.

Issue #1107, opened 2015-02-18
Structured formal logging


manishtomar:
Each log should have a message type that has corresponding structure associated with it. The structure includes the textual message, related fields along with their type in it. This structure can be used to induce ES mapping. The code logging will give message type and extra fields. The formatter will validate the fields based on type and log with textual message.

radix:
I would like an API that uses a positional argument as the otter_msg_type, for example:

`log.event("the-msg-type", **kwargs)`

perhaps for errors we could do something like...

`log.err_event(Failure_or_None, "the-msg-type")`

manishtomar:
@radix Yes. That is what I had in mind :)

manishtomar:
I am having second thoughts about validating message field types. Since I will be partially doing what a JSON schema validator does, why not use that directly? i.e. describe each log type's fields as a JSON schema and validate it using a schema validator. This does seem like a lot of work for every log message. 

@radix Any thoughts?

radix:
@manishtomar I think json-schema is a decent way to specify our schemas, and is preferable to rolling our own schema system.

Here are my thoughts about our logging situation in general, and it mostly comes down to a priority list:

1. support `log.msg('event-key')` and `log.err(x, 'event-key')` which gets converted to `otter_msg_type` in the output
2. support structured serialization of exception objects
3. implement support for strict schemas, where schemas are looked up by the `otter_msg_type`
4. define our first schemas for the Cloud Feeds log message types, since those have a pretty strict format
5. then define schemas for any messages where we're relying on a particular format for ahead-of-time indexing in ElasticSearch
6. then implement schemas for everything ever

Each of these items should be (at least) a separate PR.

PR #1116, opened 2015-02-19
Specification based logging

manishtomar:
Implements #1107 and part of #1115. This is not completed. I wanted to get the idea across and get it reviewed. Sample logging code: 
```python
log.msg("launch-servers", num_servers=4)
```
This will get validated and translated to `log.msg("Launching {num_servers} servers", num_servers=4, otter_msg_type="launch-servers")`. 

For errors, it will look for `details` property and validate fields in that. The `details` property is expected to be expanded in `otter.log.formatters.ObserverWrapper`. That may not be the best place to expand `details` but I'll think some more about it.

### DIFF
--- a/otter/log/spec.py
+++ b/otter/log/spec.py
@@ -4,6 +4,7 @@ Format logs with specification
 
 from copy import deepcopy
 
+from otter.constants import ServiceType
 from otter.log.formatters import IGNORE_FIELDS
 
 
@@ -28,8 +29,20 @@ fields = {
 }
 
 
+# Fields that will occur in error
+error_fields = {
+    "system": SystemType,
+    "operation": basestring,
+    "url": basestring,
+    "code": int,
+    "message": basestring,
+    "body": basestring,
+    "headers": dict     # need more validation. I miss type system :(
+}
+
+
 # mapping from msg type -> message
-spec = {
+msg_types = {
     "launch-servers": "Launching {num_servers} servers",
     "delete-server": "Deleting {server_id} server",
     "add-server-clb": ("Adding {server_id} with IP address {ip_address} "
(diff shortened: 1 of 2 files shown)


otter/log/spec.py
```diff
@@ -0,0 +1,118 @@
...
+        except ValueError:
+            # TODO: What to do if it is not valid? Should it instead
+            # send fixed error event with event_dict msg in it?
+            pass
+        observer(speced_event)
+
+    return validate_observer
+
+
+def validate_message(event_dict):
```
radix: This does more than validation, so the name should probably change

otter/log/spec.py
```diff
@@ -0,0 +1,118 @@
...
+        validate_field(fields, field, event_dict.get(field, None))
+
+    if event_dict.get('isError', False):
+        validate_error(event_dict)
+
+    # Format the message
+    # REVIEW: Thinking of changing event_dict in place instead of deepcopy
+    # as this code will be called very often?
+    speced_event = deepcopy(event_dict)
+    speced_event["message"] = (msg, )
```
radix: I think we should set speced_event['otter_msg_type'] = msg_type here

radix:
I hadn't thought of using the actual message argument to specify the type, I guess that works pretty well. Thanks for doing this :)

otter/log/spec.py
manishtomar: Yes. Sorry forgot to add that :)

otter/log/spec.py
```diff
@@ -0,0 +1,118 @@
...
+def SpecificationObserverWrapper(observer):
+    """
+    Return observer that validates messages based on specification
+    and delegates to given observer
+    """
+    def validate_observer(event_dict):
+        try:
+            speced_event = validate_message(event_dict)
+        except ValueError:
+            # TODO: What to do if it is not valid? Should it instead
```
cyli: That seems like a workable solution - we can look for these in logs and slowly work to fix them.  :)

otter/log/spec.py
```diff
@@ -0,0 +1,118 @@
...
+fields = {
+    "num_servers": int,
+    "server_id": UUID,
+    "clb_id": int,
+    "ip_address": IPAddress,
+}
+
+
+# Fields that will occur in error
+error_fields = {
```
cyli: Non-blocking comment - these work quite well for API errors.  How are non-API errors handled (e.g. tracebacks we happen to get because there might be a bug in our code, etc.)?

manishtomar: It is desirable to validate those too. I think adding `traceback: basestring` in `fields` should ensure that traceback is string. But `ObserverWrapper` should be wrapped before `SpecificationWrapper`. I have some thoughts to refactor formatters. I'll make a separate PR.

cyli: I was also wondering if it's useful to tag things as API errors vs non-api errors?  We may be able to find bugs more easily in our code by filtering away the API errors?  (e.g. having an exception type or error type name be a field?)

manishtomar: > (e.g. having an exception type or error type name be a field?)

+1

manishtomar:
Temporarily closing this. I plan to do minor refactor of formatters and come back to this after that.

PR #1123, opened 2015-02-21
moved error formatting to separate observer

manishtomar:
And also logging `exception_type` which is the name of the exception class in `Failure`. 

This will help in implementing #1116

### DIFF
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -281,11 +281,6 @@ def ObserverWrapper(observer, hostname, seconds=None):
             "otter_facility": eventDict.get("system", "otter"),
         }
 
-        if "file" in eventDict:
-            log_params["file"] = eventDict["file"]
-        if "line" in eventDict:
-            log_params["line"] = eventDict["line"]
-
         for key, value in eventDict.iteritems():
             if key not in PRIMITIVE_FIELDS:
                 log_params[key] = value
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -248,7 +248,7 @@ def ErrorFormattingWrapper(observer):
             "message": event.get("message", (message, )),
             "level": event.get("level", level)
         })
-        event.pop('isError', None)
+        [event.pop(k, None) for k in ERROR_FIELDS]
 
         observer(event)
 
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -42,7 +42,10 @@ def JSONObserverWrapper(observer, **kwargs):
     :rtype: :class:`ILogObserver`
     """
     def JSONObserver(eventDict):
-        observer({'message': (json.dumps(eventDict, cls=LoggingEncoder, **kwargs),)})
+        if 'message' in eventDict:
+            eventDict['message'] = ''.join(eventDict['message'])
+        observer({'message': (json.dumps(eventDict,
+                                         cls=LoggingEncoder, **kwargs),)})
 
     return JSONObserver
 
@@ -123,8 +126,9 @@ def PEP3101FormattingWrapper(observer):
     return PEP3101FormattingObserver
 
 
-IGNORE_FIELDS = set(["message", "time", "isError", "system", "id", "failure",
-                     "why", "audit_log"])
+ERROR_FIELDS = {"isError", "failure", "why"}
+
+PRIMITIVE_FIELDS = {"time", "system", "id", "audit_log"}
 
 AUDIT_LOG_FIELDS = {
     "audit_log": bool,
@@ -215,6 +219,42 @@ def audit_log_formatter(eventDict, timestamp, hostname):
     return audit_log_params
 
 
+def ErrorFormattingWrapper(observer):
+    """
+    Return log observer that will format error if any and delegate it to
+    given `observer`
+    """
+
+    def error_formatting_observer(event):
+
+        message = ""
+
+        if event.get("isError", False):
+            level = 3
+
+            if 'failure' in event:
+                excp = event['failure'].value
+                message = repr(excp)
+                event['traceback'] = event['failure'].getTraceback()
+                event['exception_type'] = excp.__class__.__name__
+
+            if 'why' in event and event['why']:
+                message = '{0}: {1}'.format(event['why'], message)
+
+        else:
+            level = 6
+
+        event.update({
+            "message": event.get("message", (message, )),
+            "level": event.get("level", level)
+        })
+        event.pop('isError', None)
+
+        observer(event)
+
+    return error_formatting_observer
+
+
 def ObserverWrapper(observer, hostname, seconds=None):
     """
     Create a log observer that will format messages and delegate to
@@ -232,7 +272,6 @@ def ObserverWrapper(observer, hostname, seconds=None):
         seconds = time.time
 
     def Observer(eventDict):
-        message = None
 
         log_params = {
             "@version": 1,
@@ -242,35 +281,14 @@ def ObserverWrapper(observer, hostname, seconds=None):
             "otter_facility": eventDict.get("system", "otter"),
         }
 
-        if eventDict.get("isError", False):
-            level = 3
-
-            if 'failure' in eventDict:
-                message = repr(eventDict['failure'].value)
-                log_params['traceback'] = eventDict['failure'].getTraceback()
-
-            if 'why' in eventDict and eventDict['why']:
-                message = '{0}: {1}'.format(eventDict['why'], message)
-
-        else:
-            level = 6
-
-        if not message:
-            message = eventDict["message"][0] if eventDict["message"] else ""
-
-        log_params.update({
-            "message": message,
-            "level": eventDict.get("level", level),
-        })
-
         if "file" in eventDict:
             log_params["file"] = eventDict["file"]
         if "line" in eventDict:
             log_params["line"] = eventDict["line"]
 
         for key, value in eventDict.iteritems():
-            if key not in IGNORE_FIELDS:
-                log_params["%s" % (key, )] = value
+            if key not in PRIMITIVE_FIELDS:
+                log_params[key] = value
 
         # emit an audit log entry also, if it's an audit log
         if 'audit_log' in eventDict:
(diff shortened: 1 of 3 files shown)


otter/log/formatters.py
```diff
@@ -215,6 +219,42 @@ def audit_log_formatter(eventDict, timestamp, hostname):
...
+                message = '{0}: {1}'.format(event['why'], message)
+
+        else:
+            level = 6
+
+        event.update({
+            "message": event.get("message", (message, )),
+            "level": event.get("level", level)
+        })
+        event.pop('isError', None)
```
cyli: Do we want to remove all the `ERROR_FIELDS`?  Since the useful failure information is being translated here, for instance, `failure` may no longer be necessary.

otter/log/formatters.py
```diff
@@ -215,6 +219,42 @@ def audit_log_formatter(eventDict, timestamp, hostname):
...
+                event['exception_type'] = excp.__class__.__name__
+
+            if 'why' in event and event['why']:
+                message = '{0}: {1}'.format(event['why'], message)
+
+        else:
+            level = 6
+
+        event.update({
+            "message": event.get("message", (message, )),
```
cyli: The previous behavior seemed to be that the newly constructed error message always overrode the eventdict message - this seems like the eventdict's message would take precedence.

I'm not sure if maybe this just never happens - I guess we don't have tests for it.

otter/log/formatters.py
```diff
@@ -242,35 +281,14 @@ def Observer(eventDict):
...
-
-        if not message:
-            message = eventDict["message"][0] if eventDict["message"] else ""
-
-        log_params.update({
-            "message": message,
-            "level": eventDict.get("level", level),
-        })
-
         if "file" in eventDict:
```
cyli: I know this PR hasn't changed this part, but I wonder if these lines are necessary - it seems like the for loop below would handle it?

otter/log/formatters.py
manishtomar: Ah yes we should remove all `ERROR_FIELDS`. I originally created `ERROR_FIELDS` for spec based PR but didn't realize it can be used here. Thank you!

otter/log/formatters.py
manishtomar: From [twisted log](https://github.com/twisted/twisted/blob/trunk/twisted/python/log.py#L123) it looks like it happens if `log.err("string")` is called which (I think) is the reason I did it.

otter/log/formatters.py
manishtomar: Oh yeah. Didn't realize. Thanks for catching this.

otter/log/formatters.py
cyli: Oh ok.  I'm not sure how to handle that.  :|  But it seems like otherwise the message gets emptied out:

Current traceback messages look like:  

```
2015-02-21_01:29:39.60887 {
2015-02-21_01:29:39.60888   "@timestamp": "2015-02-21T01:29:39.608487", 
2015-02-21_01:29:39.60889   "@version": 1, 
2015-02-21_01:29:39.60889   "host": "vm", 
2015-02-21_01:29:39.60889   "level": 3, 
2015-02-21_01:29:39.60890   "message": "Error when performing convergence: TypeError(\"'NoneType' object has no attribute '__getitem__'\",)", 
2015-02-21_01:29:39.60891   "otter_facility": "otter.rest.groups.delete_scaling_group", 
2015-02-21_01:29:39.60891   "otter_msg_type": "convergence-perform-error", 
2015-02-21_01:29:39.60891   "scaling_group_id": "7ec2111c-2145-46e8-a307-1fbaae429ddd", 
2015-02-21_01:29:39.60892   "tenant_id": "000000", 
2015-02-21_01:29:39.60892   "traceback": <snip>, 
2015-02-21_01:29:39.60897   "transaction_id": "bb4e8db6-f9d4-40d4-b988-5aa8510e0427"
2015-02-21_01:29:39.60897 }
```

And on this branch it looks like:

```
2015-02-21_01:22:40.17532 {
2015-02-21_01:22:40.17533   "@timestamp": "2015-02-21T01:22:40.175153", 
2015-02-21_01:22:40.17533   "@version": 1, 
2015-02-21_01:22:40.17533   "exception_type": "TypeError", 
2015-02-21_01:22:40.17534   "failure": "[Failure instance: <snip>\n]", 
2015-02-21_01:22:40.17534   "host": "vm", 
2015-02-21_01:22:40.17534   "level": 3, 
2015-02-21_01:22:40.17535   "message": "", 
2015-02-21_01:22:40.17535   "otter_facility": "otter.rest.groups.delete_scaling_group", 
2015-02-21_01:22:40.17535   "otter_msg_type": "convergence-perform-error", 
2015-02-21_01:22:40.17536   "scaling_group_id": "95d9a071-f989-4a71-889e-1e3f74ac9878", 
2015-02-21_01:22:40.17537   "tenant_id": "000000", 
2015-02-21_01:22:40.17537   "traceback": <snip>, 
2015-02-21_01:22:40.17538   "transaction_id": "5f7bfc8e-c8a2-4e01-98a4-2d69008288b7", 
2015-02-21_01:22:40.17539   "why": "Error when performing convergence"
2015-02-21_01:22:40.17539 }

Comment (manishtomar) [inline otter/log/formatters.py]: Oh ok. Thanks a lot for checking! Will change it. But would like to know why it happened.

**Comment by manishtomar** [comment]
@cyli Fixed it. It now looks like this:
```
{
  "@timestamp": "2015-02-23T04:41:34.856264",
  "@version": 1,
  "code": 500,
  "exception_type": "ConnectionRefusedError",
  "host": "vm",
  "level": 3,
  "message": "Request failed: Unhandled Error: ConnectionRefusedError('Connection refused',)",
  "otter_facility": "otter.rest.groups.list_all_scaling_groups",
  "request_status": "failed",
  "tenant_id": "394834",
  "traceback": "Traceback (most recent call last):\nFailure: twisted.internet.error.ConnectionRefusedError: Connection was
 refused by other side: 111: Connection refused.\n",
  "transaction_id": "7d23162d-0906-495e-a79e-ebd246eaf21a",
  "uri": "/v1.0/394834/groups"
}
```

**Comment by cyli** [comment]
This generally LGTM :)  Would it be ok to add one more test that `log.err('this is an error message')` produces `event['message'] = 'this is an error message'`?  And to possibly document in the docstring for the error formatter how we want errors formatted?

Just so the next time we don't have to figure this out again :D

Apologies for the trouble!

**Comment by radix** [comment]
@cyli getting `log.err('foo')` setting `message` to `'foo'` will be difficult. 

(sorry, this message may be overkill, it's quite possible that I misunderstood your intent. your comment just sparked some interest and I went to research some stuff about twisted.python.log's behavior....)

```
>>> def emit(e): print e
...
>>> addObserver(emit)
>>> err('foo')
{'why': None, 'time': 1424721390.463469, 'message': ("'foo'",), 'isError': 1, 'system': '-'}
```

passing a string as the first argument to `err` is kind of janky, as you can see (it basically wants a failure or exception object, and falls back to just calling `repr()` on the argument). string messages are meant to be passed as the second argument to `err` (and I really mean "second", not passed as a keyword argument, since the parameter name is `_why`).

Basically, `err` should only be considered for cases where you want to log an _exception_, not just any "error message". Perhaps a good way to represent errors without exceptions would be: `log.msg('foo', isError=True)`. This'll set the `level` parameter in our emitted json.

**Comment by cyli** [comment]
@radix:  Apologies, I thought that you and @manishtomar had agreed to do `log.err('foo')`, which is what I thought the change [here](https://github.com/rackerlabs/otter/pull/1123#discussion_r25116976) was intended to reflect (the original message going to eventDict's 'message')

`log.msg('foo', isError=True)` would go through the same code-path though (that is, the original message is used).

So yes, if that is the recommended way, @manishtomar please add a test case for that instead of what I said and maybe in the docstring document that particular use?  :)

Thanks @radix

File: otter/test/log/test_log.py [hunk 4]
```diff
@@ -158,6 +162,16 @@ def test_failure_logging(self):
         self.observer.assert_called_once_with(
             {'message': (SameJSON({'foo': str(failure)}),)})
 
+    def test_message_is_concatenated(self):
+        """
+        message tuple in event is concatenated before passing on
+        """
+        eventDict = {'message': ('mine',)}
```
radix: This doesn't really test if the message tuple is concatenated -- maybe you should add a second element here

radix:
so @cyli helped me finally understand what's going on here.

To give a little backstory: unfortunately, the way t.p.log deals with messages is _really_ terrible, because it's inconsistent. The whole `message` vs `why` thing is partly bad design and partly a backwards compatibility hack from ancient times (IIRC).

In order to still allow us to use the standard `msg` and `err` functions to easily specify the `otter_msg_type`, we'll use the following syntaxes:

    log.msg("event-type", **kwargs)
    log.err(Failure, "event-type", **kwargs)

In the `msg` case, the "event-type" string becomes the `message` entry in the dict. In the `err` case, it becomes the `why` entry int he dict. There's *no way* to pass a `message` key when using `log.err`:

```
>>> err(Exception("ugh"), 'because this', message='extra message')
{'failure': <twisted.python.failure.Failure <type 'exceptions.Exception'>>, 'why': 'because this', 'time': 1424724320.036373, 'message': (), 'isError': 1, 'system': '-'}
```
Notice that it simply ignores the `message` we passed and replaces it with `()`. Gross.

So yeah. Because of this, it's very unlikely that we'd see both a `why` and a `message` in the same dict. A user would have to manually pass those in with a `log.msg("message", isError=True, why="why")` in order to run into this message-vs-why thing.

Because of this, I don't think it's a big deal that we allow one of those keys to take precedence over the other, regardless of which one does. But as @cyli points out, whatever precedence we do choose should have a unit test.

I think it would be _slightly_ better if we didn't throw away either of them, in case someone does construct such a `log.msg` call, but I don't think it should block this PR.

otter/test/log/test_log.py
```diff
@@ -382,11 +403,73 @@ def test_isError_includes_why_in_short_message(self):
...
+        Empty message in event is overwritten with failure message
+        """
+        self.wrapper({'message': (), 'isError': True,
+                      'failure': Failure(ValueError())})
+        self.observer.assert_called_once_with(
+            matches(ContainsDict({'message': Equals(('ValueError()',))})))
+
+    def test_message_why_iserror(self):
+        """
+        When message, why and isError is given, then message is updated
```
cyli: Doc nitpick - The message in this case is not overwritten - why is thrown out.

cyli:
Aside from the nitpick above, LGTM!  Apologies for the delay.

PR #1125, opened 2015-02-23
Change delete group to submit a for real launch config and not None

### LINKED ISSUE #1120: Convergence failure when force-deleting a scaling group because launch config is None
When force-deleting a scaling group, the config is updated so that the min/max is 0.  Then `obey_config_change` is called with a null launch config, because hey the launch config is not needed to delete a group.

However, this breaks the convergence backend functionality, which extracts the load balancers from what it expects to be a real launch config.

We can probably grab the manifest instead of just the config when force-deleting, so that we can grab both the config and the launch config.

```
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/effect/_continuation.py", line 25, in bounce
    trampoline(func, *args, **kwargs)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/effect/_continuation.py", line 56, in trampoline
    f(bouncer, *args, **kwargs)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/effect/_base.py", line 134, in _run_callbacks
    bouncer.bounce(_run_callbacks, chain, result)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/effect/_base.py", line 83, in guard
    return (True, sys.exc_info())
--- <exception caught here> ---
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/effect/_base.py", line 81, in guard
    return (False, f(*args, **kwargs))
  File "/opt/otter/otter/convergence/service.py", line 79, in got_all_data
    scaling_group.uuid, launch_config, desired_capacity)
  File "/opt/otter/otter/convergence/composition.py", line 56, in get_desired_group_state
    lbs = freeze(launch_config['args'].get('loadBalancers', []))
exceptions.TypeError: 'NoneType' object has no attribute '__getitem__'
```

cyli:
This involved changing some parts of the cass model around.  :|  I just modified view_manifest to optionally not get the state, since we just want both the config and the launch config.

Kinda ugly, and the whole rest/cass model could use a big cleanup after convergence, but this was the smallest change I could think of.  I had originally intended changing `modify_state` to use this function too, so every time the state is modified the state, the launch config, and the config will be passed as one (hence the branch name), but that change turned out to be much bigger because of all the tests.

Still 2 failing functional tests, but the tracebacks are gone (fixes #1120)

### DIFF
--- a/otter/models/interface.py
+++ b/otter/models/interface.py
@@ -263,12 +263,14 @@ class IScalingGroup(Interface):
     uuid = Attribute("UUID of the scaling group - immutable.")
     tenant_id = Attribute("Rackspace Tenant ID of the owner of this group.")
 
-    def view_manifest(with_webhooks=False):
+    def view_manifest(with_policies=True, with_webhooks=False):
         """
         The manifest contains everything required to configure this scaling:
         the config, the launch config, and all the scaling policies.
 
-        :param bool with_webhooks: Should webhooks information be included?
+        :param bool with_policies: Should policies information be included?
+        :param bool with_webhooks: If policies are included, should webhooks
+            information be included?
 
         :return: a dictionary corresponding to the JSON schema at
             :data:`otter.json_schema.model_schemas.manifest`
--- a/otter/models/interface.py
+++ b/otter/models/interface.py
@@ -250,10 +250,14 @@ class ScalingGroupStatus(Names):
     """
     Status of scaling group
     """
-    ACTIVE = NamedConstant()    # Group is active and executing policies/converging
-    ERROR = NamedConstant()     # Group has errored due to (mostly) invalid
-                                # launch configuration and has stopped executing
-                                # policies/converging
+    ACTIVE = NamedConstant()
+    "Group is active and executing policies/converging"
+
+    ERROR = NamedConstant()
+    """
+    Group has errored due to (mostly) invalid launch configuration and has
+    stopped executing policies/converging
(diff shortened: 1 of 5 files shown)


cyli:

cyli:
Is this a real pull request?  why is travis and jenkins not picking up...

cyli:
"I wish I were a real pull request!"
"If you are good and brave, keep and keep working hard, someday Travis and Jenkins may run your tests!"

cyli:

radix:
I need a "Like" button for GitHub comments.

otter/models/interface.py
```diff
@@ -263,12 +267,14 @@ class IScalingGroup(Interface):
...
-    def view_manifest(with_webhooks=False):
+    def view_manifest(with_policies=True, with_webhooks=False):
         """
         The manifest contains everything required to configure this scaling:
         the config, the launch config, and all the scaling policies.
 
-        :param bool with_webhooks: Should webhooks information be included?
+        :param bool with_policies: Should policies information be included?
+        :param bool with_webhooks: If policies are included, should webhooks
+            information be included?
```
manishtomar: Non-blocking: Maybe it should raise `ValueError` if webhooks is asked without policies?

manishtomar:
LGTM :) 

All the `test_force_delete_group` tests pass against mimic for convergence tenant.

PR #1127, opened 2015-02-23
Return None in controller.converge only if delta is zero.

### LINKED ISSUE #1126: Executing a policy in convergence always results in a 403 - Cannot Execute Policy because no change in servers
This is because we do a janky thing [in the controller where if None is returned by `converge` it means no change](https://github.com/rackerlabs/otter/blob/master/otter/controller.py#L233).  However, if `defer.succeed(None)` is returned it's all good.  

If the tenant is convergence-enabled [`None` is always returned](https://github.com/rackerlabs/otter/blob/master/otter/controller.py#L150).

So change it so that None is only returned if delta is 0, and `defer.succeed(None)` is returned otherwise.

cyli:
Fixes #1126

I am not doing audit logging in this case because there is already some existing convergence logging story.  Not sure if this is a good idea.

This should fix the remaining functional test failures.

### DIFF
--- a/otter/test/test_controller.py
+++ b/otter/test/test_controller.py
@@ -1105,7 +1105,6 @@ class ConvergeTestCase(SynchronousTestCase):
         result = controller.converge(log, 'txn-id', group_config, self.group,
                                      state, 'launch', policy,
                                      config_value=config_data.get)
-        self.assertIsInstance(result, defer.Deferred)
         self.assertIs(self.successResultOf(result), None)
         self.converger_mock.start_convergence.assert_called_once_with(
             log, self.group, state, 'launch')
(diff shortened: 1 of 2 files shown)


cyli:

cyli:

otter/test/test_controller.py
```diff
@@ -1104,6 +1105,30 @@ def test_real_convergence(self):
         result = controller.converge(log, 'txn-id', group_config, self.group,
                                      state, 'launch', policy,
                                      config_value=config_data.get)
+        self.assertIsInstance(result, defer.Deferred)
```
radix: IsInstance is probably overkill here, I think just giving it to successResultOf is good enough

radix:
LGTM :) merge after build!

PR #1130, opened 2015-02-25
Return create server failures

### LINKED ISSUE #1043: CreateServer should fail and specify the reason when the flavor is wrong
Refs #956, #1041.

### LINKED ISSUE #1044: CreateServer should fail and specify the reason when the image is wrong
Refs #956, #1041.

### LINKED ISSUE #1045: CreateServer should fail and specify the reason when the tenant is disabled or hit quota
Tenants can hit quotas, be disabled (fraudulent?)...

Refs #956, #1041.

cyli:
Fixes #1043 
Fixes #1044 
Fixes #1045 

We should probably log unrecognized failures too.  I'm not sure what to do with that.  But these are the user-visible errors that can be surfaced.

Right now it just uses the Nova info, but we can always provide more info too (for instance, the Nova flavor too small message isn't super helpful).

It'd be nice to do so in a structured format though, instead of a just a string.  So just leaving this as is for now.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -77,10 +77,6 @@ _NOVA_403_PUBLIC_SERVICENET_BOTH_REQUIRED = _forbidden_plaintext(
 _NOVA_403_RACKCONNECT_NETWORK_REQUIRED = _forbidden_plaintext(
     "Exactly 1 isolated network\(s\) must be attached")
 
-_NOVA_403_QUOTA_REACHED = re.compile(
-    "^Quota exceeded for (?P<limit>\S+): Requested \d+, but already used \d+ "
-    "of \d+ (?P=limit)$")
-
 
 def _parse_nova_user_error(api_error):
     """
@@ -102,7 +98,7 @@ def _parse_nova_user_error(api_error):
     elif api_error.code == 403:
         message = _try_json_message(api_error.body,
                                     ("forbidden", "message"))
-        if message and _NOVA_403_QUOTA_REACHED.match(message):
+        if message:
             return message
 
         for pat in (_NOVA_403_RACKCONNECT_NETWORK_REQUIRED,
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -82,6 +82,37 @@ _NOVA_403_QUOTA_REACHED = re.compile(
     "of \d+ (?P=limit)$")
 
 
+def _parse_nova_create_error(api_error):
+    """
+    Parse API errors for user failures on creating a server.
+
+    :param api_error: The error returned from Nova
+    :type api_error: :class:`APIError`
+
+    :return: the nova message as to why the creation failed, if it was a user
+        failure.  None otherwise.
+    :rtype: `str` or `None`
+    """
+    if api_error.code == 400:
+        message = _try_json_message(api_error.body,
+                                    ("badRequest", "message"))
+        if message:
+            return message
+
+    elif api_error.code == 403:
+        message = _try_json_message(api_error.body,
+                                    ("forbidden", "message"))
+        if message and _NOVA_403_QUOTA_REACHED.match(message):
+            return message
+
+        for pat in (_NOVA_403_RACKCONNECT_NETWORK_REQUIRED,
+                    _NOVA_403_NO_PUBLIC_NETWORK,
+                    _NOVA_403_PUBLIC_SERVICENET_BOTH_REQUIRED):
+            m = pat.match(api_error.body)
+            if m:
+                return m.groups()[0]
+
+
 @implementer(IStep)
 @attributes(['server_config'])
 class CreateServer(object):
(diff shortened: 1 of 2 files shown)


cyli:

cyli:
On reflection this could probably have been 2 PRs - one for 400's and one for 403's.  I can break this up if that's preferred.

cyli:

otter/convergence/steps.py
```diff
@@ -70,12 +102,38 @@ def got_name(random_name):
                 'POST',
                 'servers',
                 data=thaw(server_config),
-                success_pred=has_code(202))
+                success_pred=has_code(202),
+                reauth_codes=(401,))
```
manishtomar: The default is already `(401, 403)`. Why the change here?

otter/convergence/steps.py
```diff
@@ -70,12 +102,38 @@ def got_name(random_name):
...
 
         def report_success(result):
             return StepResult.SUCCESS, []
 
         def report_failure(result):
+            """
```
manishtomar: I think the parsing should be in separate function: `parse_nova_api_error` that takes `APIError` and returns `message`. We soon need to decide about how to transfer structured info from the error.

lvh: Is that a blocker? If so, why is that function better in a way that won't come out in the (refactoring) wash?

manishtomar: I wouldn't consider it a blocker but would prefer it get done here as this code does more of error parsing than _reporting_. The similar code to parse is also there in [`otter.utils.http.UpstreamError`](https://github.com/rackerlabs/otter/blob/master/otter/util/http.py#L59). I just want to ensure that code is not duplicated. I don't feel particularly strong about it as long as its done eventually. I leave it to @cyli to decide whether to do it here or not :)

cyli: I've moved it to another function, although currently just a private function.  This definitely decreases the complexity of the `report_failure` function, thanks  :)

I think the only part that may be de-duplicatable between this function and `UpstreamError` might be in `_try_json_message`, since UpstreamError attempts to just get a nice message out of something that looks like `{'key': {'message': '...'}}`, and the parsing of nova errors is checking for specific messages and specific keys.

otter/convergence/steps.py
```diff
@@ -50,6 +50,69 @@ def set_server_name(server_config_args, name_suffix):
...
+    """
+    try:
+        error_body = json.loads(maybe_json_error)
+    except (ValueError, TypeError):
+        return None
+    else:
+        return get_in(keys, error_body, None)
+
+
+def _forbidden_plaintext(message):
```
manishtomar: +1. Like the structuring of regexp creation :)

otter/convergence/steps.py
```diff
@@ -50,6 +50,69 @@ def set_server_name(server_config_args, name_suffix):
     return server_config_args.set_in(('server', 'name'), name)
 
 
+def _try_json_message(maybe_json_error, keys):
```
manishtomar: Non-blocking: This is quite generic enough to be in some utils.

cyli: Will do - do you want me to try to make it slightly more generic?  Like `UpstreamError` just grabs the first outer key, so maybe we can add some utility to handle that instead of just a specific key like "badRequest" ?

otter/convergence/steps.py
cyli: As a clarification - we can make this a public function once we determine how we want logging to happen, etc, and how this will interact with `UpstreamError`.  I am just being lazy because I don't want to write tests for a public function yet, if it needs to be refactored soon.  :)

manishtomar: Oh yeah. Definitely :)

otter/convergence/steps.py
```diff
@@ -50,6 +50,69 @@ def set_server_name(server_config_args, name_suffix):
...
+    "11111111-1111-1111-1111-111111111111\) required but missing")
+_NOVA_403_RACKCONNECT_NETWORK_REQUIRED = _forbidden_plaintext(
+    "Exactly 1 isolated network\(s\) must be attached")
+
+_NOVA_403_QUOTA_REACHED = re.compile(
+    "^Quota exceeded for (?P<limit>\S+): Requested \d+, but already used \d+ "
+    "of \d+ (?P=limit)$")
+
+
+def _parse_nova_create_error(api_error):
```
manishtomar: non-blocking: It will be nice if this parses all errors - user and system. For now, (this is not really required), can you rename this to `parse_nova_user_errors`?

otter/convergence/steps.py
```diff
@@ -50,6 +50,69 @@ def set_server_name(server_config_args, name_suffix):
...
+    if api_error.code == 400:
+        message = _try_json_message(api_error.body,
+                                    ("badRequest", "message"))
+        if message:
+            return message
+
+    elif api_error.code == 403:
+        message = _try_json_message(api_error.body,
+                                    ("forbidden", "message"))
+        if message and _NOVA_403_QUOTA_REACHED.match(message):
```
manishtomar: Not that I am suggesting to undo these checks but I was wondering why it is necessary to check that messages conform to specific format? Wouldn't we want to return any message on 400 or 403?

cyli: Mainly because I didn't know if there are any other 403's that may be caused by maybe temporary identity issues - these were the only ones we knew about.  We should probably check with Nova to see we can just blanket count all 403's as permanent, and then yes, definitely we should probably make this generic.

manishtomar:
LGTM :)

cyli:
(sorry sorry, will address the comments above, just slightly delayed :))

otter/convergence/steps.py
cyli: No after further thought, I think you are right, we should probably just make this generic :)

otter/convergence/steps.py
cyli: https://github.com/rackerlabs/otter/issues/1159, since this is a slightly bigger change, and this PR has a big enough diff.

PR #1142, opened 2015-03-02
SingleTenantAuthenticator fixes

manishtomar:
- extracting service catalog correctly
- Removing `expire_in` when requesting tokens. That is only valid in impersonation call.
- changed test tenant id to be strings


radix:
LGTM :)

PR #1143, opened 2015-03-02
Cloudfeeds custom content type

manishtomar:
Sending custom content type when talking to cloud feeds. Changed the preference of headers being passed in `add_headers` to do this.


radix:
LGTM. Not really sure why I made it so add_headers took precedence in the first place...

PR #1144, opened 2015-03-03
cloud feeds uuid fix

manishtomar:
CF was trying to creating JSON with UUID object causing problem. Fixed it by stringing before creating JSON


radix:
Woops! LGTM.

PR #1145, opened 2015-03-03
ISO8601 format time in event

manishtomar:
`datetime.isoformat` does not include `Z` for UTC time that causes problem when adding event in cloud feeds as it expects UTC time appended with `Z`.


radix:
It'd be nice to have a separate test for the new function

radix:
LGTM :)

PR #1146, opened 2015-03-03
event data logged as separate key `event_data`

manishtomar:
instead of directly binding the keys to the logs. This is because the log object got is alredy bound with system=otter and fails when we try to bind it again with system=otter.cloud_feed. This is better because the event data can be explicitly analyzed instead of confusing it with other parts of the log.


radix:
Looks good! Thanks :)

PR #1147, opened 2015-03-03
not listing outdated packages on travis build

manishtomar:
Since no one *looks* at travis build log unless it fails, it is useless to list outdated packages. Instead travis will only lint code. The default `lint` target will still list outdated packages. This will have minor speedup in travis build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
     - pip install coveralls
 
 script:
-    - make lint
+    - make lint-code
     - make docbook
     - make coverage
 
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,9 @@ run:
 env:
 	./scripts/bootstrap-virtualenv.sh
 
-lint: listoutdated flake8diff
+lint: listoutdated lint-code
+
+lint-code: flake8diff
 	pyflakes ${PYDIRS}
 
 listoutdated:


lvh:
This build ran in 4m and change; current master ran in 3m and change. It seems that our build times are still almost entirely determined by how Travis is feeling. I agree that nobody looks at the build, and that's an issue, but I'm not sure simply removing something and not adding something to replace it is an adequate fix. How do we get people to pay (appropriate) attention to outdated builds? Do you think that simply removing this is good enough?

cyli:
Actually even if the build fails, I skip right over the outdated packages and look at why the build failed.  I would suggest we have a timed jenkins job that runs the outdated packages list and emails once a week, but I'm pretty sure everyone will just ignore the email.

I think the issue is more that upgrading packages and other minor tech debt like this is incredibly low in priority since the priority right now is reliability and convergence, besides-which there are a bunch of known failures which require more work to address than just bumping up the requirements file.

One could argue that upgrading packages could also improve reliability, but this only shows which packages were updated.  It's a manual effort to go see what changes occurred between the current version and the next, and whether they are critical.  Also a manual effort to test each package bump.

Maybe if there were automated tests that could test whether an upgraded package works, and automated adding an issue to bump the version for that particular package?  That would decrease the effort involved in making easy (no code change required) version bumps.

radix:
I also like the addition of `lint-code` since I *never* run `make lint` locally because it takes forever when I'm just trying to find all the lint in my change.

lvh:
OK, cool, whatever.

PR #1149, opened 2015-03-04
Option to disable kazoo logs

manishtomar:
Default is enabled. You've to specify `{"no_logs": true}` to disable it. This will helpful when running locally for testing feature

### DIFF
--- a/otter/tap/api.py
+++ b/otter/tap/api.py
@@ -248,12 +248,14 @@ def makeService(config):
     # Setup Kazoo client
     if config_value('zookeeper'):
         threads = config_value('zookeeper.threads') or 10
+        disable_logs = config_value('zookeeper.no_logs')
         kz_client = TxKazooClient(
             hosts=config_value('zookeeper.hosts'),
             # Keep trying to connect until the end of time with
             # max interval of 10 minutes
             connection_retry=dict(max_tries=-1, max_delay=600),
-            threads=threads, txlog=log.bind(system='kazoo'))
+            threads=threads,
+            txlog=None if disable_logs else log.bind(system='kazoo'))
         # Don't timeout. Keep trying to connect forever
         d = kz_client.start(timeout=None)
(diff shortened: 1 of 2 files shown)


lvh:
Can you explain why this managed inside Zookeeper, instead of just managing local log observers?

manishtomar:
It didn't strike me to do this via a log observer. However, I think it is better like this because:
1) The logs will be generated in the system only to realize its not required by the observer. Why not filter it at the source if there is an option?
2) You'll need to create another log observer and make it part of the chain. This will include tests. Comparatively more code
3) I would rather not prefer to use `config_value` global config in some other part of the code. Here, the code using `zookeeper` settings is small and in one place.

cyli:
Either way this is implemented, +1 on the feature!  I have a local branch that does this (and also disables the scheduler logs), which I merge into branches when I test.

Super helpful.

cyli:
@lvh Is there a blocking reason why it should be managed by local log observers rather than by zookeeper for testing?

lvh:
To answer the points:

1. The same reason you have any observer that filters some messages: so that not every piece of code needs to know about producing log messages being important or not.
2. This code also needs tests. A log observer is a lot simpler to test than Zookeeper interactions; that's just danced around here because of the mock.
3. By that logic, why do we have config_value at all? Just dump everything in Zookeeper, right? (That's not intrinsically a bad idea; I'm suggesting that exceptions to the rule of where you store configuration is a bad idea.)

I think this is bad, but I am not going to block it, and mostly for bad reasons.

lvh:
I misunderstood how this worked; while I still think an observer would be better, this is totally fine. Sorry about that. (Thanks @cyli for figuring it out :-))

manishtomar:
> By that logic, why do we have config_value at all?

@lvh We shouldn't and I want to avoid using anymore if possible.

PR #1152, opened 2015-03-04
Log impersonation failures

wirehead:
Logging a string we can search against in ELK.

### DIFF
--- a/otter/auth.py
+++ b/otter/auth.py
@@ -239,7 +239,7 @@ class ImpersonatingAuthenticator(object):
             """
             Log this as a string we know we can find in the logging feed
             """
-            log.err(err, 'Failed to get a new identity admin token.', 
+            log.err(err, 'Failed to get a new identity admin token.',
                     otter_msg_type='impersonation-failed')
             return err
 
--- a/otter/auth.py
+++ b/otter/auth.py
@@ -235,6 +235,12 @@ class ImpersonatingAuthenticator(object):
 
     @wait(ignore_kwargs=['log'])
     def _auth_me(self, log=None):
+        def _log_failed_auth(err):
+            """
+            Log this as a string we know we can find in the logging feed
+            """
+            log.err('Failed to do identity impersonation')
+
         if log:
             log.msg('Getting new identity admin token')
         d = authenticate_user(self._url,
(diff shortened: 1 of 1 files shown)


otter/auth.py
```diff
@@ -235,6 +235,12 @@ def __init__(self, identity_admin_user, identity_admin_password, url, admin_url)
 
     @wait(ignore_kwargs=['log'])
     def _auth_me(self, log=None):
+        def _log_failed_auth(err):
+            """
+            Log this as a string we know we can find in the logging feed
+            """
+            log.err('Failed to do identity impersonation')
```
radix: I suggest adding an `otter_msg_type='impersonation-failed'`, to give us a nicer identifier to filter on (in addition to the English message you have now).

radix: Oh, and two more things:

- you want to pass `err` as the first argument to `log.err`. That function takes a failure or exception as its first argument, and the message as its second argument.
- You should `return err` from `_log_failed_auth`, because returning None will cause it to consider the Deferred "successful", which will cause the next callback to run when it wouldn't before this change.

otter/auth.py
```diff
@@ -235,6 +235,13 @@ def __init__(self, identity_admin_user, identity_admin_password, url, admin_url)
 
     @wait(ignore_kwargs=['log'])
     def _auth_me(self, log=None):
+        def _log_failed_auth(err):
+            """
+            Log this as a string we know we can find in the logging feed
+            """
+            log.err(err, 'Failed to get a new identity admin token.', otter_msg_type='impersonation-failed')
```
radix: Line too long, that's why the build's failing. sorry :P lint now expects <80 column lines.

otter/auth.py
```diff
@@ -235,6 +235,15 @@ def __init__(self, identity_admin_user, identity_admin_password, url, admin_url)
 
     @wait(ignore_kwargs=['log'])
     def _auth_me(self, log=None):
+        def _log_failed_auth(err):
+            """
+            Log this as a string we know we can find in the logging feed
+            """
+            if log:
+                log.err(err, 'Failed to get a new identity admin token.',
+                        otter_msg_type='impersonation-failed')
```
radix: I hate to keep nitpicking (especially since it's nitpicking *my own* suggestion), but now that I saw your new message here and read the surrounding code, I realize that this isn't an error in the impersonation step, but just the admin login step, so I guess the otter_msg_type should be 'admin-login-failed'. If you wanna make this change you can go ahead and merge right after that. Sorry & thanks. :-)

stretchservers:

radix:
LGTM! (ha) merging

PR #1153, opened 2015-03-05
Add intents for certain interactions with ZooKeeper

radix:
I split this change out of the `dirty-groups-2` branch, where I'm implementing dirty-tracking for Convergence.

For context, here's how I'm using them:

- `CreateOrSet`: for creating the dirty flag (or updating it when it already exists). We need this behavior so the version number gets bumped for every operation that potentially requires convergence.
- `GetChildrenWithStats`: for listing all the dirty flags, along with their stat information so we can know what version they're at.
- `DeleteNode`: for deleting the dirty flag, with a specified version so if the version has change the delete will fail.

The PR itself should be pretty easy to review. I didn't add stuff to `otter.effect_dispatcher` yet because there are further changes coming to that file and I don't want to create conflicts for myself.

### DIFF
--- a/otter/test/util/test_zk.py
+++ b/otter/test/util/test_zk.py
@@ -13,7 +13,8 @@ from twisted.internet.defer import fail, succeed
 from twisted.trial.unittest import SynchronousTestCase
 
 from otter.util.zk import (
-    CreateOrSet, DeleteNode, GetChildrenWithStats,
+    CreateOrSet, CreateOrSetLoopLimitReachedError,
+    DeleteNode, GetChildrenWithStats,
     perform_create_or_set, perform_delete_node,
     perform_get_children_with_stats)
 
@@ -121,6 +122,25 @@ class CreateOrSetTests(SynchronousTestCase):
         # It must be at version 0 because it's a creation, whereas normally if
         # the node were being updated it'd be at version 1.
         self.assertEqual(self.model.nodes, {'/foo': ('bar', 0)})
+
+    def test_loop_limit(self):
+        """
+        performing a :obj:`CreateOrSet` will only try to loop between creating
+        and setting 100 times.
+        """
+        def hacked_set(path, value):
+            return fail(NoNodeError())
+
+        def hacked_create(path, content, makepath):
+            return fail(NodeExistsError())
+
+        self.model.set = hacked_set
+        self.model.create = hacked_create
+
+        d = self._cos('/foo', 'bar')
+        failure = self.failureResultOf(d)
+        self.assertEqual(failure.type, CreateOrSetLoopLimitReachedError)
+        self.assertEqual(failure.getErrorMessage(), '/foo')
 
 
 class GetChildrenWithStatsTests(SynchronousTestCase):
--- a/otter/test/util/test_zk.py
+++ b/otter/test/util/test_zk.py
@@ -1,0 +1,174 @@
+"""Tests for otter.util.zk"""
+
+from functools import partial
+
+from characteristic import attributes
+
+from effect import Effect, TypeDispatcher
+from effect.twisted import perform
+
+from kazoo.exceptions import BadVersionError, NoNodeError, NodeExistsError
+
+from twisted.internet.defer import fail, succeed
+from twisted.trial.unittest import SynchronousTestCase
+
+from otter.util.zk import (
+    CreateOrSet, DeleteNode, GetChildrenWithStats,
+    perform_create_or_set, perform_delete_node,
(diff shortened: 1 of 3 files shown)


otter/util/zk.py
```diff
@@ -0,0 +1,115 @@
...
+    content = create_or_set.content
+
+    def create():
+        d = kz_client.create(path, content, makepath=True)
+        d.addErrback(_handle(NodeExistsError, set_content))
+        return d
+
+    def set_content():
+        d = kz_client.set(path, content)
+        d.addErrback(_handle(NoNodeError, create))
```
manishtomar: Non-blocking: This can go in infinite loop. I do realize that it shouldn't if the exceptions returned are accurate. But I am not sure if you want to allow this path to exist.

radix: Yeah, I realized that while developing but didn't bother doing anything about it. I'll add a counter to limit the number of loops.

radix:
addressed your comment @manishtomar :-)

otter/util/zk.py
```diff
@@ -0,0 +1,132 @@
...
+def perform_get_children_with_stats(kz_client, dispatcher, intent):
+    """
+    Perform :obj:`GetChildrenWithStats`. Must be partialed with ``kz_client``.
+
+    :param kz_client: txKazoo client
+    :param dispatcher: dispatcher, supplied by perform
+    :param GetChildrenWithStats intent: the intent
+    """
+    path = intent.path
+    children = kz_client.get_children(path)
```
manishtomar: You can use `get_children(path, include_data=True)` as suggested in the [docs](http://kazoo.readthedocs.org/en/latest/api/client.html#kazoo.client.KazooClient.get_children)

otter/util/zk.py
```diff
@@ -0,0 +1,132 @@
...
+    """Perform :obj:`DeleteNode`.
+
+    :param kz_client: txKazoo client
+    :param dispatcher: dispatcher, supplied by perform
+    :param DeleteNode intent: the intent
+    """
+    kz_client.delete(intent.path, version=intent.version)
+
+
+def _handle(exc_type, fn, *args, **kwargs):
```
manishtomar: This is pretty useful. Would you mind putting this in `deferredutils`?

otter/util/zk.py
radix: wow, I missed that! I think I saw the name of the parameter and assumed it returned the *content*, but thanks for pointing out it returns the stat object. I'll fix this :)

otter/test/util/test_zk.py
```diff
@@ -0,0 +1,194 @@
...
+        d = self._cos('/foo', 'bar')
+        self.assertEqual(self.successResultOf(d), '/foo')
+        # It must be at version 0 because it's a creation, whereas normally if
+        # the node were being updated it'd be at version 1.
+        self.assertEqual(self.model.nodes, {'/foo': ('bar', 0)})
+
+    def test_loop_limit(self):
+        """
+        performing a :obj:`CreateOrSet` will only try to loop between creating
+        and setting 100 times.
```
manishtomar: Non-blocking: The test is not really checking the limit of 100. You could change the doc to "performing createorset will not go in infinite loop and eventually give up with CreateOrSetLimitReachedError"

otter/test/util/test_zk.py
```diff
@@ -0,0 +1,194 @@
...
+        and ``exists`` are not returned.
+        """
+        def exists(p):
+            if p == '/path/foo':
+                return succeed(ZNodeStatStub(version=0))
+            if p == '/path/bar':
+                return succeed(ZNodeStatStub(version=1))
+            if p == '/path/baz':
+                return succeed(None)
+        self.model.get_children = {'/path': succeed(['foo', 'bar', 'baz'])}.get
```
manishtomar: Clever! :)

otter/test/util/test_zk.py
```diff
@@ -0,0 +1,194 @@
...
+    perform_create_or_set, perform_delete_node,
+    perform_get_children_with_stats)
+
+
+@attributes(['version'])
+class ZNodeStatStub(object):
+    """Like a :obj:`ZnodeStat`, but only supporting the data we need."""
+
+
+class ZKCrudModel(object):
```
manishtomar: This is pretty cool!

manishtomar:
LGTM modulo nitpick :)

radix:
Thank you @manishtomar!

radix:
jenkins please run unit tests

PR #1154, opened 2015-03-05
add GetScalingGroupInfo, an intent looks up all the info about a group

radix:
This is necessary so the convergence code can get the group info at the start of each convergence cycle.

I didn't add the performers to `otter.effect_dispatcher` because I have some other changes coming to that file and I don't want to create unnecessary conflicts.

### DIFF
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -23,7 +23,7 @@ def perform_modify_group_state(dispatcher, mgs_intent):
 
 @attributes(['tenant_id', 'group_id'])
 class GetScalingGroupInfo(object):
-    """Look up a scaling group, along with its `state` and launch config."""
+    """Get a scaling group and its manifest."""
 
 
 @deferred_performer
@@ -37,15 +37,14 @@ def perform_get_scaling_group_info(log, store, dispatcher, intent):
     :param dispatcher: dispatcher provided by perform
     :param GetScalingGroupInfo intent: the intent
     """
-    group = yield store.get_scaling_group(
-        log, intent.tenant_id, intent.group_id)
-    state = yield group.view_state()
-    lc = yield group.view_launch_config()
-    returnValue((group, state, lc))
+    group = store.get_scaling_group(log, intent.tenant_id, intent.group_id)
+    manifest = yield group.view_manifest(with_policies=False,
+                                         with_webhooks=False)
+    returnValue((group, manifest))
 
 
-def get_cassandra_dispatcher(log, store):
-    """Get a dispatcher that can handle all the cass-related intents."""
+def get_model_dispatcher(log, store):
+    """Get a dispatcher that can handle all the model-related intents."""
     return TypeDispatcher({
         ModifyGroupState:
             perform_modify_group_state,
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -1,6 +1,11 @@
+from functools import partial
+
 from characteristic import attributes
 
+from effect import TypeDispatcher
 from effect.twisted import deferred_performer
+
+from twisted.internet.defer import inlineCallbacks, returnValue
 
 
 @attributes(['scaling_group', 'modifier'])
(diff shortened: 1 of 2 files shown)


otter/models/intents.py
```diff
@@ -14,3 +19,35 @@ class ModifyGroupState(object):
...
+    Perform :obj:`GetScalingGroupInfo`.
+
+    :param log: bound log
+    :param IScalingGroupCollection store: collection of groups
+    :param dispatcher: dispatcher provided by perform
+    :param GetScalingGroupInfo intent: the intent
+    """
+    group = yield store.get_scaling_group(
+        log, intent.tenant_id, intent.group_id)
+    state = yield group.view_state()
```
cyli: Non-blocking:  since `view_manifest` got changed to deal with force-deleting groups (so you can return the state, the config, and the launch config), it can probably be used here to make this a single cass call instead of 2.  That may or may not be a pointless optimization.

otter/models/intents.py
```diff
@@ -14,3 +19,35 @@ class ModifyGroupState(object):
...
+    lc = yield group.view_launch_config()
+    returnValue((group, state, lc))
```
cyli: Non-blocking question - is the group itself needed because of `ModifyGroupState`?

otter/test/models/test_intents.py
```diff
@@ -15,6 +19,32 @@ def test_perform(self):
                                modifier=lambda g, o: 'new state')
         dispatcher = TypeDispatcher({
```
cyli: I know this part wasn't really changed, but would it make sense to just use the `get_cassandra_dispatcher` here, so that it tests that `get_cassandra_dispatcher` can get performers for `ModifyGroupState`?

otter/test/models/test_intents.py
```diff
@@ -15,6 +19,32 @@ def test_perform(self):
...
+    def test_perform(self):
+        """Performing returns the group, the state, and the launch config."""
+        log = mock_log()
+        state = object()
+        lc = object()
+        group = mock_group(state)
+        group.view_state.return_value = succeed(state)
+        group.view_launch_config.return_value = succeed(lc)
+
+        data = {('00', 'g1'): group}
```
cyli: Nice!  To make sure it has the right tenant ID and group ID?  Non-blocking just-curiosity question - why the dictionary here, but the assertion about the log below?  Are there going to be more groups here in a future PR?

otter/test/models/test_intents.py
radix: Yes, it would, I'll make that change.

otter/test/models/test_intents.py
radix: No, I just wasn't thinking straight I guess. Thanks for the suggestion :) I'll put it in there.

otter/test/models/test_intents.py
```diff
@@ -15,6 +19,32 @@ def test_perform(self):
...
+
+        class Store(object):
+            def get_scaling_group(s_self, _log, tenant_id, group_id):
```
cyli: Nonblocking - this is probably overkill, but we've had problems in the past where we've changed method signatures and interfaces without changing the mocks for tests, and hence unit tests would falsely pass.  Hence the [iMock](https://github.com/rackerlabs/otter/blob/master/otter/test/utils.py#L191) and otherwise providing a spec to `Mock`.

Would it make sense to include an assertion that the mock method signature remains the same as the actual store's `get_scaling_group` signature, so it's a verified fake?

I'm also fine with just adding a tech-debt ticket - we are unlikely to change the store interface, particularly this one, any time in the near future.

otter/models/intents.py
radix: Yes, in the code that actually uses this intent, we then go on to use ModifyGroupState.

otter/test/models/test_intents.py
radix: okie doke, I've updated it to use iMock :)

cyli:
LGTM if travis +1's

PR #1155, opened 2015-03-05
Miscellaneous testing utility fixes/improvements

radix:
- A number of classes with `__eq__` didn't support `__ne__`, which means they wouldn't do the right thing when != was used.
- move FakePartitioner from test_scheduler to utils.py, so it can be used in the forthcoming convergence service tests, which also use the partitioner.
- add a `transform_eq` function, which transforms an object before quality comparison.
- make mock_log's log.err and log.msg functions return None, like the real functions do.

e.g., if you have a data structure `d = {'foo': [1,2,3]}`, and you want to compare it in your tests, but you don't care about the order of the list, you can use

`self.assertEqual(d, {'foo': transform_eq(set, set([1,2,3]))})`

### DIFF
--- a/otter/test/test_scheduler.py
+++ b/otter/test/test_scheduler.py
@@ -5,7 +5,6 @@ from datetime import datetime, timedelta
 
 import mock
 
-from twisted.application.service import Service
 from twisted.internet import defer
 from twisted.trial.unittest import SynchronousTestCase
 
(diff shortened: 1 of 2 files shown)


cyli:
Yay, `transform_eq` seems pretty useful!  LGTM if travis +1's

cyli:
boo lint.

radix:
Ok, merging since the only change was a trivial lint fix. Thanks @cyli :)

PR #1156, opened 2015-03-06
iMock test fixes

cyli:
When working on https://github.com/rackerlabs/otter/pull/1154 @radix discovered that iMock doesn't actually check the signature on the interface methods.

Add some fixes that auto-specs the methods, and also ensures that the attributes are not callable.

After this change, some of the unit tests were broken, so fix them.

### DIFF
--- a/otter/models/interface.py
+++ b/otter/models/interface.py
@@ -855,7 +855,7 @@ class IAdmin(Interface):
     Interface to administrative information and actions.
     """
 
-    def get_metrics(self, log):
+    def get_metrics(log):
         """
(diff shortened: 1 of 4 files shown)


otter/models/interface.py
```diff
@@ -855,7 +855,7 @@ class IAdmin(Interface):
     Interface to administrative information and actions.
     """
 
-    def get_metrics(self, log):
+    def get_metrics(log):
```
lvh: Whoops :-)

lvh:
LGTM! Thanks for working on this :)

Issue #881, opened 2014-12-31
Remove deleted server's nodes from LBs


manishtomar:
Convergence should remove any out-of-band deleted servers from all the LBs it was attached to. This requires getting deleted servers from `get_scaling_group_servers` which needs time since it was last fetched. [Docs](http://docs.rackspace.com/servers/api/v2/cs-devguide/content/ChangesSince.html) for this. Otherwise Nova will return all servers deleted in last 30 days. It is not clear where this state will be stored. It would most likely be otter db. So, this task can be divided in:

UPDATE 05/22: Based on discussion below following tasks:
- [x] Server cache model changes
- [x] Convergence gathering from cache if it exists (updating if it doesn't)
- [x] Deleting old cache after convergence succeeds
- [ ] Heuristic check for CLB IP reuse: Check 2nd point of https://github.com/rackerlabs/otter/issues/881#issuecomment-102537953
- [x] Add `changes_since` support in mimic: https://github.com/rackerlabs/mimic/issues/252
- [x] Add support to get deleted servers in `get_server_details` and `get_scaling_group_servers`: #1252
- [x] Change `converge` to remove dangling LB nodes

cyli:
Discussion from Slack:

We probably don't need to store any server data, we just don't update the last-fetched date until we have successfully cleaned up all the nodes from the LB. 

We can probably do other optimizations to that date too.

There is a possibility that (especially the farther back the server was deleted) the IP from that server has been reused, and cleaning that IP off of a CLB will result in removing a currently valid server from the CLB.

We should ask how long it is before servicenet addresses are re-used.

cyli:
Also, this may not be testable in Mimic is because Mimic does not support changes-since yet.

cyli:
@manishtomar and @wirehead and I were discussing this today, and a couple of things came out of the discussion:

1.  @manishtomar suggested a solution that means we don't have to do any fancy figuring out of which servers have been cleaned up and which have yet to be cleaned up:
    1.  A convergence cycle starts, and we gather any deleted servers since time T=0.  The gathering happens at time T=1.
    2.  If any of those servers are still attached to any load balancers, this convergence cycle will attempt to remove them.
    3.  Hence if the convergence cycle succeeds at time T=2, that means that the load balancer nodes were all removed successfully (or that the server was not on any load balancers to begin with), and so we can record T=1 the last fetched time.
[...]

PR #1164, opened 2015-03-10
Clean up dangling load balancers from deleted servers

cyli:
One of the tasks in #881.

This will remove (in planning) all load balancers associated with a server if that server has DELETED server state.


radix:
Awesome tiny change! Thank you @cyli :)

PR #1173, opened 2015-03-10
Update effect_dispatcher to have a bunch more stuff

radix:
- ZK dispatcher from otter.util.zk
- model dispatcher from otter.models.intents
- reference dispatcher from effect

also adds a get_legacy_dispatcher for use in supervisor, which doesn't need the new stuff.


manishtomar:
LGTM!

Issue #1170, opened 2015-03-10
AddNodesToCLB should return success, [reasons]


lvh:

manishtomar:
Closing as this was fixed in #1210.

PR #1175, opened 2015-03-10
StepResults in SetMetadataItem and AddNodesToCLB

### LINKED ISSUE #1169: SetMetadataItemOnServer should return success, [reasons]

cyli:
Fixes #1169 and #1170.

Modifies the tests to test for the StepResult as the final result.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -275,7 +275,7 @@ class AddNodesToCLB(object):
 
     def as_effect(self):
         """Produce a :obj:`Effect` to add nodes to CLB"""
-        return service_request(
+        eff = service_request(
             ServiceType.CLOUD_LOAD_BALANCERS,
             'POST',
             append_segments('loadbalancers', str(self.lb_id), "nodes"),
@@ -288,6 +288,14 @@ class AddNodesToCLB(object):
                 has_code(202, 413),
                 _check_clb_422(_CLB_DUPLICATE_NODES_PATTERN,
                                _CLB_PENDING_UPDATE_PATTERN)))
+
+        def report_success(result):
+            return StepResult.SUCCESS, []
+
+        def report_failure(result):
+            return StepResult.FAILURE, []
+
+        return eff.on(success=report_success, error=report_failure)
 
 
 @implementer(IStep)
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -210,7 +210,7 @@ class SetMetadataItemOnServer(object):
             return StepResult.SUCCESS, []
 
         def report_failure(result):
-            return StepResult.RETRY, []
+            return StepResult.RETRY, [result[1]]
 
         return eff.on(success=report_success, error=report_failure)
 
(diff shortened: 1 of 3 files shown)


otter/convergence/steps.py
```diff
@@ -192,14 +193,26 @@ class SetMetadataItemOnServer(object):
...
             append_segments('servers', self.server_id, 'metadata', self.key),
-            data={'meta': {self.key: self.value}})
+            data={'meta': {self.key: self.value}},
+            success_pred=has_code(200, 404))
+
+        def report_success(result):
+            return StepResult.SUCCESS, []
+
+        def report_failure(result):
+            return StepResult.RETRY, []
```
radix: I think you might as well put result[1] into the list here

otter/convergence/steps.py
```diff
@@ -279,8 +292,25 @@ def as_effect(self):
...
+            If the error is a 422 - PENDING UPDATE, retry.  Otherwise, fail.
+            """
+            err_type, error, traceback = result
+            if err_type == APIError and error.code == 422:
+                # body has already become JSON
+                message = error.body.get("message", None)
+                if message and _CLB_PENDING_UPDATE_PATTERN.search(message):
+                    return StepResult.RETRY, []
+
+            return StepResult.FAILURE, []
```
radix: likewise in these spots

otter/convergence/steps.py
cyli: Oh is that happening now?  cool

manishtomar:
It will be nice if their are specific exceptions like `CLBPendingUpdate` that is moved around instead of `APIError`. Now, I am not sure if should compose of or inherit from `APIError` but will be nice if we define such exceptions and use them through out. I also feel that error parsing code should be more modular and separate from reporting code. This will be easier with more structured exceptions. Otherwise, we might end up with bits and pieces of error parsing in different step implementations.

manishtomar:
I am not sure if it should be done in this PR but we are getting to a point where it is needed.

cyli:
@manishtomar Good point.  I think it's worth doing in this PR.  I think I'm leaning towards ignoring APIError - besides which APIError may be getting replaced with UpstreamError?

manishtomar:
> besides which APIError may be getting replaced with UpstreamError?

@cyli I am torn :/ https://github.com/rackerlabs/otter/issues/1184#issuecomment-78184828

otter/test/utils.py
```diff
@@ -711,3 +712,41 @@ def __ne__(self, other):
...
+        }
+        eff = concretize_service_request(
+            authenticator=mock.MagicMock(),
+            log=mock.MagicMock(),
+            service_configs=service_configs,
+            tenant_id='000000',
+            service_request=service_request_intent)
+
+        # "authenticate"
+        eff = resolve_effect(eff, ('token', []))
```
radix: Makes me want to switch this to EQDispatcher :)

radix:
LGTM!

radix:
Oh, sorry, didn't notice you're still working on stuff here. I'll put it into in-progress.

cyli:
As per IRL discussion, the points above should probably be addressed in an other PR.  So merged forward, and if passes, I'll merge?

radix:
@cyli yes please, thanks :)

PR #1178, opened 2015-03-10
Hopefully make a scheduler test less flaky

### LINKED ISSUE #1100: update schedule policy test has incorrect wait
`test_system_update_change_for_cron_style_scheduled_policy`. The intent of the test is to ensure that updating cron policy change does not affect the execution. To do that the test should wait for cron to execute after creating the policy and before updating the change. Right now, it doesn't wait for execution after creating policy.

cyli:
As per the issue, wait for a policy execution before updating the policy on this scheduled policy test, since it's possible that the policy executed before the policy gets updated.

Fixes #1100, hopefully.  @manishtomar?

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -407,23 +407,28 @@ class AutoscaleFixture(BaseTestFixture):
                     api=api, asserter=self))
 
     def wait_for_expected_group_state(self, group_id, expected_servers,
-                                      wait_time=180):
+                                      wait_time=180, interval=None):
         """
         :summary: verify the group state reached the expected servers count.
         :param group_id: Group id
         :param expected_servers: Number of servers expected
         """
+        if interval is None:
+            interval = self.interval_time
+
         end_time = time.time() + wait_time
         while time.time() < end_time:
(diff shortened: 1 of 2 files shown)


manishtomar:
LGTM once travis builds :)

cyli:
@lekhajee - does this look ok to you?

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -407,23 +407,28 @@ def wait_for_expected_number_of_active_servers(self, group_id,
...
             if group_state.desiredCapacity == expected_servers:
                 return
-            time.sleep(self.interval_time)
+            time.sleep(interval)
         else:
             self.fail(
-                "wait_for_exepected_group_state ran for 180 seconds for group {0} and did not "
-                "observe the active server list achieving the expected servers count: {1}.".format(
-                    group_id, expected_servers))
+                "wait_for_exepected_group_state ran for 180 seconds for group "
```
lekhajee: the `180 seconds` that was defaulted would need to be changed to be `wait_time`

cyli: Nice catch, thanks!

lekhajee:
+1 except the nit, thanks!

PR #1182, opened 2015-03-11
Create Server should return RETRY

### LINKED ISSUE #1181: CreateServer should return a RETRY on success
Since CreateServer is always asynchronous, and we need to wait until it's ACTIVE, we always need to retry convergence after sending the CreateServer request.

cyli:
Fixes #1181


radix:
thank you @cyli!

PR #1183, opened 2015-03-11
Instead of StopIteration, public_endpoint_url now raises NoSuchEndpoint

cyli:
So change that in the integration test library


kivattik:
+1
Informative exceptions FTW :)

Issue #818, opened 2014-12-11
Implement `ConvergeLater` step


manishtomar:
This step will be returned if there are no steps to be performed but convergence need to be called again. One concrete example is when removing servers from CLB they are first put into draining and removed after timeout has expired. If `converge` is called when they are in draining and timeout has not expired then there are no steps to be be performed but convergence should be called again. 

This step can also be useful to implement rate-limiting a particular service as discussed in https://github.com/rackerlabs/otter/issues/801#issuecomment-65518455

lvh:
The limiter introduced by #939 should also also limit the number of `ConvergeLater` steps to 1.

radix:
This was resolved by #1185. We need to start using it, and we should have separate tickets for that.

PR #1185, opened 2015-03-11
ConvergeLater step

manishtomar:
Implements #818

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -561,7 +561,6 @@ def _rcv3_check_bulk_delete(attempted_pairs, result):
 
 
 @implementer(IStep)
-@attributes([Attribute('seconds', instance_of=float)])
 class ConvergeLater(object):
     """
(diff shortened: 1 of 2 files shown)


otter/convergence/steps.py
```diff
@@ -558,3 +558,17 @@ def _rcv3_check_bulk_delete(attempted_pairs, result):
         return next_step.as_effect()
     else:
         return StepResult.SUCCESS, []
+
+
+@implementer(IStep)
+@attributes([Attribute('seconds', instance_of=float)])
```
radix: I thought we decided against having a time encoded in the step? Given that we don't even have a mechanism to communicate this time back to any other code I think we should just leave it out.

manishtomar: I was thinking on similar lines. Just wanted another eyes to have a look at it :) Will change it.

radix:
LGTM!

PR #1187, opened 2015-03-11
Keep track of divergent groups in ZooKeeper

### LINKED ISSUE #971: Keep track of "dirty" groups in zookeeper for Convergence
Consider the usual flow:

1. policy execution happens to add one server (where a config has an associated LB)
2. we converge, CreateServer is run, convergence returns a RETRY since we need to wait for the server to build
3. we reconverge because converge returned a RETRY
4. converge (gather data step finds the server ACTIVE), AddToLB is run, convergence returns a status indicating that 

If Otter gets shut down between step 2 and 3, when it restarts, it doesn't know that it should try to reconverge this group. It may be years before the next policy execution or config change (or observer event), so we can't rely on it "naturally" occuring.

So, we should store a flag in ZK for a group when we first start converging. When Otter starts up, it will search for all flagged groups and immediately reconverge. Only when a convergence returns DONE will the flag be removed from ZK.

radix:
This fixes #971.

The main benefit of this PR is that it finally allows us to survive a restart while we're in the middle of performing actions on behalf of the user. The main changes are:

- the `Converger` service has been split up into two parts: `ConvergenceStarter`, which is what the controller calls `start_convergence` on when it executes a policy or obeys a config change, and `Converger` which actually does the work of convergence (potentially on a different node than the API node where `start_convergence` was invoked).
- The only thing `ConvergenceStarter` does is create (or update) a ZNode in ZooKeeper, at a path like `/groups/divergent/{tenant_id}_{group_id}`.
-  `Converger` uses a ZooKeeper set partitioner to divvy up the work of converging the set of divergent groups between all servers running a Converger.
- We converge all groups every 10 seconds (as long as we're not already converging them).
- When the converger is done converging a group, it deletes the ZNode which indicates it's divergent *only if its version has not changed*. This is an important behavior that allows us to avoid a race condition where a second policy execution happens *just* as an existing convergence finishes up. This behavior ensures that the second policy execution will cause another convergence. Thanks to @cyli for coming up with this elegant design :)

Note to reviewers: I recommend jumping straight to ConvergenceStarter and Converger, reading them (they're very small) and then following the chain of dependent functions down from there, in a top-down fashion. (For example, `Converger.buckets_acquired` calls `converge_all` calls `converge_one` calls `execute_convergence`). I think it'll help to review it in that order so you understand the context that each function is used in.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -7,6 +7,8 @@ from functools import partial
 from effect import Effect
 from effect.twisted import perform
 
+from pyrsistent import thaw
+
 from toolz.itertoolz import concat
 
 from twisted.application.service import Service
@@ -28,7 +30,7 @@ def server_to_json(server):
     Convert a NovaServer to a dict representation suitable for returning to the
     end-user as a part of group state.
     """
-    return {'id': server.id}
+    return {'id': server.id, 'links': thaw(server.links)}
 
 
 def determine_active(servers, lb_nodes):
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -16,7 +16,7 @@ import six
 
 from twisted.application.service import MultiService, Service
 
-from otter.constants import CONVERGENCE_DIRTY_DIR, CONVERGENCE_DIRTY_PATH
+from otter.constants import CONVERGENCE_DIRTY_DIR
 from otter.convergence.composition import get_desired_group_state
 from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
@@ -122,6 +122,16 @@ def execute_convergence(tenant_id, group_id, log,
     ))
 
 
+def format_dirty_flag(tenant_id, group_id):
+    """Format a dirty flag ZooKeeper node name."""
+    return tenant_id + '_' + group_id
+
+
+def parse_dirty_flag(flag):
+    """Parse a dirty flag ZooKeeper node name into (tenant_id, group_id)."""
+    return flag.split('_', 1)
+
+
 def mark_divergent(tenant_id, group_id):
     """
     Indicate that a group should be converged.
@@ -153,7 +163,7 @@ def mark_divergent(tenant_id, group_id):
     # group is marked clean, then the changes desired by the second policy
     # execution will not happen.
 
-    # So instead of just a boolean flag, we'll takie advantage of ZK node
+    # So instead of just a boolean flag, we'll take advantage of ZK node
     # versioning. When we mark a group as dirty, we'll create a node for it
     # if it doesn't exist, and if it does exist, we'll write to it with
     # `set`. The content doesn't matter - the only thing that does matter
@@ -171,8 +181,8 @@ def mark_divergent(tenant_id, group_id):
     # changed), only if there are _any_ outstanding requests for
     # convergence, since convergence always uses the most recent data.
 
-    path = CONVERGENCE_DIRTY_PATH.format(tenant_id=tenant_id,
-                                         group_id=group_id)
+    flag = format_dirty_flag(tenant_id, group_id)
+    path = CONVERGENCE_DIRTY_DIR + '/' + flag
     eff = Effect(CreateOrSet(path=path, content='dirty'))
     return eff
 
@@ -184,8 +194,8 @@ def delete_divergent_flag(log, tenant_id, group_id, version):
 
     :return: Effect of None.
     """
-    path = CONVERGENCE_DIRTY_PATH.format(tenant_id=tenant_id,
-                                         group_id=group_id)
+    flag = format_dirty_flag(tenant_id, group_id)
+    path = CONVERGENCE_DIRTY_DIR + '/' + flag
     return Effect(DeleteNode(path=path, version=version)).on(
         success=lambda r: log.msg('mark-clean-success'),
         error=lambda e: log.err(exc_info_to_failure(e), 'mark-clean-failure'))
@@ -271,7 +281,7 @@ def get_my_divergent_groups(my_buckets, all_buckets):
     def structure_info(x):
         # Names of the dirty flags are {tenant_id}_{group_id}.
         path, stat = x
-        tenant, group = x[0].split('_', 1)
+        tenant, group = parse_dirty_flag(x[0])
         return {'tenant_id': tenant, 'group_id': group,
                 'version': stat.version}
 
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -14,7 +14,7 @@ from pyrsistent import thaw
 
 import six
 
-from twisted.application.service import MultiService, Service
+from twisted.application.service import MultiService
 
 from otter.cloud_client import TenantScope
 from otter.constants import CONVERGENCE_DIRTY_DIR
@@ -95,8 +95,6 @@ def execute_convergence(tenant_id, group_id, log,
     sg_eff = Effect(GetScalingGroupInfo(tenant_id=tenant_id,
                                         group_id=group_id))
     gather_eff = get_all_convergence_data(group_id)
-
-    log.msg("convergence-gathering")
 
     def got_all_data(((scaling_group, manifest), (servers, lb_nodes))):
         group_state = manifest['state']
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -297,8 +297,8 @@ def get_my_divergent_groups(my_buckets, all_buckets):
     return eff.on(got_children_with_stats)
 
 
-def converge_one(log, group_locks, tenant_id, group_id, version,
-                 execute_convergence=execute_convergence):
+def converge_one_group(log, group_locks, tenant_id, group_id, version,
+                       execute_convergence=execute_convergence):
     """
     Converge one group, non-concurrently, and clean up the dirty flag when
     done.
@@ -316,18 +316,18 @@ def converge_one(log, group_locks, tenant_id, group_id, version,
 
 
 @do
-def converge_all(log, group_locks, my_buckets, all_buckets,
-                 get_my_divergent_groups=get_my_divergent_groups,
-                 converge_one=converge_one):
+def converge_all_groups(log, group_locks, my_buckets, all_buckets,
+                        get_my_divergent_groups=get_my_divergent_groups,
+                        converge_one_group=converge_one_group):
     """
     Check for groups that need convergence and which match up to the
     buckets we've been allocated.
     """
     group_infos = yield get_my_divergent_groups(my_buckets, all_buckets)
-    log.msg('converge-all', group_infos=group_infos)
-    effs = [converge_one(log, group_locks,
-                         info['tenant_id'], info['group_id'],
-                         info['version'])
+    log.msg('converge-all-groups', group_infos=group_infos)
+    effs = [converge_one_group(log, group_locks,
+                               info['tenant_id'], info['group_id'],
+                               info['version'])
             for info in group_infos]
     yield do_return(parallel(effs))
 
@@ -372,7 +372,7 @@ class Converger(MultiService):
     """
 
     def __init__(self, log, dispatcher, buckets, partitioner_factory,
-                 converge_all=converge_all):
+                 converge_all_groups=converge_all_groups):
         """
         :param log: a bound log
         :param dispatcher: The dispatcher to use to perform effects.
@@ -390,7 +390,7 @@ class Converger(MultiService):
         self.partitioner = partitioner_factory(self.log, self.buckets_acquired)
         self.partitioner.setServiceParent(self)
         self.group_locks = make_lock_set()
-        self._converge_all = converge_all
+        self._converge_all_groups = converge_all_groups
 
     def buckets_acquired(self, my_buckets):
         """
@@ -399,10 +399,10 @@ class Converger(MultiService):
         This is used as the partitioner callback.
         """
         self.log.msg('buckets-acquired', my_buckets=my_buckets)
-        eff = self._converge_all(self.log, self.group_locks,
-                                 my_buckets, self._buckets)
+        eff = self._converge_all_groups(self.log, self.group_locks,
+                                        my_buckets, self._buckets)
         result = perform(self._dispatcher, eff).addErrback(
-            self.log.err, 'converge-all-error')
+            self.log.err, 'converge-all-groups-error')
         # the return value is ignored, but we return this for testing
         return result
 
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -336,18 +336,17 @@ def converge_all_groups(log, group_locks, my_buckets, all_buckets,
 
 def _cleanup_convergence_error(log, tenant_id, group_id, version, exc_info):
     failure = exc_info_to_failure(exc_info)
+    if failure.check(ConcurrentError):
+        # We don't need to spam the logs about this, it's to be expected
+        return
     if failure.check(NoSuchScalingGroupError):
         log.err(failure, 'converge-fatal-error')
         return delete_divergent_flag(log, tenant_id, group_id, version)
-    if failure.check(ConcurrentError):
-        # We don't need to spam the logs about this, it's to be expected
-        pass
-    # TODO: Check for other fatal errors, and put the group into ERROR
-    # state.
-    else:
-        log.err(failure, 'converge-non-fatal-error')
-        # We specifically don't clean up the dirty flag in this case, so
-        # convergence will be retried.
+
+    # TODO: Check for other fatal errors, and put the group into ERROR state.
+    log.err(failure, 'converge-non-fatal-error')
+    # We specifically don't clean up the dirty flag in this case, so
+    # convergence will be retried.
 
 
 def _stable_hash(s):
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -96,6 +96,8 @@ def execute_convergence(tenant_id, group_id, log,
                                         group_id=group_id))
     gather_eff = get_all_convergence_data(group_id)
 
+    log.msg("convergence-gathering")
+
     def got_all_data(((scaling_group, manifest), (servers, lb_nodes))):
         group_state = manifest['state']
         launch_config = manifest['launchConfiguration']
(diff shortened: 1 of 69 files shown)


radix:

radix:

otter/tap/api.py
```diff
@@ -286,6 +289,21 @@ def on_client_ready(_):
...
 
 
+def setup_converger(parent, kz_client, dispatcher):
+    """Create a Converger service."""
+    converger_buckets = range(1, 10)
+    partitioner_factory = partial(
+        Partitioner,
+        kz_client,
+        10,  # interval
+        '/converger_partition',
```
cyli: Should this be `CONVERGENCE_PARTITIONER_PATH`?  That doesn't seem to be used elsewhere.

otter/convergence/service.py
```diff
@@ -56,99 +62,352 @@ def all_met(server, current_lb_nodes):
...
 
-class Converger(Service, object):
-    """Converger service"""
 
-    def __init__(self, reactor, kz_client, dispatcher):
-        self._reactor = reactor
-        self._kz_client = kz_client
+class ConvergenceStarter(Service, object):
+    """
+    A service that allows registering interest in convergence, but does not
```
cyli: Nitpick: "A service that accepts reports that a group as diverged and needs converging" or some such?

otter/convergence/service.py
```diff
@@ -56,99 +62,352 @@ def all_met(server, current_lb_nodes):
...
+
+    # Here, a policy was executed on group N twice, and A tried to mark it
+    # dirty twice. The problem is that when the converger node finished
+    # converging, it then marked it clean *after* node A tried to mark it dirty
+    # a second time. It's a small window of time, but if it happens at just the
+    # right moment, after the final iteration of convergence and before the
+    # group is marked clean, then the changes desired by the second policy
+    # execution will not happen.
+
+    # So instead of just a boolean flag, we'll takie advantage of ZK node
```
cyli: Stupid spelling nitpick:  "take" :)

otter/convergence/service.py
```diff
@@ -56,99 +62,352 @@ def all_met(server, current_lb_nodes):
...
+    :param my_buckets: collection of buckets allocated to this node
+    :param all_buckets: collection of all buckets
+
+    :returns: list of dicts, where each dict has ``tenant_id``,
+        ``group_id``, and ``version`` keys.
+    """
+    def structure_info(x):
+        # Names of the dirty flags are {tenant_id}_{group_id}.
+        path, stat = x
+        tenant, group = x[0].split('_', 1)
```
cyli: Completely non-blocking, and this is just yak-shaving - wondering if this should be an object that knows how to convert from `(tenant, group)` -> dirty flag, and back?  Just asking because the string formatter that creates the flag is in another module, but the thing that parses it is here.

otter/tap/api.py
radix: woops! yep it should :)

otter/convergence/service.py
```diff
@@ -56,99 +62,353 @@ def all_met(server, current_lb_nodes):
...
     """
-    all_data_eff = get_all_convergence_data(scaling_group.uuid)
-
-    def got_all_data((servers, lb_nodes)):
-        active = determine_active(servers, lb_nodes)
-        desired_group_state = get_desired_group_state(
-            scaling_group.uuid, launch_config, desired_capacity)
-        steps = plan(desired_group_state, servers, lb_nodes, now)
-        active = {server.id: server_to_json(server) for server in active}
+    # Huh! It turns out we can parallelize the gathering of data with the
```
cyli: Yay!

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,425 @@
...
-    resolve_stubs)
-from otter.util.fp import assoc_obj
+    CheckFailureValue, FakePartitioner, IsBoundWith,
+    matches,
+    mock_group, mock_log,
+    transform_eq)
+from otter.util.zk import CreateOrSet, DeleteNode, GetChildrenWithStats
+
+
+class MarkDivergentTests(SynchronousTestCase):
```
cyli: Totally non-blocking - I am generally all in favor of more test coverage, but this seems like a duplicate test of `test_start_convergence`, since `mark_divergent` seems more like an implementation detail of `start_convergence`?

radix: Yeah, you make a good point, I'll remove it.

otter/test/convergence/test_service.py
```diff
@@ -110,38 +440,54 @@ def setUp(self):
...
+        gsgi = GetScalingGroupInfo(tenant_id='tenant-id',
+                                   group_id='group-id')
+        manifest = {  # Many details elided!
+            'state': self.state,
+            'launchConfiguration': self.lc,
+        }
+        gsgi_result = (self.group, manifest)
+        self.expected_intents = [(gsgi, gsgi_result)]
+
+    def _get_dispatcher(self, expected_intents=None):
```
cyli: Non-blocking - this seems really useful :)  Should this go into test/utils?

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,425 @@
...
 
-        def execute_convergence(group, desired, lc, now, log):
-            return Effect(Constant((group, desired, lc, now, log)))
+        my_buckets = [0, 5]
+        converger = Converger(
+            self.log, self.dispatcher, self.buckets,
+            self._pfactory, converge_all=converge_all)
+
+        result = self.fake_partitioner.got_buckets(my_buckets)
+        self.assertEqual(self.successResultOf(result), 'foo')
```
cyli: +1 for testing that all those assertions above actually get called - I often forget that :D

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,409 @@
...
+            [(self.tenant_id, self.group_id,
+              matches(IsBoundWith(tenant_id=self.tenant_id,
+                                  group_id=self.group_id)))])
+        self.assertEqual(self.deletions, [True])
+        self.log.msg.assert_any_call(
+            'mark-clean-success',
+            tenant_id=self.tenant_id, group_id=self.group_id)
+
+    def test_non_concurrent(self):
+        """Runs execute_convergence non-concurrently, based on the group ID."""
```
cyli: Apologies, this was kind of confusing to me - maybe just "Do not run execute_convergence again, if the group is already converging."?  Since this test isn't actually testing that convergence will be run again later.

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,409 @@
...
+                                                      self.group_id)),
+            'converge-fatal-error',
+            tenant_id=self.tenant_id, group_id=self.group_id)
+
+        self.assertEqual(self.deletions, [True])
+        self.log.msg.assert_any_call(
+            'mark-clean-success',
+            tenant_id=self.tenant_id, group_id=self.group_id)
+
+    def test_concurrent_error_ignored(self):
```
cyli: Non-blocking - not sure if this should be combined with the `test_non_concurrent`, since it seems like the test is that if the converge cycle is already running, it doesn't run again, doesn't cause errors, and doesn't delete anything?

It's fine as is, just slightly confusing since I'd expect `ConcurrentError` to be raised by the locks instead of the effect.

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,409 @@
...
+
+
+class GetMyDivergentGroupsTests(SynchronousTestCase):
+
+    def test_get_my_divergent_groups(self):
+        """
+        :func:`get_my_divergent_groups` gets information about divergent groups
+        that are associated with the given buckets.
+        """
+        # sha1('00') % 10 is 6, sh1('01') % 10 is 1.
```
cyli: Thanks for the explanation!

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,409 @@
...
+
+        def execute_stuff():
+            self.assertEqual(self._get_locks(), pset(['the-key']))
+            return 'foo'
+
+        eff = Effect(Func(execute_stuff))
+
+        non_c_eff = non_concurrently(self.locks, 'the-key', eff)
+        self.assertEqual(sync_perform(dispatcher, non_c_eff), 'foo')
+        # and after convergence, nothing is marked as converging
```
cyli: Nitpick: "and after the effect is resolved, its lock is released" ?

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,409 @@
...
+        self.assertEqual(self._get_locks(), pset([]))
+
+    def test_refuses_concurrency(self):
+        """
+        :func:`non_concurrently` returns None when the key is already locked.
```
cyli: ":func:`non_concurrently` raises `ConcurrentError` when..."?

otter/test/convergence/test_service.py
cyli: Hmm... this seems to have changed/moved since I made this comment.  Possibly I saw a cached/outdated version?  At any rate, this seems more specific now, although maybe a generalized version might be useful at some point in test/utils, my comment is less applicable to this PR now.  :)

otter/test/convergence/test_service.py
```diff
@@ -152,34 +482,67 @@ def test_success(self):
...
-            set([pmap({'weight': 1, 'type': 'PRIMARY', 'port': 80,
-                       'condition': 'ENABLED', 'address': '10.0.0.2'}),
-                 pmap({'weight': 1, 'type': 'PRIMARY', 'port': 80,
-                       'condition': 'ENABLED', 'address': '10.0.0.1'})]))
-        r = resolve_effect(eff, ['stuff'])
-        # The result of the parallel is returned directly
-        # TODO: This must change with issue #844.
-        self.assertEqual(r, ['stuff'])
+        expected_intents = self.expected_intents + [
+            (expected_req.intent, 'stuff')]
```
cyli: Nitpick:  "stuff" -> "successful response" might make it more clear what is being stubbed out here.

otter/tap/api.py
```diff
@@ -286,6 +289,21 @@ def on_client_ready(_):
     return s
 
 
+def setup_converger(parent, kz_client, dispatcher):
+    """Create a Converger service."""
```
cyli: + ", which has a Partitioner as a child service, so that if the Converger is stopped, the partitioner is also stopped"?

manishtomar:
The coverage seems to have reduced to 97.97%. I remember it was above 98%. If coveralls is not failing the build on reduced coverage then I think we should enable coverage comments. Is that ok @radix ?

radix:
@manishtomar I don't think this PR is the right place to discuss it. at least, it doesn't seem my PR is the one that's reducing the coverage, since other PRs currently show even less, at 97.95% coverage.

otter/tap/api.py
manishtomar: @cyli Yeah. Since partitioner is child of converger, it should stop when converger stops. Converger should stop when `parent` stops.

@radix Non-blocking: On a different note, I feel `Converger` should take partitioner arguments and create the partitioner internally. This way, if converger implementation ever changes to not use Partitoner, it should be easy to do that. This is definitely non-blocking :)

cyli: @manishtomar Right, that was more of a "please document this" comment.  :)

manishtomar:
@radix Sorry, you are right. We shouldn't discuss it here.

otter/convergence/service.py
```diff
@@ -56,99 +62,363 @@ def all_met(server, current_lb_nodes):
...
+        self._buckets = buckets
+        self.log = log.bind(system='converger')
+        self.partitioner = partitioner_factory(self.log, self.buckets_acquired)
+        self.partitioner.setServiceParent(self)
+        self.group_locks = make_lock_set()
+        self._converge_all = converge_all
+
+    def buckets_acquired(self, my_buckets):
+        """
+        Perform the effectful result of :func:`check_convergence`.
```
manishtomar: I am not sure where but somewhere there should be doc mentioning this function is called every 10 seconds. Unfortunately, the 10 is `partial`ed in `partitioner_factory` above. So, changing the seconds from caller will outdate the doc here.

radix: this is why I added "This is used as the partitioner callback" in the next line. 10 is basically a configuration number, the important part is that it's used as the partitioner callback.

otter/convergence/service.py
```diff
@@ -56,99 +62,363 @@ def all_met(server, current_lb_nodes):
...
+
+    def got_children_with_stats(children_with_stats):
+        dirty_info = map(structure_info, children_with_stats)
+        num_buckets = len(all_buckets)
+        converging = (
+            info for info in dirty_info
+            if _stable_hash(info['tenant_id']) % num_buckets in my_buckets)
+        return list(converging)
+
+    eff = Effect(GetChildrenWithStats(CONVERGENCE_DIRTY_DIR))
```
manishtomar: Non-blocking: I almost wanna say that you should get only children first and find the ones which this node _owns_ and then find their stats. But that may be an over-optimization not really required at this point. Its good to be aware though :)

otter/convergence/service.py
```diff
@@ -56,99 +62,363 @@ def all_met(server, current_lb_nodes):
...
+    return eff.on(got_children_with_stats)
+
+
+def converge_one(log, group_locks, tenant_id, group_id, version,
```
manishtomar: `converge_group` or `converge_one_group`?

otter/tap/api.py
radix: @manishtomar: The Converger does create the partitioner internally (note that a factory is passed in, not a partitioner instance). If we passed in all the arguments instead, and later changed the implementation to not use a partitioner, we'd still need to change the __init__ signature to get rid of the useless arguments that are specific to ZK partitioning.

otter/convergence/service.py
```diff
@@ -56,99 +62,363 @@ def all_met(server, current_lb_nodes):
...
+                    desired=desired_group_state, active=active)
+            eff = _update_active(scaling_group, active)
+            eff = eff.on(lambda _: steps_to_effect(steps))
+            return eff
+        return time_eff.on(got_time)
+
+    return Effect(TenantScope(
+        parallel([sg_eff, gather_eff]).on(
+            success=got_all_data,
+            error=catch(FirstError, lambda fe: six.reraise(*fe[1].exc_info))),
```
manishtomar: Should `six` be added to `requirements.txt`?

radix: @manishtomar - thanks, I've added six. It was already being depended upon transitively, but it's good to be explicit.

otter/tap/api.py
```diff
@@ -275,10 +276,12 @@ def on_client_ready(_):
                 stop=partial(call_after_supervisor,
                              kz_client.stop, supervisor)))
 
-            # setup converger service
-            converger_service = Converger(reactor, kz_client, dispatcher)
+            # setup ConvergenceStarter service
+            converger_service = ConvergenceStarter(dispatcher)
```
manishtomar: Non-blocking: It looks like `ConvergenceStarter` need not be a service. Can be a class that encapsulates `dispatcher`. In fact, I wonder if its possible to have `start_convergence` as a function that takes `dispatcher` as first argument and this code just `partial`s it with dispatcher.

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,415 @@
...
+        self.assertEqual(
+            result,
+            [{'tenant_id': '00', 'group_id': 'gr1', 'version': 0},
+             {'tenant_id': '00', 'group_id': 'gr2', 'version': 3}])
+
+
+def _get_dispatcher():
+    return ComposedDispatcher([
+        TypeDispatcher({
+            ParallelEffects: perform_parallel_async,
```
manishtomar: This is pretty useful. It will be nice if this `TypeDispatcher` can be there in `effect.async` for test purposes.

otter/convergence/service.py
radix: Yeah, you make a good point. I'll add a comment noting the inefficiency.

otter/test/convergence/test_service.py
cyli: As @radix points out, #3/#4 are invalid. #3 is poorly described and a duplicate of #2, and the purpose of them was mostly to demonstrate that a call coming in the middle of another one will not interfere with the original call.  which isn't really necessary.

radix:

otter/test/convergence/test_service.py
```diff
@@ -152,34 +505,63 @@ def test_success(self):
...
+            self._get_dispatcher()])
+
+        # And make sure that exception isn't wrapped in FirstError.
+        e = self.assertRaises(
+            RuntimeError,
+            sync_perform, dispatcher, tscope_eff.intent.effect)
+        self.assertEqual(str(e), 'foo')
+
+
+def raise_(e):
```
manishtomar: +1. Pretty useful. This should be in standard library. Please add it in utils.py. (Non-blocking)

otter/test/convergence/test_service.py
```diff
@@ -1,95 +1,432 @@
...
-        self.state = GroupState('tenant-id', 'group-id', 'group-name',
-                                {}, {}, None, {}, False)
-        self.group = mock_group(self.state, 'tenant-id', 'group-id')
+        self.tenant_id = 'tenant-id'
+        self.group_id = 'group-id'
+        self.state = GroupState(self.tenant_id, self.group_id, 'group-name',
+                                {}, {}, None, {}, False, desired=2)
+        self.group = mock_group(self.state, self.tenant_id, self.group_id)
         self.lc = {'args': {'server': {'name': 'foo'}, 'loadBalancers': []}}
         self.desired_lbs = s(CLBDescription(lb_id='23', port=80))
```
manishtomar: Non-blocking thought: I don't like `s` function. It pollutes the namespace with very probable local variable.

radix:
Thanks for the reviews, everyone. I'm merging this.

otter/test/convergence/test_service.py
radix: Already done in my next branch. thanks. :)

radix:
Oh, for the record, since no one has actually commented with "LGTM": 

<manisht> radix: i've looked at dirty PR test_service.py. they look great :simple_smile: . i am still looking. however, feel free to merge if you have the other dependent PR completed

PR #1194, opened 2015-03-12
delete server step retries

### LINKED ISSUE #1180: DeleteServer should have its own retry loop
Since 204 from Nova on a delete call can't be trusted, we should implement a loop in the DeleteServer step that checks if the server has actually moved into DELETING, and if not, try again.

It shouldn't be allowed to be infinite, maybe 30-60 seconds, and if we never see DELETING then we should return a RETRY.

manishtomar:
Closes #1180

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -191,6 +191,8 @@ def delete_and_verify(server_id):
     """
 
     def check_task_state((resp, json_blob)):
+        if resp.code == 404:
+            return
         server_details = json_blob['server']
         is_deleting = server_details.get("OS-EXT-STS:task_state", "")
         if is_deleting.strip().lower() != "deleting":
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -197,19 +197,18 @@ def delete_and_verify(server_id):
             raise UnexpectedServerStatus(server_id, is_deleting, "deleting")
 
     def verify((_type, error, traceback)):
-        if _type is not APIError or error.code != 204:
-            raise error
+        if error.code != 204:
+            raise _type, error, traceback
         ver_eff = service_request(
             ServiceType.CLOUD_SERVERS, 'GET',
             append_segments('servers', server_id, 'details'),
-            success_pred=has_code(200))
-        return ver_eff.on(check_task_state).on(
-            error=catch(APIError, lambda (t, e, _): e.code == 404))
+            success_pred=has_code(200, 404))
+        return ver_eff.on(check_task_state)
 
     return service_request(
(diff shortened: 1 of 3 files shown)


otter/convergence/steps.py
```diff
@@ -161,6 +163,55 @@ def report_failure(result):
...
+            ServiceType.CLOUD_SERVERS, 'GET',
+            append_segments('servers', server_id, 'details'),
+            success_pred=has_code(200))
+        return ver_eff.on(check_task_state).on(
+            error=catch(APIError, lambda (t, e, _): e.code == 404))
+
+    return service_request(
+        ServiceType.CLOUD_SERVERS, 'DELETE',
+        append_segments('servers', server_id),
+        success_pred=has_code(404)).on(error=verify)
```
radix: You can use .on(error=catch(APIError, verify)) here to guard against non-APIError. You can update requirements.txt to effect==0.1a14 to get this function (don't worry about conflicts with other PRs).

otter/convergence/steps.py
```diff
@@ -161,6 +163,55 @@ def report_failure(result):
...
+
+    def check_task_state((resp, json_blob)):
+        server_details = json_blob['server']
+        is_deleting = server_details.get("OS-EXT-STS:task_state", "")
+        if is_deleting.strip().lower() != "deleting":
+            raise UnexpectedServerStatus(server_id, is_deleting, "deleting")
+
+    def verify((_type, error, traceback)):
+        if _type is not APIError or error.code != 204:
+            raise error
```
radix: I guess even if you use catch() you'll still need this conditional, so I suggest changing this raise statement to `raise _type, error, traceback` so the traceback is maintained when we log it.

otter/convergence/steps.py
```diff
@@ -161,6 +163,55 @@ def report_failure(result):
...
+        ver_eff = service_request(
+            ServiceType.CLOUD_SERVERS, 'GET',
+            append_segments('servers', server_id, 'details'),
+            success_pred=has_code(200))
+        return ver_eff.on(check_task_state).on(
+            error=catch(APIError, lambda (t, e, _): e.code == 404))
```
radix: oh, now I see you're already using catch... nice :)

radix: Ok, this is weird. I don't think anything cares about a bool return value from this effect, so it seems you should actually be raising the exception if it's not a 404, not just returning False. I think the test you're missing is a non-404 APIError from verify.

manishtomar: Yes. Good catch. Thats a bug! It should raise error if its non 404. Otherwise the retry would not continue. Thanks!

radix:
LGTM :)

PR #1195, opened 2015-03-13
Don't spam logs with buckets

radix:
This makes it so buckets are only logged when they change.

### DIFF
--- a/otter/test/util/test_zkpartitioner.py
+++ b/otter/test/util/test_zkpartitioner.py
@@ -109,7 +109,7 @@ class PartitionerTests(SynchronousTestCase):
         self.partitioner.startService()
         self.log.msg.assert_called_once_with(
             'Got buckets {buckets}',
-            buckets=[2, 3], path=self.path,
+            buckets=[2, 3], old_buckets=[], path=self.path,
             otter_msg_type='partition-acquired')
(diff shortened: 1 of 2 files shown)


otter/test/util/test_zkpartitioner.py
```diff
@@ -122,12 +122,44 @@ def test_repeat(self):
...
+        self.kz_partitioner.__iter__.return_value = [2, 3]
+        self.partitioner.startService()
+        self.clock.advance(10)
+        self.clock.advance(10)
+        self.log.msg.assert_called_once_with(
+            'Got buckets {buckets}',
+            buckets=[2, 3], old_buckets=[], path=self.path,
+            otter_msg_type='partition-acquired')
+
+    def test_no_log_spam(self):
```
cyli: These two test names are the same, which means that one will will not run.  Might be useful to differentiate the docstrings too?  One is testing that if there is no change, there is only the initial bucket allocation log, and the other testing that when there are changes after the initial bucket allocation, it is logged.

cyli:
Aside from the test name and docstring change, LGTM!  Thank you for reducing logs!

radix:
yeah, sorry for the sloppy test names/docstrings. thanks very much for the review @cyli :) I'll merge once travis runs with the new revision.

PR #1196, opened 2015-03-13
Add a throttling observer wrapper to drastically reduce the amount of log spam

### LINKED ISSUE #1193: Filter kazoo pings out of logs
As per discussion with @radix and @manishtomar, just log errors, and maybe for pings log rollups?  We can use [yunomi](https://github.com/dreid/yunomi), and maybe do this as a meter or counter metric?

A very large portion of our logs are the ping logs, and reducing it would probably make storing operationally easier.

radix:
This fixes #1193.

This introduces a general throttling wrapper and adds throttling for three messages:

- sending ping to ZK
- received successful ping response from ZK
- fetching events from cassandra

It will wait for 50 events before letting a message through (which is actually not very long)

### DIFF
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -5,6 +5,8 @@ import json
 import time
 from datetime import datetime
 
+from pyrsistent import pmap
+
 from twisted.python.failure import Failure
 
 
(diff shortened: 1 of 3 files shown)


cyli:
Yay, logs are cleaner!  :)  thanks!  LGTM

PR #1197, opened 2015-03-16
error details are logged

manishtomar:
If exception object has `details` property, it is extracted and logged as `error_details`

### DIFF
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -261,7 +261,7 @@ def ErrorFormattingWrapper(observer):
                 message = repr(excp)
                 event['traceback'] = event['failure'].getTraceback()
                 event['exception_type'] = excp.__class__.__name__
-                details = getattr(excp, 'details', None)
+                details = serialize_exception(excp)
                 if details is not None:
                     event['error_details'] = details
 
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -9,6 +9,8 @@ from pyrsistent import pmap
 
 from singledispatch import singledispatch
 
+from toolz.functoolz import identity
+
 from twisted.python.failure import Failure
 
 
@@ -247,12 +249,9 @@ def serialize_to_jsonable(obj):
     return None
 
 
-@serialize_to_jsonable.register(basestring)
-def serialize_str(s):
-    """
-    Serialize string
-    """
-    return s
+# serialize base types
+for t in (list, tuple, basestring, int, float, long):
+    serialize_to_jsonable.register(t, identity)
 
 
 def ErrorFormattingWrapper(observer):
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -272,7 +272,7 @@ def ErrorFormattingWrapper(observer):
                 event['traceback'] = event['failure'].getTraceback()
                 event['exception_type'] = excp.__class__.__name__
                 details = serialize_to_jsonable(excp)
-                if details != repr(excp):
+                if details != message:
                     event['error_details'] = details
 
             if 'why' in event and event['why']:
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -261,6 +261,9 @@ def ErrorFormattingWrapper(observer):
                 message = repr(excp)
                 event['traceback'] = event['failure'].getTraceback()
                 event['exception_type'] = excp.__class__.__name__
+                details = getattr(excp, 'details', None)
+                if details is not None:
+                    event['error_details'] = details
 
             if 'why' in event and event['why']:
                 message = '{0}: {1}'.format(event['why'], message)
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -9,8 +9,6 @@ from pyrsistent import pmap
 
 from singledispatch import singledispatch
 
-from toolz.functoolz import identity
-
 from twisted.python.failure import Failure
 
 
@@ -246,12 +244,7 @@ def serialize_to_jsonable(obj):
     """
     Serialize any object to a JSONable form
     """
-    return None
-
-
-# serialize base types
-for t in (list, tuple, basestring, int, float, long):
-    serialize_to_jsonable.register(t, identity)
+    return repr(obj)
 
 
 def ErrorFormattingWrapper(observer):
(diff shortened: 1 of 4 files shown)


radix:
I'll suggest a slightly more general idea:

Have a separate map of Exception types to functions, where the functions are `exception -> json-serializable`.

Then we can put an entry for `UpstreamError` in this mapping, which is associated with `lambda e: e.details`.

Then we can also have a default for all exceptions that is just `lambda e: e.args`.

Bonus points for using the [singledispatch](https://pypi.python.org/pypi/singledispatch) module for this :)


Here's an example:

```
from singledispatch import singledispatch

@singledispatch
def serialize_exception(exc):
    raise TypeError('tried to serialize a non-exception')

@serialize_exception.register(Exception)
def _(exc):
    return exc.args

# elsewhere...
@serialize_exception.register(UpstreamError)
def _(exc):
    return exc.details
```

Then you can just call serialize_exception(exc) on the exception to get the error details in your code.

manishtomar:
As part of spec based logging, we want to know types of all fields being logged. For that we need to log error information in one field (which is `"error_details"` here). Will the json serializable returned from different errors be logged in this same field? As of now, I've tried to encapsulate type information [here](https://github.com/rackerlabs/otter/blob/formal-logging/otter/log/spec.py#L33) but with different forms of json serializable returned it might get complicated to represent that information.

radix:
@manishtomar If you're asking about my response, I'm just suggesting this as the way to _get_ the details, you can serialize them exactly the same way you already are (putting them in the `error_details` field).

manishtomar:
@radix: Thanks for the confirmation. Sorry I should've be clearer earlier. However, I am still a little confused about generic `json-serializable` part. If you agree that it is always going to be `dict` with known fields, then I am fine with this change :)

radix:
@manishtomar Sure, that's fine, though I think we should allow additional fields beyond what the schema requires so we can still have structured ad-hoc data.

manishtomar:
@radix With string reason in #1229, do you think we should instead have `serialize_to_dict` function that also overrides for `str`?

radix:
@manishtomar yeah, that sounds like a pretty good option. It sounds more and more like the log serialization stuff in the forthcoming Twisted logging code.

I'm still not sure exactly how we should log the stuff in `reasons` -- right now we're just passing the whole structure (containing all the steps and all the reasons) in a single message, so any logger code that specifically tries to handle the first argument to `log.err` (like what you have in this PR) isn't going to take effect for them anyway.

From what I can discern, ElasticSearch supports objects nested inside arrays just fine, and allows pretty complex queries on fields of objects within lists, but Kibana doesn't expose this because they can't figure out a syntax to do it. https://github.com/elastic/kibana/issues/998

So we may end up `flattening` out each reason into its own message, with enough additional columns to allow correlating it with the other operations going on in the same convergence iterations.

Sorry, this is a bit of an off-topic rant, but I needed to make sure I recorded this for posterity :)

otter/log/formatters.py
```diff
@@ -237,6 +239,22 @@ def audit_log_formatter(eventDict, timestamp, hostname):
     return audit_log_params
 
 
+@singledispatch
+def serialize_to_jsonable(obj):
+    """
+    Serialize any object to a JSONable form
+    """
+    return None
```
radix: This should probably be `return repr(obj)`

otter/log/formatters.py
```diff
@@ -237,6 +239,22 @@ def audit_log_formatter(eventDict, timestamp, hostname):
...
+
+
+@serialize_to_jsonable.register(basestring)
+def serialize_str(s):
+    """
+    Serialize string
+    """
+    return s
```
radix: Could you register the identity function for all of lists, tuples, strings, dicts, ints, floats, and longs?

```
for t in list, tuple, basestring, int, float, long:
    serialize_to_jsonable.register(t, identity)
```

(fwiw, I prefer this to a `map` or a listcomp to emphasize that it has side-effects)

radix: oh, and we should do this for all of the pyrsistent data structures, but not necessary in this branch if you don't want to :)

radix: agh, and I just realized the containers need to be recursively call serialize_to_jsonable on their contents. okay, you can leave that off for now :)

otter/log/formatters.py
manishtomar: So, I thought of that but left it due to formatter [checking](https://github.com/rackerlabs/otter/pull/1197/files#diff-193b770a377724dd1647e589b8022d0dR283) it. Since exception message is already taken care I was not sure if it was a good idea to add it in `error_details` also.

otter/log/formatters.py
```diff
@@ -237,6 +241,19 @@ def audit_log_formatter(eventDict, timestamp, hostname):
...
+def serialize_to_jsonable(obj):
+    """
+    Serialize any object to a JSONable form
+    """
+    return None
+
+
+# serialize base types
+for t in (list, tuple, basestring, int, float, long):
+    serialize_to_jsonable.register(t, identity)
```
radix: I guess you missed some of my later comments :) We can't just use `identity` for container types, since they may contain non-jsonable objects, so we actually need to recurse on their contents.

I think it's fine if you want to skip this entirely for this PR, since we're only using this function for exceptions anyway.

manishtomar: Oh yeah. Sorry I've not been concentrating. I'll take this.

otter/log/formatters.py
```diff
@@ -261,6 +271,9 @@ def error_formatting_observer(event):
                 message = repr(excp)
                 event['traceback'] = event['failure'].getTraceback()
                 event['exception_type'] = excp.__class__.__name__
+                details = serialize_to_jsonable(excp)
+                if details != repr(excp):
```
radix: This seems weird, why not just always populate `error_details` with this value?

radix:
Other than my one comment this LGTM :)

otter/log/formatters.py
manishtomar: This is because exception repr is already stored in message. I was thinking of not duplicating it in `error_details`. The intention is the give extra details only when it is there. I don't feel that strongly about it. If you think its ok to put same info, I can change it.

radix: I see now. I only suggest changing the "if details != repr(excp):" to "if details != message" then. Sorry & thanks!

Issue #1192, opened 2015-03-12
Race condition between `delete_server` and convergence when deleting a group's lock znode


cyli:
See this error now when attempting to delete an empty scaling group:

```
Request failed: Unhandled Error: NotEmptyError((), {})", "otter_facility": "otter.rest.groups.delete_scaling_group", "request_status": "failed", "scaling_group_id": "1abcd010-5226-4cd0-b126-54ad317f1b08", "tenant_id": "000000", "traceback": "Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 524, in __bootstrap
    self.__bootstrap_inner()
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
--- <exception caught here> ---
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/threadpool.py", line 196, in _worker
    result = context.call(ctx, function, *args, **kwargs)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/kazoo/client.py", line 1243, in delete
    return self._delete_recursive(path)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/kazoo/client.py", line 1278, in _delete_recursive
    self.delete(path)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/kazoo/client.py", line 1245, in delete
    return self.delete_async(path, version).get()
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/kazoo/handlers/threading.py", line 102, in get
    raise self._exception
kazoo.exceptions.NotEmptyError: ((), {})
```

(Seen as a failure in the functional tests, `scaling_group.test_scaling_group_negative.ScalingGroupNegative.test_update_group_after_deletion`, line 239, when it's attempting to delete the empty group, before it attempts to update the group)

cyli:
[...]

PR #1199, opened 2015-03-16
Ignore znode deletion failures when we delete the lock path for a group

cyli:
Addresses the immediate failure of #1192.

Not sure if we want to leave that issue open because we also want to do the reaping.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -26,6 +26,7 @@ from silverberg.client import ConsistencyLevel
 from toolz.dicttoolz import keymap
 
 from twisted.internet import defer
+from twisted.python.failure import Failure
 
 from zope.interface import implementer
 
@@ -1215,8 +1216,17 @@ class CassScalingGroup(object):
         def _delete_lock_znode(result):
             d = self.kz_client.delete(LOCK_PATH + '/' + self.uuid,
                                       recursive=True)
-            # If we fail to delete the node, ignore it.
-            d.addBoth(lambda _: result)
+
+            def return_result_ignore_errors(zk_delete_result):
+                # If we fail to delete the node, log and otherwise ignore it.
+                if isinstance(zk_delete_result, Failure):
+                    self.log.msg(
+                        "Error cleaning up lock path (when deleting group)",
+                        exc=zk_delete_result.value)
+
+                return result
+
+            d.addBoth(return_result_ignore_errors)
             return d
 
         lock = self.kz_client.Lock(LOCK_PATH + '/' + self.uuid)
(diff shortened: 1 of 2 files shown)


cyli:
jenkins please run unit tests

otter/models/cass.py
```diff
@@ -1215,7 +1215,8 @@ def _delete_group():
         def _delete_lock_znode(result):
             d = self.kz_client.delete(LOCK_PATH + '/' + self.uuid,
                                       recursive=True)
-            d.addCallback(lambda _: result)
+            # If we fail to delete the node, ignore it.
+            d.addBoth(lambda _: result)
```
manishtomar: It will be nice to log error before ignoring it.

cyli: Good point, adding.

cyli:
jenkins please run unit tests

otter/models/cass.py
```diff
@@ -1215,7 +1216,17 @@ def _delete_group():
...
+            def return_result_ignore_errors(zk_delete_result):
+                # If we fail to delete the node, log and otherwise ignore it.
+                if isinstance(zk_delete_result, Failure):
+                    self.log.msg(
+                        "Error cleaning up lock path (when deleting group)",
+                        exc=zk_delete_result.value)
+
+                return result
+
+            d.addBoth(return_result_ignore_errors)
```
radix: I'd prefer 

```
d.addErrback(log.msg, "Error cleaning up lock path (when deleting group)")
d.addCallback(lambda _: result)
```

also, could you add an `otter_msg_type`, maybe something like "ignored-delete-lock-error"?

cyli: We have to return result in either case, right?  Although the result is technically `None`.

radix: oh, my d.addErrback line is wrong. I should have written `d.addErrback(lambda f: log.msg("Error...", exc=f.value)`.

cyli:
jenkins please run unit tests

cyli:
jenkins please run unit tests

otter/models/cass.py
radix: @cyli I'm not sure what you mean. The code I gave will always result in `result` -- the `addErrback` of `log.msg` will switch things back to the callback chain, so that last callback will always run.

cyli: Yes, you're right, apologies, forgot about that.  I basically went with what you suggested anyway, but just returned None at the end, since apparently that's what the interface says is the success result anyway

radix:
LGTM :)

PR #1200, opened 2015-03-16
return the result of kazoo delete from DeleteNode

radix:
super trivial!

This means that 1) we'll know when a deletion is actually complete and 2) we'll get errors from it.


radix:
jenkins please run unit tests

cyli:
LGTM if tests pass

radix:
jenkins please run unit tests

PR #1205, opened 2015-03-17
Script to run cloudcafe and move its logs to a particular directory

cyli:
Hoping to use this to run CI in a way that we can archive logs.


lvh:
LGTM! Also yeah whatever coveralls

Issue #1135, opened 2015-02-26
Replicate or falsify this bug.


sam-falvo:
```
"exc": "CannotExecutePolicyError('Cannot execute scaling policy cfc5a6d2-d593-4483-ba42-881584deed19 for group 27c1ca5a-0999-4d0f-bf61-842541b551a5 for tenant 10029579: No change in servers',)",

How to setup this up:
1.) Create AutoScale group with single Cloud Load Balancer and two AutoScale WebHooks (scaleup by 1 and scaledown by 1).
2.) Scale the Group up by 2
3.) Change the Launch Configuration to add the second Cloud Load Balancer
4.) Attempt to execute either AutoScale WebHook

Source ticket we found the problem with:
https://lon.cloudcontrol.rackspacecloud.com/customer/10029579/taas/150225-ord-0000477/edit
adf

Customer's Production Group:
https://lon.autoscale.api.rackspacecloud.com/v1.0/10029579/groups/27c1ca5a-0999-4d0f-bf61-842541b551a5


Workaround:
2a.)Before editing the LaunchConfiguration, scale the AutoScale Group down to 0
4a.)Scale back to where you want


The problem is only found via LogStash/Kibana.


My Test Group that you can use and break:
https://lon.autoscale.api.rackspacecloud.com/v1.0/10002904/groups/1a54203c-2c87-43b4-af57-b0cd9886af33



Please ReplyAll so that my team can respond as needed, if I am out or travelling.

-Brian Metzler
Stratus Cloud Operations
```

sam-falvo:
After breaking the ticket into engineering tasks, closing ticket.

PR #1209, opened 2015-03-17
Completes reproduction of bug 1135

sam-falvo:
It's done.  It works.  Bug was _not_ reproduced, as per @cyli's findings.  But at least we have test automation for it!  And I have a general feel for how to write these kinds of test in the future.  Given the current state of the library, if I were to write a new test of this ilk, I'd wager I could complete it in about 4 to 6 hours now (including development testing time).

About the branch name -- no idea.  Haven't a clue.  But it was already in a state ready for working on bug #1135, so I figured I'd not waste the branch, and changing branch names after an initial push is, shall we say, invasive.  Might be for something I was thinking of working on during otterweek but never got to?  I don't even remember when I created this branch.

### DIFF
--- a/otter/integration/tests/test_scaling.py
+++ b/otter/integration/tests/test_scaling.py
@@ -202,7 +202,7 @@ class TestScaling(unittest.TestCase):
         """
         rcs = TestResources()
 
-        def create_1st_load_balancer(self):
+        def create_1st_load_balancer():
             """First, we authenticate and create a single load balancer."""
             self.clb1 = cloud_load_balancer.CloudLoadBalancer(pool=self.pool)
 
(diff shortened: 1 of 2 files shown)


stretchservers:

otter/integration/tests/test_scaling.py
```diff
@@ -202,9 +202,23 @@ def test_policy_execution_after_adding_clb(self):
         """
         rcs = TestResources()
 
-        self.clb1 = cloud_load_balancer.CloudLoadBalancer(pool=self.pool)
+        def create_1st_load_balancer(self):
```
cyli: This probably doesn't need to take "self", since it's closed over by the instance method.

cyli:
Aside from the `self` issue above, which is causing errors because `create_1st_load_balancer` is not being called with any arguments, LGTM :)

PR #1210, opened 2015-03-18
AddNodesToCLB fails on known conditions only

manishtomar:
Earlier it was failing on unknown errors. This is incorrect since we should retry convergence on intermittent network failures. Now, it fails on known conditions only and return retry on any other error

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -367,13 +367,8 @@ class AddNodesToCLB(object):
                     return StepResult.FAILURE, [error]
             return StepResult.RETRY, [error]
 
-        def report_retry((_t, error, tr)):
-            return StepResult.RETRY, [error]
-
-        return eff.on(
-            success=report_success,
-            error=catch(APIError, report_api_failure)).on(
-                error=report_retry)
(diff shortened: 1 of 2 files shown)


radix:
in my convergence cycling branch, I changed `steps_to_effect` to turn any unknown error into a RETRY, that way we wouldn't have to handle this for every step. what do you think?

manishtomar:
How does converger know which errors are known and unknown? I thought that it will use only step statuses to make decisions and error information will be for logging purpose. If converger is able to make decisions about the kind of error then steps wouldn't need to return status right? They can always raise errors.

radix:
@manishtomar converger is only using step status.

radix:
@manishtomar this is what I mean, the change in effecting.py: https://github.com/rackerlabs/otter/compare/dirty-groups-3...repeat-convergence#diff-807b0b5ab96b01e953fae93a368cbd8f

radix:
after talking on slack, conclusion is: this change is good, but you don't need the `report_retry` errback, since that can be generalized up a level.

radix:
LGTM :) +1

Issue #1189, opened 2015-03-11
Parse nova/parse CLB errors in steps


cyli:
These should be functions that parse out specific Nova/CLB errors raise them with the relevant info, or reraise the original error.

Then the error handlers can do something with them or not, as desired.

See https://github.com/rackerlabs/otter/pull/1175#issuecomment-78183533

radix:
I'm going to put this in the Servers milestone, though it also applies to the CLB milestone too.

manishtomar:
@radix / @cyli : We need a ticket that decides about the specific error type like `CLBPendingUpdate` / defaulting to `UpstreamError` like error. Can this be that ticket? If so, then I will close #1115 and refer it to this. I originally intended that ticket to be there until we start logging errors or decide how to log them.

cyli:
@manishtomar that sounds good :)

cyli:
Wondering if we should just have 1 error for Nova, like NovaLaunchConfigError or something, that way we don't have to parse the 400 messages.  Otherwise:

-  NovaNetworkInvalidError:  (`Networks (00000000-0000-0000-0000-000000000000) not allowed`, `Networks (00000000-0000-0000-0000-000000000000,11111111-1111-1111-1111-111111111111) required but missing`, and `Exactly 1 isolated network(s) must be attached`)
- NovaQuotaExceededError: (all the quota messages: for RAM, for instances, for onmetal...)
- NovaFlavorError: (invalid flavor, flavor disk too small)
- NovaImageError: (invalid image, image has automatic disk size disabled, image not active)
- NovaSSHKeyError: (invalid key name)
- NovaDiskConfigError: `OS-DCF:diskConfig must be either 'MANUAL' or 'AUTO'.` - we can check for this one in our validation though, and eliminate this as a possible error.
- NovaLaunchConfigError: any 400 from Nova with a message we haven't seen before

cyli:
For CLB:

- CLBDuplicateNodesError - when trying to add a node that already exists to a CLB (not user-visible, we will deal with this one)
- CLBPendingUpdateError - when a CLB is in PENDING UPDATE state and we need to wait a bit (not user-visible - we will deal with this)
- CLBDeletedError - when a CLB is in PENDING DELETE or is deleted - user surfaced error
- CLBNodeRemovedError - when a node no longer exists on a CLB (not user-visible, we can deal with this one I think)
[...]

PR #1211, opened 2015-03-18
General mechanism to optionally parse APIErrors by service type

cyli:
This allows passing of a dictionary of parsers by service type to `concretize_service_request`.

If no parsers are passed, no parsing happens, but this should make it easier to incrementally implement parsers, etc, as part of #1189.

This also swaps the order in which JSON parsing/success predicate testing happens, since although on a successful request the body should be JSON, in error cases it's very possible the body will not be something JSON parseable (maybe a repose error, or some proxy error that returns HTML, etc.).

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -13,6 +13,8 @@ from effect import (
     perform,
     sync_performer)
 
+import six
+
 from otter.auth import Authenticate, InvalidateToken, public_endpoint_url
 from otter.util.http import APIError
 from otter.util.http import headers as otter_headers
@@ -154,9 +156,16 @@ def _add_service_error_parsing(parser, request_func):
     if parser is None:
         return request_func
 
+    def call_parser(excinfo):
+        # ensures that if the parser doesn't raise another exception,
+        # the original exception is raised
+        parser(excinfo)
+        six.reraise(*excinfo)
+
     @wraps(request_func)
     def request(*args, **kwargs):
-        return request_func(*args, **kwargs).on(error=catch(APIError, parser))
+        return request_func(*args, **kwargs).on(
+            error=catch(APIError, call_parser))
 
     return request
 
@@ -176,6 +185,10 @@ def concretize_service_request(
     :param BoundLog log: info about requests will be logged to this.
     :param dict service_configs: As returned by
         :func:`otter.constants.get_service_configs`.
+    :param dict service_error_parsers: A mapping of service types to error
+        parsers, which should accept the excinfo tuple, and raise another
+        error (or the original) in its place.  If it doesn't raise an
+        error, the original error is re-raised.
     """
     if service_error_parsers is None:
         service_error_parsers = {}
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -169,10 +169,10 @@ def concretize_service_request(
         else:
             request_ = add_bind_service(
                 catalog, service_name, region, log, request_)
-        if service_request.json_response:
-            request_ = add_json_response(request_)
         request_ = add_error_handling(
             service_request.success_pred, request_)
+        if service_request.json_response:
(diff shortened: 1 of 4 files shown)


otter/cloud_client.py
```diff
@@ -169,10 +212,12 @@ def got_auth((token, catalog)):
         else:
             request_ = add_bind_service(
                 catalog, service_name, region, log, request_)
-        if service_request.json_response:
-            request_ = add_json_response(request_)
         request_ = add_error_handling(
             service_request.success_pred, request_)
+        if service_request.json_response:
+            request_ = add_json_response(request_)
```
manishtomar: Unfortunately, due to this clients expecting JSON error will get text error. I realize you've done this because errors can be non-JSON. I feel it is better if `add_json_response` does not error when parsing, instead returns text. This way the caller can handle JSON or text. I am not sure how good an idea this is. @radix: I remember raising this concern when you were implementing `add_json_response`. What are your thoughts?

cyli: Hm... I changed the CLB ones which expect JSON to expect text.  I understand what you're saying though, although as an amendment to your proposal I think it should raise an error and return text, maybe?  Otherwise the callers never know whether they're getting text or JSON, and will have to test every time.

otter/cloud_client.py
```diff
@@ -135,10 +143,38 @@ def __init__(self, effect, tenant_id):
...
+
+    return request
+
+
 def concretize_service_request(
         authenticator, log, service_configs,
         tenant_id,
-        service_request):
+        service_request,
+        service_error_parsers=None):
```
manishtomar: I realize we discussed this on slack but passing parsers in this function seems like over-engineering to me. The intent here is to gradually provide specific exceptions so that existing callers do not break but eventually they would handle exceptions. So,
1. I don't see use-case to provide custom error parsers. Maybe I am missing something?
2. The new callers will now need to have (global?) parser dict and pass it here. 
I think a providing `specific_exceptions=True` argument should do the job.

otter/cloud_client.py
manishtomar: Yeah. I was also thinking that callers will have to check everytime. That will be a pain. I wonder if providing a `json_response_text_fallback=True` argument would be useful. Maybe not. I am not really convinced either way. So, lets leave it like this :)

otter/cloud_client.py
cyli: The callers should definitely not provide dicts.  This was mostly a mechanism to provide incremental functionality.  I wanted to not touch the steps at all until the parsers were done, and also implement parsers one by one.  

This was mostly for easy splitting of PRs and also test injection - I wanted to make sure the mechanism by which the parsers were called were correct before actually implementing a parser.  I had figured that eventually I'd make the default be the correct dict.

I'm fine with getting rid of it later - how do you think we should proceed?  Should I just merge my CLB parser into this PR, and use that to test that if there is a parser for a service type, errors are parsed, and if not, errors are not parsed?

cyli:
After further discussion with @manishtomar, am just going to implement the boolean behavior he suggested (add a boolean to ServiceRequest as to whether or not to parse APIErrors, defaulting to No until all the steps are converted to utilize these errors).  Putting this into in-progress.

lvh:
LGTM; thanks for working on this! Sorry that people think it's okay to return inconsistent data formats from APIs. `Accept` headers should be a real thing...

manishtomar:
@lvh : Sorry this was not intended to be merged. @cyli had plans to change it which is why she left "in progress".

lvh:
@manishtomar She did not leave it "in progress"; she said she was putting it "in progress", put it "in progress", made some commits, and then removed the "in progress" label.

manishtomar:
@lvh Oh yeah. Sorry didn't see commits. Please ignore my comment :)

PR #1212, opened 2015-03-18
Re-auth on every create server and every delete server

cyli:
Re-auth (still using the cache, so by doing this re-auth ensures that the token is at most 5 minutes old) automatically every time we attempt to create a server or delete a server).

A job goes thusly now:
1.  Supervisor auths once for the batch of servers to be created, to make sure we have a token cached.
2.  When the job creates a server, it auths again.  The first time, it just gets the cached token from when the supervisor auth-ed.
3.  We poll Nova waiting for the server to go active, using the auth token we got right before we created the server.
3.  If the server goes into error, the job re-auths again and issues a delete.
4.  We re-auth yet again (will probably be a cached token from the delete), and retry creating a server.
5.  Rinse and repeat.

This is to deal with 401 errors from Nova when it takes a long time to build/go into error, and we retry.

### DIFF
--- a/otter/supervisor.py
+++ b/otter/supervisor.py
@@ -108,21 +108,25 @@ class SupervisorService(object, Service):
                                            self.service_configs)
         lb_region = config_value('regionOverrides.cloudLoadBalancers')
 
-        log.msg("Authenticating for tenant")
-        d = self.authenticator.authenticate_tenant(tenant_id, log=log)
+        def authenticate():
+            log.msg("Authenticating for tenant")
+            d = self.authenticator.authenticate_tenant(tenant_id, log=log)
 
-        def when_authenticated((auth_token, service_catalog)):
-            bag = RequestBag(
-                lb_region=lb_region or self.region,
-                region=self.region,
-                dispatcher=dispatcher,
-                tenant_id=tenant_id,
-                auth_token=auth_token,
-                service_catalog=service_catalog
-            )
-            return bag
+            def when_authenticated((auth_token, service_catalog)):
+                bag = RequestBag(
+                    lb_region=lb_region or self.region,
+                    region=self.region,
+                    dispatcher=dispatcher,
+                    tenant_id=tenant_id,
+                    auth_token=auth_token,
+                    service_catalog=service_catalog
+                )
+                bag.re_auth = authenticate
+                return bag
 
-        return d.addCallback(when_authenticated)
+            return d.addCallback(when_authenticated)
+
+        return authenticate()
 
     def execute_config(self, log, transaction_id, scaling_group, launch_config):
         """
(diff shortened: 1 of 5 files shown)


cyli:
Integration tests (minus the scheduler ones) pass on my dev vm.

cyli:
Jenkins please run integration tests

cyli:

otter/test/test_supervisor.py
```diff
@@ -56,12 +57,13 @@ def setUp(self):
         self.group.uuid = 'group-id'
         self.region = "ORD"
 
-        self.auth_token = 'auth-token'
+        self.auth_token = ('auth-token1', 'auth-token2', 'auth-token3')
+        self._auth_token = list(self.auth_token)
```
radix: I suggest naming these variables "self.auth_tokens" and "self.auth_token_queue" or something like that

otter/supervisor.py
```diff
@@ -108,23 +108,28 @@ def _get_request_bag(self, log, scaling_group):
...
+            def when_authenticated((auth_token, service_catalog)):
+                bag = RequestBag(
+                    lb_region=lb_region or self.region,
+                    region=self.region,
+                    dispatcher=dispatcher,
+                    tenant_id=tenant_id,
+                    auth_token=auth_token,
+                    service_catalog=service_catalog
+                )
+                bag.re_auth = authenticate
```
radix: Could you declare re_auth on RequestBag and then just pass it normally to the constructor instead of assigning it afterwards?

otter/test/worker/test_launch_server_v1.py
```diff
@@ -86,31 +88,48 @@
...
+        def new_bag(counter=0):
+            bag = RequestBag(
+                lb_region="DFW",
+                region="DFW",
+                dispatcher=mock.Mock(),
+                tenant_id=self.tenant_id,
+                auth_token="my-auth-token{0}".format(counter),
+                service_catalog=fake_service_catalog
+            )
+            bag.re_auth = partial(new_bag, counter + 1)
```
radix: same here re: passing re_auth to constructor

radix:
After my nitpicks, this LGTM and is good to merge.

However, I also want to comment: shouldn't we also reauth for add-to-LB as well? If we have a long build we can also have an issue if we're passing a stale token to CLB when trying to add the server.

I don't want to block this PR, or even suggest that we have to make this change in another PR soon, since we've already determined that the root cause for our recent problems was in Nova, but I want to get your thoughts on that.

manishtomar:
@radix Thats a good point. However, like you said, we don't want to block this PR. We can keep this in mind and implement it if there are still 401s in future.

radix:

radix:
I've addressed my own comments, and I'll merge now.

PR #1217, opened 2015-03-20
fix CLB node gathering to expect the correct format

radix:
listing LB nodes returns a JSON structure like {"nodes": [...]}, but the code was just expecting a list to be returned directly.


lvh:
Whoa. Good catch! Found this before pointing it at real data?

PR #1218, opened 2015-03-20
Steps with unexpected errors should result in a RETRY

radix:
This makes steps_to_effect wrap all the effects in an error handler that returns (StepResult.RETRY, [exception]).


lvh:
LGTM! I'm a little surprised, but happy to see that keeping steps.to.effect around as a separate function ended up paying off (a little but not too much).

PR #1219, opened 2015-03-20
repeat convergence if any steps return RETRY

radix:
This implements #844 and #956.

There are two changes here:

- execute_convergence now inspects the results of steps, and returns the most severe status found (FAILURE is more severe than RETRY is more severe than SUCCESS).
- converge_one_group now looks at the result of execute_convergence, and in the case of RETRY, simply doesn't delete the divergent flag from ZK, so convergence will be repeated.

These changes enable a lot more integration tests to run, but since I have to deal with a ton of lint in the existing tests, I'm separating that into another PR.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -81,6 +81,7 @@ def _update_active(scaling_group, active):
                                    modifier=update_group_state))
 
 
+@do
 def execute_convergence(tenant_id, group_id, log,
                         get_all_convergence_data=get_all_convergence_data,
                         plan=plan):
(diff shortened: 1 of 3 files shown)


radix:

manishtomar:
Woo! Excited to see the tests pass! :)

radix:

otter/convergence/service.py
```diff
@@ -1,4 +1,9 @@
-"""Converger service"""
+"""
+Converger service
+
+The top-level entry-points into this module are :obj:`ConvergenceStarter` and
+:obj:`Converger`.
+"""
```
lvh: :heart:

otter/convergence/service.py
```diff
@@ -299,6 +321,7 @@ def got_children_with_stats(children_with_stats):
     return eff.on(got_children_with_stats)
 
 
+@do
```
lvh: Man, seeing this in Python feels real weird. (That is not something you should address in this PR; I'm just saying that `@do` doesn't really help me when I first read this code even though I knew about both Haskell's `do` and `effect.do.do`.)

otter/convergence/service.py
```diff
@@ -310,11 +333,24 @@ def converge_one_group(log, group_locks, tenant_id, group_id, version,
...
     log = log.bind(tenant_id=tenant_id, group_id=group_id)
     eff = execute_convergence(tenant_id, group_id, log)
-    return group_locks(group_id, eff).on(
-        success=lambda r: delete_divergent_flag(log, tenant_id,
-                                                group_id, version),
-        error=partial(_cleanup_convergence_error, log, tenant_id,
-                      group_id, version))
+    try:
+        result = yield group_locks(group_id, eff)
+    except ConcurrentError:
```
lvh: This is supposed to be raised when the current Otter process is already converging, right?

otter/convergence/service.py
```diff
@@ -310,11 +333,24 @@ def converge_one_group(log, group_locks, tenant_id, group_id, version,
...
+        return
+    except Exception:
+        # We specifically don't clean up the dirty flag in the case of
+        # unexpected errors, so convergence will be retried.
+        log.err(None, 'converge-non-fatal-error')
+    else:
+        if result in (StepResult.FAILURE, StepResult.SUCCESS):
+            yield delete_divergent_flag(log, tenant_id, group_id, version)
+        # TODO: if result is FAILURE, put the group into ERROR state.
+        # https://github.com/rackerlabs/otter/issues/885
```
lvh: This reminds me that we need a good standard for linking TODOs to issues. Notably, you should be able to `ag` (or Github search, I suppose) your way through the codebase and find all related bits.

lvh: (Do you have any suggestions? Previously I thought I wanted `TODO: ... (#123)`, figured that was hard to grep for, then maybe `TODO #123: ...`, but I really like that this is a URL and therefore potentially clickable, also that it allows linking to other issues from e.g. other projects.)

lvh:
LGTM modulo comments, which are mostly details and don't really need addressing. This is surprisingly clear in what it does and where it's going; the comments with specific issue links are *super* helpful, and really helped me in review. Thanks a lot for working on this, and great job @radix :)

otter/convergence/service.py
radix: Yeah, it's raised by `non_concurrently`

otter/convergence/service.py
```diff
@@ -110,10 +116,26 @@ def got_time(now):
...
 
+    def got_step_results(steps, results):
+        order = [StepResult.FAILURE, StepResult.RETRY, StepResult.SUCCESS]
+        priority = sorted(results,
+                          key=lambda (status, reasons): order.index(status))
+        worst_status = priority[0][0]
+        log.msg('execute-convergence-results',
+                results=zip(steps, results),
+                worst_status=worst_status)
+        return worst_status
```
manishtomar: The docstring should be changed to reflect this return type.

otter/convergence/service.py
```diff
@@ -334,21 +377,6 @@ def converge_all_groups(log, group_locks, my_buckets, all_buckets,
     yield do_return(parallel(effs))
 
 
-def _cleanup_convergence_error(log, tenant_id, group_id, version, exc_info):
```
manishtomar: +1. This is better. I like this moved to `converge_one_group` function.

otter/convergence/service.py
```diff
@@ -77,7 +82,8 @@ def update_group_state(group, old_state):
 
 
 def execute_convergence(tenant_id, group_id, log,
-                        get_all_convergence_data=get_all_convergence_data):
+                        get_all_convergence_data=get_all_convergence_data,
+                        plan=plan):
```
manishtomar: Non-blocking nitpick: Have you considered using `do` on this function? It seems to be serial and has lot of callbacks in it. It might be easier to read with `do`.

otter/convergence/service.py
manishtomar: Nice! This is function is *much* more clear now :)

otter/convergence/service.py
radix: Good idea. It grew from a very simple function and I kept adding callbacks, now it definitely seems @do would be better.

radix:
jenkins please run unit tests

radix:
Okay, jenkins ran the unit tests + lint fine, so I'm gonna merge. thanks for the reviews!

PR #1222, opened 2015-03-20
get rid of a ton of log spam from converger

radix:
Two sources of spam
- converge-all-groups messages when there are no groups to converge
- buckets-acquired, which is purely a duplicate of what the Partitioner object logs already (but on a throttled basis)


manishtomar:
LGTM

radix:
Thanks @manishtomar!

Issue #1221, opened 2015-03-20
throttle server creation requests in worker code


radix:
Nova doesn't want us to hit them with multiple concurrent Create Server requests, so we need to make sure we're putting gaps (of about 1 second) in between each POST to /servers.

lvh:
Some boring issue tracker management points:

- I don't know if this changed during the creation of this ticket, but this was given to various upstream Compute folks as the place to track it *right now*, which is about worker
- There already was a ticket for doing this in convergence: https://github.com/rackerlabs/otter/issues/866, where the theory was posited that the stuff we will be doing for convergence is sufficient.

Would it be OK to make this ticket be about worker instead?

radix:
It didn't change, this ticket was always intended to be about converger, not the worker. I was thinking something more specific than #866 with this ticket, specifically so it doesn't require multiple convergence cycles.

Given that nova folks are tracking *this* ticket now, though, it's fine to repurpose it.

lvh:
OK, cool, thanks. Would you like me to create another ticket for convergence? (I don't think anything has changed since I closed #866; I still don't know if it will actually be necessary to implement more throttling once we have convergence.)

lvh:
Closed by #1224. Not yet deployed.

radix:
I've created #1225 to represent the original intent of this ticket.

lvh:
Awesome, thanks!

PR #1224, opened 2015-03-22
1 second delay during server creation

manishtomar:
This is done by limiting server creation to 1 server at a time and injecting 1 second delay from create server response.

### DIFF
--- a/otter/test/worker/test_launch_server_v1.py
+++ b/otter/test/worker/test_launch_server_v1.py
@@ -896,6 +896,7 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
         """
         super(ServerTests, self).setUp()
         self.log = mock_log()
+        self.clock = Clock()
         set_config_data(fake_config)
         self.addCleanup(set_config_data, {})
 
@@ -1136,7 +1137,12 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
         _treq = StubTreq([(req, resp)], [(resp, '{"server": "created"}')])
 
         d = create_server('http://url/', 'my-auth-token', {'some': 'stuff'},
-                          _treq=_treq)
+                          _treq=_treq, clock=self.clock)
+
+        # No result initially. Will be avail after 1 second due to
+        # injected delay
+        self.assertNoResult(d)
+        self.clock.advance(1)
 
         result = self.successResultOf(d)
         self.assertEqual(result, {"server": "created"})
@@ -1144,7 +1150,8 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
 
     def test_create_server_limits(self):
         """
-        create_server when called many times will post only 2 requests at a time
+        create_server when called many times will post only 1 request at
+        a time
         """
         deferreds = [Deferred() for i in range(3)]
         post_ds = deferreds[:]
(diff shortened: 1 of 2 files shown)


lvh:
This closes #1221, right?

otter/worker/launch_server_v1.py
```diff
@@ -308,6 +308,13 @@ def _check_server_created(f):
         d.addBoth(_check_results, f)
         return d
 
+    def _create_with_delay():
+        d = _treq.post(path, headers=headers(auth_token),
+                       data=json.dumps({'server': server_config}), log=log)
+        # Add 1 second delay to space 1 second between server creations
+        d.addCallback(delay, clock, 1)
```
lvh: It is a bit unfortunate that this gets the result, and then sits on it for a second. We need to space outgoing requests, but you can still asynchronously space results without having to wait on them here.

otter/test/worker/test_launch_server_v1.py
```diff
@@ -1156,24 +1163,37 @@ def test_create_server_limits(self):
...
         }
 
-        ret_ds = [create_server('http://url/', 'my-auth-token', server_config)
+        ret_ds = [create_server('http://url/', 'my-auth-token',
+                                server_config, clock=self.clock)
                   for i in range(3)]
 
-        # no result in any of them and only first 2 treq.post is called
+        # no result in any of them and only 1 treq.post is called
         [self.assertNoResult(d) for d in ret_ds]
```
lvh: I realize you didn't write this in this PR, but side-effectful LCs are super gross and we should probably turn that into a for loop :)

otter/worker/launch_server_v1.py
manishtomar: This is the easiest way I thought of throttling requests on a single node. Another way is to space out job creation in [supervisor](https://github.com/rackerlabs/otter/blob/master/otter/supervisor.py#L559). Is this what you were suggesting? This is probably simpler but doesn't ensure spacing when multiple policies (of possibly different groups) are getting executed at same time. I can understand if that is considered an over-optimization at this point.

lvh: I don't think that this should block this PR right now, FWIW :-)

There are two ways to do this:

- Supervisor, as you were suggesting. That would space out *jobs*, not server creations; not quite what Compute is asking for, and, in the problematic cases, not something that actually performs their experiment.
- Separate the concern of "one at a time" from "at most one per second". I.e., don't change the thing you're changing here at all; that fn creates a server, and it still creates a server (no added delay). Instead, Make the delay be about adding things to the `DeferredSemaphore` in the first place, by e.g. putting create server requests in a queue, and consuming them every *timeout* seconds.

manishtomar: That does sound much cleaner way to do it. It makes its intent very clear. I got stuck in current code base and thought from that angle :). As you suggested, this doesn't need to block this PR now. Instead the idea can be utilized in converger.

lvh:
LGTM in the sense that it does address #1221. I am concerned that with the number of servers we create (and considering the Hurst exponent in that process), the one second delay is going to cause some pretty severe degradation; but that's part of the exercise, I guess.

manishtomar:
This addresses #1221 in worker only. That ticket is for converger. I think we should keep that open until it is implemented in converger.

manishtomar:
jenkins please run integration tests.

lvh:
@manishtomar Argh; did that change? We definitely gave out that ticket to a bunch of other people who are waiting for it to happen in worker right now, so I'll make a separate ticket for doing it in convergence.

lvh:
Assuming the tests pass, this LGTM. I would like to see if we can get the "what does #1221 even mean" issue resolved before merging it, but that's not a blocker.

lvh:
Huh; where did the build info go?

lvh:
Never mind, Github's auto-update went a little wonky :)

PR #1229, opened 2015-03-23
Add meaningful `reasons` lists and get rid of redundant error handlers in steps

radix:
Few related changes

- CreateServer's success-case RETRY now returns `RETRY, ['waiting for server to become active']`, so we can distinguing between RETRY due to error and RETRY due to success.
- got rid of a few `report_failure` errbacks that were just handling arbitrary errors that were written before I added that logic to steps_to_effect.
- add exception to results list in CreateServer

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -144,7 +144,7 @@ class CreateServer(object):
                 if message is not None:
                     return StepResult.FAILURE, [message]
 
-            return StepResult.RETRY, [result[1]]
+            return StepResult.RETRY, [error]
 
         return eff.on(got_name).on(success=report_success,
                                    error=report_failure)
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -129,7 +129,7 @@ class CreateServer(object):
             a while, and we need to retry convergence to ensure it goes into
             active.
             """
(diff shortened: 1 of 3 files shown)


otter/util/http.py
```diff
@@ -159,6 +161,8 @@ def _segments(segments):
     return uri
 
 
+@attributes(['code', 'body',
```
manishtomar: Why is this needed now?

radix: I added it so I could easily compare instances of APIError in the unit tests

radix:
jenkins please run unit tests

otter/convergence/steps.py
```diff
@@ -144,7 +144,7 @@ def report_failure(result):
                 if message is not None:
                     return StepResult.FAILURE, [message]
 
-            return StepResult.RETRY, []
+            return StepResult.RETRY, [result[1]]
```
manishtomar: Since this code happens to have extracted `result`, it will be better to use `error` instead of `result[1]` instead.

manishtomar:
LGTM

radix:
jenkins please run unit tests

lvh:

radix:
merging, thanks :)

PR #1233, opened 2015-03-23
switch to Container-based Travis builds

radix:
See http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 python:
     - 2.7
 


lvh:
LGTM assuming builds pass :)

radix:
So it eventually passed, and the build was significantly faster than previous builds, but it didn't seem to do anything for the amount of time we had to wait to get the build to *start*. :-(

Well, I'll merge it. Thanks @lvh.

PR #1239, opened 2015-03-24
Update active after delete

### LINKED ISSUE #1234: `active` servers aren't being updated after server deletion
Right now:

- the convergence loop updates the `active` list in the group state in cassandra based on the gathering step of convergence
- DeleteServer returns `SUCCESS` when it sees the server successfully deleted.
- When all steps return `SUCCESS`, no further convergence iterations are done

this all means that `active` never gets updated after a successful server deletion.

radix:
Fixes #1234

This adds an extra gathering step after getting SUCCESS from all steps so we can update the `active` servers list.

- I've updated a few integration tests that pass with this change
- I've also improved the error reporting in check_for_expected_number_of_building_servers in the integration tests a bit, since it was somewhat misleading and also lacked some interesting information

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -338,6 +338,19 @@ def converge_one_group(log, group_locks, tenant_id, group_id, version,
         log.err(None, 'converge-non-fatal-error')
     else:
         if result in (StepResult.FAILURE, StepResult.SUCCESS):
+            # Do one last gathering + writing to `active` so we get updated
+            # based on any DELETEs or other stuff that happened.
+
+            sg_eff = Effect(GetScalingGroupInfo(tenant_id=tenant_id,
+                                                group_id=group_id))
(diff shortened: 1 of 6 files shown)


lvh:
Awesome, LGTM assuming integration tests pass.

lvh:

otter/convergence/service.py
```diff
@@ -124,13 +124,23 @@ def execute_convergence(tenant_id, group_id, log,
         yield do_return(StepResult.SUCCESS)
     results = yield steps_to_effect(steps)
 
-    order = [StepResult.FAILURE, StepResult.RETRY, StepResult.SUCCESS]
+    severity = [StepResult.FAILURE, StepResult.RETRY, StepResult.SUCCESS]
```
manishtomar: +1. Better name.

radix:
Thanks @lvh and @manishtomar :)

PR #1240, opened 2015-03-25
log as effect idea

manishtomar:
Logging is a side-effect and this is one way to treat it as one. The idea binding of log to different fields is inspired from `TenantScope`. This is just an idea for now and would like comments before I move further along.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -82,7 +82,7 @@ def _update_active(scaling_group, active):
 
 
 @do
-def execute_convergence(tenant_id, group_id, log,
+def execute_convergence(tenant_id, group_id,
                         get_all_convergence_data=get_all_convergence_data,
                         plan=plan):
     """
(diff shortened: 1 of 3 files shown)


radix:
I'm also interested in exploring this, the problem has been floating around in my head for a while (especially since we're now losing some binding information between effect-producing code and effect-performing code), and your solution does seem to solve that problem. I think this is a pretty good solution!

as an aside, I want to spend a bit of time thinking about how to make it easier to specify these sort of "contextualizing" effects, since it takes quite a bit of complexity to implement them, but this shouldn't affect this PR :)

otter/convergence/service.py
```diff
@@ -115,6 +115,7 @@ def execute_convergence(tenant_id, group_id, log,
     desired_group_state = get_desired_group_state(
         group_id, launch_config, group_state.desired)
     steps = plan(desired_group_state, servers, lb_nodes, now)
+    yield Effect(BoundLogIntent(log, log_plan(steps)))
```
radix: It seems like the BoundLogIntent should be wrapped at a much higher level, like at the same place we're calling log.bind, though I assume you were just putting it here as a way to test it

radix:
I also suggest to make this easier to use to maybe add a new method to BoundLog, maybe `eff_msg` and `eff_err` or something, so it's just as simple to use as the existing `msg` and `err` functions. These would return `Effect(Log(...))`

radix:
This looks nice, but I still think we should have some convenience methods for creating these so we don't have to be so verbose when we want to create a logging effect.

manishtomar:
Closing for now. Will reopen sometime later with proper tests and usage.

Issue #1244, opened 2015-03-25
CLBChangeNode should handle over limit errors


lvh:
Unfortunately, it appears that 413 (normally for requested entity too large) is used for several failure modes, specifically "too many nodes on the CLB" and "too many requests".

```
[11:22:19] <lvh> which one of those errors do I get back if the LB has too many nodes?
[11:22:38] <lvh> the obvious one is  overLimit (413) but IIRC that's also used if you're issuing too many requests
[11:27:33] <SCO_bmetzler> lvh:  There is no direct way to figure that out.  The only way to figure this out is to pull the Rate Limits programmatically and test against those.  http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/Rate_Limits-d1e821.html
```

#1214 exists to handle not issuing the requests that cause this in the first place, which leads me to believe that the correct resolution for *this* ticket is to just retry.

Issue #1243, opened 2015-03-25
CLBChangeNode should fail on 400


lvh:
*Except* that the [relevant documentation](http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173.html#PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173-Request) that 400 can also happen on (ostensibly transient) load balancer faults. On the one hand, we need to eventually stop retrying if something is really wedged, on the other, we don't want to wedge a group because CLB accidentally threw a 400 when it meant 500.

Issue #1242, opened 2015-03-25
CLBChangeNode should retry on 503 Service Unavailable


lvh:
Relevant docs: http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173.html#PUT_updateNode_v1.0__account__loadbalancers__loadBalancerId__nodes__nodeId__Nodes-d1e2173-Request

PR #1245, opened 2015-03-25
ChangeCLBNode should return success, [reasons]

### LINKED ISSUE #1172: ChangeCLBNode should return success, [reasons]

lvh:
Closes #1172. (Tagged in commit.)

This adds basic support for handling CLB Change Node responses. Specifically, it knows how to handle 202 (all good) and 404 (missing server). This does not cover all the cases. According to the docs, the following error response codes exist:

- loadbalancerFault (400, 500), which is an exciting one. Despite being grouped together, it seems like 400 should be a hard failure (see below), and 500 should be a retry, although the fact that they group them together seems to imply that they reserve the right to throw 400 on internal problems... Filed a ticket: #1241.
- serviceUnavailable (503), which should be a retry. Filed a ticket: #1242.
- badRequest (400), which should probably be a hard failure, except how do I know if it's a transient fault in the LB? Filed a ticket:  #1243.
- unauthorized (401), handled by the underlying reauth logic.
- overLimit (413), which is really hard to handle. Filed a ticket: #1244.
- itemNotFound (404), handled by this PR.

I used a dict to store the ways different error codes are handled. This was useful so that we can statically verify that we have the correct codes in the `success_pred` of `ChangeCLBNode.as_effect`.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -455,7 +455,7 @@ def _clb_check_change_node(result):
     """
     Check to what extent a :class:`ChangeCLBNode` response was successful.
     """
-    return None
+    return StepResult.SUCCESS, []
 
 
 def _rackconnect_bulk_request(lb_node_pairs, method, success_pred):
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -451,11 +451,16 @@ class ChangeCLBNode(object):
         return eff.on(_clb_check_change_node)
 
 
-def _clb_check_change_node(result):
+def _clb_check_change_node(step, result):
     """
     Check to what extent a :class:`ChangeCLBNode` response was successful.
     """
-    return StepResult.SUCCESS, []
+    response, body = result
+
+    if response.code == 202:
+        return StepResult.SUCCESS, []
+    else:
+        return StepResult.RETRY, []
 
 
 def _rackconnect_bulk_request(lb_node_pairs, method, success_pred):
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -440,6 +440,7 @@ class ChangeCLBNode(object):
 
     def as_effect(self):
         """Produce a :obj:`Effect` to modify a load balancer node."""
+        handled_codes = _clb_check_change_node_handlers.keys()
         eff = service_request(
             ServiceType.CLOUD_LOAD_BALANCERS,
             'PUT',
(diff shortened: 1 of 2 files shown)


lvh:
harald.jpg

otter/convergence/steps.py
```diff
@@ -440,14 +440,40 @@ class ChangeCLBNode(object):
...
+    handler = _clb_check_change_node_handlers[response.code]
+    return handler(step, result)
+
+
+def _clb_check_change_node_retry_on_404(step, result):
+    """
+    When updating a node results in a 404, convergence should be retried.
+    """
+    return StepResult.RETRY, [{'reason': 'CLB node not found',
+                               'lb': step.lb_id,
```
manishtomar: Structured info is cool :). I am assuming we should do this everywhere. Do you think `clb_id` and `clb_node_id` would be better due to rcv3 ids also?

otter/test/convergence/test_steps.py
```diff
@@ -451,21 +453,28 @@ def test_change_load_balancer_node(self):
...
+            'loadbalancers/abc123/nodes/node1',
+            data={'condition': 'DRAINING',
+                  'weight': 50},
+            success_pred=has_code(*handled_codes)).intent
+        self.assertEqual(eff.intent, expected_intent)
+
+        (on_success, on_error), = eff.callbacks
+        self.assertIdentical(on_error, None)
+        self.assertIdentical(on_success.func, _clb_check_change_node)
+        self.assertEqual(on_success.args, (change_node,))
```
manishtomar: Non-blocking: This is literal testing of callback functions. Would you consider testing by resolving effect with status codes and looking at the result? I can understand if you don't want to do that since there are separate tests for `_clb_check_change_node` and it is expected to be used elsewhere.

manishtomar:
LGTM modulo comments. All non-blocking :)

otter/convergence/steps.py
lvh: i think it should be `"lb_type": "clb"` then, i.e. have predictable names :)

otter/test/convergence/test_steps.py
lvh: The other reason is that I specifically wanted to make sure the handled codes bit is correct :-) Also, this entire thing gets thrown out once we use Ying's code introduced in #1246, so I'm going to pass for now.

lvh:
I've created https://github.com/rackerlabs/otter/issues/1259 to keep track of better structured information.

Thanks for the review, @manishtomar!

PR #1246, opened 2015-03-25
Parse some generic CLB errors, mainly for single node changes

cyli:
As per discussion in #1189, have client calls that can include structured information.

This also encapsulates our knowledge about what errors can occur during particular calls.

### DIFF
--- a/otter/test/test_cloud_client.py
+++ b/otter/test/test_cloud_client.py
@@ -10,27 +10,67 @@ from effect import (
     TypeDispatcher,
     base_dispatcher,
     sync_perform)
+from effect.testing import EQFDispatcher
 
 from twisted.trial.unittest import SynchronousTestCase
 
 from otter.auth import Authenticate, InvalidateToken
 from otter.cloud_client import (
+    CLBDeletedError,
+    CLBPendingUpdateError,
+    CLBRateLimitError,
+    NoSuchCLBError,
+    NoSuchCLBNodeError,
     ServiceRequest,
     TenantScope,
     add_bind_service,
+    change_clb_node,
     concretize_service_request,
     perform_tenant_scope,
     service_request)
 from otter.constants import ServiceType
-from otter.test.utils import resolve_effect, stub_pure_response
+from otter.test.utils import (
+    resolve_effect,
+    stub_pure_response)
 from otter.test.worker.test_launch_server_v1 import fake_service_catalog
 from otter.util.http import APIError, headers
 from otter.util.pure_http import Request, has_code
 
 
+fake_service_configs = {
+    ServiceType.CLOUD_SERVERS: {
+        'name': 'cloudServersOpenStack',
+        'region': 'DFW'},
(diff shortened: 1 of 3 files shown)


radix:
Sorry I've been so out of touch about this stuff

I think this is going in a good direction. It should be pretty easy to continue writing specific functions for specific operations that can then provide more useful exceptions.

otter/test/utils.py
```diff
@@ -370,6 +370,16 @@ def __init__(self, code, headers, data=None):
...
         self._data = data
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, self.__class__) and
+            self.code == other.code and
+            self.headers == other.headers and
+            self._data == other._data)
+
+    def __neq__(self, other):
```
radix: This is supposed to be `__ne__`

otter/test/utils.py
```diff
@@ -370,6 +370,16 @@ def __init__(self, code, headers, data=None):
...
+        return self != other
```
radix: and this should be `not self == other`

otter/test/test_cloud_client.py
```diff
@@ -304,3 +294,113 @@ def test_perform_srvreq_nested(self):
             sync_perform(self.dispatcher, Effect(tscope)),
             ('concretized', self.authenticator, self.log, self.service_configs,
              1, ereq.intent))
+
+
+class CLBClientTests(SynchronousTestCase):
+    """
+    """
```
lvh: Missing docstring here :)

otter/test/utils.py
cyli: Thanks :)

otter/test/utils.py
lvh: This doesn't actually work, so I'm guessing it's missing some test coverage.

1. The magic method for inequality is spelled `__ne__`, not `__neq__`
2. To implement inequality in terms of equality, you want to do `not (a == b)`, not with `!=`, otherwise you get a recursion error.
3.

cyli: @lvh:  thanks for the explanation :D  I should have looked this up before I did it - I was just trying to get the test to pass, so pretty sloppy :)

otter/test/utils.py
```diff
@@ -732,6 +742,9 @@ def get_fake_service_request_performer(stub_response):
     :param service_request: the :class:`ServiceRequest` to "perform"
     :param stub_response: a tuple of (:class:`StubResponse`, string body),
         supposedly the "response" of an http request
+
+    TODO: hopefully once all service requests are made by the cloud client,
+    this can be moved into the client's test module instead.
```
lvh: Could we move this to a tech debt ticket instead?

lvh:
Other than the TODO thing, which is nonblocking, this LGTM.

cyli:
Apologies for not getting around to this yet - will remove that.

PR #1250, opened 2015-03-26
Stop using versions for dirty flags, switch to one dirty flag per request

radix:
The goal here is to allow us to use a ZooKeeper "child watch" to receive immediate notification when a convergence request is made. The current system of only having one dirty flag that has its version bumped every convergence request doesn't work, because ZK child watches don't get triggered when child versions get bumped. They only get triggered when children are created or deleted. Thus, this PR changes the dirty bookkeeping to create a new node for every ZK request, with a random suffix. This allows us to create a child watch on /groups/divergent and get all the events we need to immediately converge when necessary.

The downside to this is that we lose the optimization in the old version-based approach, which allowed us to only converge once even if multiple requests have been received. With this system there will always be one convergence per request, which is not the end of the world but also not as efficient as it could be.

If anyone has any better ideas on how to do this so that we can maintain that optimization, please let me know.

Note that this PR does not implement the ZK watcher just yet, it just changes the dirty flag formatting.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -20,6 +20,8 @@ from pyrsistent import thaw
 
 import six
 
+from toolz.itertoolz import unique
+
 from twisted.application.service import MultiService
 
 from otter.cloud_client import TenantScope
@@ -32,7 +34,7 @@ from otter.convergence.planning import plan
 from otter.models.intents import GetScalingGroupInfo, ModifyGroupState
 from otter.models.interface import NoSuchScalingGroupError
 from otter.util.fp import assoc_obj
-from otter.util.zk import CreateOrSet, DeleteNode, GetChildrenWithStats
+from otter.util.zk import Create, DeleteNode, GetChildren
 
 
 def server_to_json(server):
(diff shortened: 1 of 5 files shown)


lvh:
Coveralls failure seems spurious; it  doesn't know if anything's changed...

lvh:
LGTM, thanks for working on this.

radix:
There's a real reduction if coverage, in the ZK model code. I'll fix that.

radix:
ok, @manishtomar pointed out that this doesn't make any sense, since we only need to be notified of the transition from no convergence requests -> some convergence requests. We don't need to be notified of further requests when there is already a convergence going on.

At least I'm pretty sure this is true. I'll consider the possibility of race conditions, but we could keep the 10 second check loop as a fall-back in the worst case.

lvh:
@radix So, wouldn't that be covered by just setting `/convergence/{tenant}` to 1 when you get an incoming request? Which is kind of like the dirty flag, I guess, you just stop caring about the value because you can't subscribe to changes in that value :)

radix:
@lvh I don't think I follow... We still do need to know which group has diverged, so I don't think just flagging the tenant is enough. We also already don't care about the value of the dirty flags, we only use its existence + its version number to define the behavior of convergence, the content is always just set to the string `"dirty"`.

lvh:
```
lvh [3:11 PM] 
radix: okay, so I don't understand what the problem is with the dirty flags

lvh [3:11 PM]
radix: I understand why you don't care about the randomness in the names, you don't have to be notified every time there's something that triggers convergence

lvh [3:12 PM]
only when there's a state change

lvh [3:12 PM] 
so why doesn't `/convergence/{group}` work

radix [3:12 PM] 
lvh: so the latest idea is that this PR isn't necessary at all, and the flag management in `master` should still work and be watchable
```

PR #1249, opened 2015-03-26
Return ConvergeLater from convergence if there are building servers

cyli:
Hopefully this fixes #1235.

This returns ConvergeLater only if there are building servers that aren't being deleted.

Also fixed an issue where it's possible for building servers to be drained.

PR #1255, opened 2015-03-26
`ConvergeLater` takes reasons

### LINKED ISSUE #1251: ConvergeLater should take `reasons`
To make logging better and debugging easier, it'd be nice if ConvergeLater took a parameter to return as the `reasons` list along with `StepResult.RETRY`. That way we can include the specific reasons we're retrying a convergence, like "servers are building".

manishtomar:
Closes #1251

### DIFF
--- a/otter/test/convergence/test_service.py
+++ b/otter/test/convergence/test_service.py
@@ -632,7 +632,7 @@ class ExecuteConvergenceTests(SynchronousTestCase):
         def plan(*args, **kwargs):
             return pbag([
                 TestStep(Effect(Constant((StepResult.SUCCESS, [])))),
-                ConvergeLater(),
+                ConvergeLater(reasons=freeze(['mywish'])),
                 TestStep(Effect(Constant((StepResult.SUCCESS, []))))])
 
         eff = execute_convergence(self.tenant_id, self.group_id, log,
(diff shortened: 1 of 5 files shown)


lvh:
This fails the tests, because it doesn't update the places that previously created instances of `ConvergeLater`.

manishtomar:
Yes. Thanks :) I am fixing it.

otter/test/convergence/test_service.py
```diff
@@ -632,7 +632,7 @@ def test_returns_retry(self):
         def plan(*args, **kwargs):
             return pbag([
                 TestStep(Effect(Constant((StepResult.SUCCESS, [])))),
-                ConvergeLater(),
+                ConvergeLater(reasons=freeze(['mywish'])),
```
cyli: Non-blocking - are reasons supposed to be a pyrsistent data structure?  If so, should that be in the ConvergeLater attributes constructor?

cyli:
Aside from the above nitpick about persistent data structures, LGTM :)

cyli:
Apologies, #1249 got merged :|  So possibly needs rebasing, etc.

otter/test/convergence/test_service.py
manishtomar: > Non-blocking - are reasons supposed to be a pyrsistent data structure?

The need to be prysistent DS so that they are hashable and usable in tests. 

> If so, should that be in the ConvergeLater attributes constructor?

I tried putting `Attribute('reasons', instance_of=PVector)` in constructor but `PVector` conflicted with some other class and I didn't want to spend time figuring out how to fix it.

cyli:
LGTM!

radix:
LGTM! Thanks :)

PR #1258, opened 2015-03-26
Speed up mimic tests

cyli:
Refactor the cloudcafe tests so that retries use the `retrying` library, and then scale down the interval and timeouts if we are using mimic, capping the max timeout for mimic at 60 seconds currently.

This should make the integration tests fail faster on mimic, so the non-gating convergence tests can actually maybe produce useful results (we can see if any new tests pass).

I will probably need to go through the scheduling tests, and set it not to scale down the timeouts for mimic, but right now all the worker non-schedule tests pass, the converger non-schedule gating tests pass, and the converger non-schedule failing tests all fail within 15:04 minutes.

Prior to this change, I gave up running the converger non-schedule failing tests after about 47 minutes.

The one issue is that if a test should take more than 1 minute with the converger, this can make it artificially fail, and we might not know.  :|

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/fixtures.py
+++ b/autoscale_cloudroast/test_repo/autoscale/fixtures.py
@@ -374,77 +374,74 @@ class AutoscaleFixture(BaseTestFixture):
         return [s for s in self.get_non_deleting_servers() if is_in_group(s)]
 
     def verify_server_count_using_server_metadata(self, group_id,
-                                                  expected_count):
+                                                  expected_count,
+                                                  time_scale=True):
         """
         Asserts the expected count is the number of servers with the groupid
         in the metadata. Fails if the count is not met in 60 seconds.
         """
-        end_time = time.time() + 60
-        while time.time() < end_time:
+        def verify(elapsed_time):
             actual_count = len(
                 self.get_group_servers_based_on_metadata(group_id)
             )
-            if actual_count is expected_count:
-                break
-            time.sleep(5)
-        else:
-            self.fail('Waited 60 seconds, expecting {0} servers with group id '
-                      ': {1} in the '
-                      'metadata but has {2} servers'.format(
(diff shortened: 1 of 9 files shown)


cyli:
Ok, the schedule policy tests (for worker) now pass completely on my dev vm.  I guess it might be ready for a look now - does this seem like a terrible idea?

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -455,61 +452,69 @@ def get_server_list():
...
+                          server_name, elapsed_time))
+        return self.autoscale_behaviors.retry(
+            check_deleted,  timeout=900, time_scale=time_scale)
 
     def delete_nodes_in_loadbalancer(self, node_id_list, load_balancer):
         """
         Given the node id list and load balancer id, check for lb status
         'PENDING UPDATE' and delete node when lb is ACTIVE
         """
+        class _TimeoutError(Exception):
```
manishtomar: Could you move this above and make it generic that can be raised by `self.retry`?

manishtomar:
LGTM modulo comment above :)

Also, need to forward merge this.

autoscale_cloudroast/test_repo/autoscale/fixtures.py
cyli: Weird, I don't remember why I did this instead of `self.fail` like I did everywhere else - would it be ok if I just changed it to the `self.fail` instead? Retrying will just raise the last error if it can't retry anymore, right?

cyli: hehehe, thanks :D  thinking the same thing!

cyli:

cyli:

cyli:

cyli:

radix:
looks good, merging :)

PR #1261, opened 2015-03-27
Updating DELETING status for convergence

### LINKED ISSUE #1179: force-deleting groups doesn't delete the servers in convergence
Based on discussions, the tasks to do are:
- [ ] Set group status to DELETING on force delete: #1261
- [x] Return 404 on DELETING groups: #1268
- [x] Support `get_deleting` with status in `view_manifest` to use it in Converger: #1268
- [x] Converger deletes all servers for DELETING groups and then deletes the group: #1271 

- [ ] After force delete is fixed, remove 403 from the passing criteria list in the otter/integration/lib/autoscale.py delete_scaling_group() function (https://github.com/rackerlabs/otter/blob/lk_improve_catc6/otter/integration/lib/autoscale.py#L169)

Following is original comment from @radix:

Right now the way force-delete is implemented is

1. set desired to 0
2. kick off an asynchronous controller.converge()
3. delete the group

This works in the worker code because the worker code has everything it needs in-memory. With convergence, the second step above simply marks the group as dirty, and the converger comes along and tries to look up information about the group and can't find any, so it blows up, and the servers never get deleted.

One way to fix this would be to add a "DELETING" status to groups (and set desired to 0), and make it so force-delete sets that status. Then the converger would converge as normal, deleting everything, but then it would need to check the status and actually delete the group from Cassandra after it's done.

We considered making it so any time convergence got a `GroupNotFoundError`, it would just remove all the servers in the group, but that's "scary" from the perspective that there may be situations where a `GroupNotFoundError` is spuriously raised, likely after data loss, and we don't want to delete all the servers in that case.

manishtomar:
Part of #1179.

On group delete request the code updates group status to DELETING and triggers convergence for convergence tenants. For worker tenants, it updates min/max to 0 and triggers worker.

Some refactoring:
* Separate `empty_group` function in controller that only empties the group
* `force_delete_group` function in controller that does switching based on type of tenant
* `rest.group.delete_group` uses `extract_bool_arg`

Note that this does not fix #1179. The converger need look for DELETING status and remove everything. Will be done in coming PRs.


radix:
With this change, we'll also need to make sure we don't allow policy execution or group modification on a group that's currently in the DELETING state. 

I think in the name of expediency we should make all the view functions raise a `NoSuchScalingGroupError` if the status is DELETING. That way we don't need to go update all the existing code to handle DELETING specifically. We can add a `get_deleted` parameter that defaults to False to return groups that are currently in DELETING.

otter/controller.py
```diff
@@ -125,6 +128,66 @@ def obey_config_change(log, transaction_id, config, scaling_group, state,
...
+    :param str trans_id: Transaction ID of request doing this
+    :param otter.models.interface.IScalingGroup scaling_group: the scaling
+        group object
+
+    :return: Deferred that fires with None
+    """
+    if tenant_is_enabled(group.tenant_id, config_value):
+        # For convergence tenants update group status and trigger convergence
+        # TODO: What about status consistency? Will this work if another
+        # process is trying to update status to ERROR at same time?
```
manishtomar: @cyli made a suggestion: Instead of relying on single status field, we can have a separate column "deleting" in CASS that will be set True here. The model will still be `update_status` and `get_status` but it will precede DELETING over other states.

manishtomar:
> I think in the name of expediency we should make all the view functions raise a NoSuchScalingGroupError if the status is DELETING.

@radix : #1269

cyli:
Will this API will work for convergence tenants if `force=False`?  Because if there are servers currently building for the group, it's not stored in the active/pending list.  

The `delete_group` function in `cass.py` checks if `len(pending) + len(active) > 0`, and if so, refuses to delete.  But for convergence, if `pending = desired - active`, and we don't store pending servers in the DB anywhere, then the group will be deleted even if convergence has already started to spin up servers.

radix:
@cyli and I had a conversation about this, and she came up with a pretty good idea:

It's still hopefully possible to avoid maintaining a `pending` list -- in the `force=False` case, we can basically do the same exact codepath as the `force=True` case, but guarded by a check for `desired == 0`. So, we'll still set the group state to `DELETING` and kick off a convergence, and then the converger will always check if the group status is `DELETING` upon quiescence, and delete the group.

manishtomar:

radix:

manishtomar:

manishtomar:

manishtomar:

radix:
@manishtomar have you looked into those integration test failures? do you know what's causing them?

manishtomar:
@radix Yeah. I've been looking at these failures and trying to debug. Can't find any obvious reasons though. They do occur on dev vm. So, clearly something is wrong. I'll spend my time today on this and hope to come out with something.

manishtomar:
When > 100 groups are created and deleted the DB contains rows with deleting=True. Listing groups does not return many valid groups since `_paginated_list` functions limits the number of rows returned to 100 and there can be 10 valid groups in 250 rows. I am going to fix this by creating an index on `deleting` column and using that directly in the query. Will be creating separate PR for that.

radix:

otter/controller.py
```diff
@@ -150,6 +152,91 @@ def obey_config_change(log, transaction_id, config, scaling_group, state,
...
+    def check_and_delete(_group, state):
+        if state.desired == 0:
+            d = trigger_convergence_deletion(log, group)
+            return d.addCallback(lambda _: state)
+        else:
+            raise GroupNotEmptyError(group.tenant_id, group.uuid)
+
+    if tenant_is_enabled(group.tenant_id, config_value):
+        return (trigger_convergence_deletion(log, group)
+                if force else group.modify_state(check_and_delete))
```
radix: I don't understand why this uses `modify_state`. It doesn't look like the state object is being mutated anywhere, so it seems like a no-op, OR perhaps just a roundabout way to lock the group. But since you're not locking the group for the `force` case, I'm not sure why you'd do it in the non-`force` case.

manishtomar: In force case, we are not concerned about current desired and hence trigger convergence deletion since converger will ignore desired. Without force, the check for desired needs to be atomic. Otherwise, one node executing policy could be reading desired 0 and changing it to say 2 returning success and another node deleting could be reading 0 and triggering deletion instead of returning `GroupNotEmptyError`. With the lock, we ensure that deletion and policy execution do not conflict each other.

radix: I see. To be nitpicky, then, I have two suggestions:

1. put a comment inside `check_and_delete`, indicating that we need a lock surrounding the read and assignment of DELETING
2. put a comment in the tenant_is_enabled block indicating that we _don't_ need a lock in the force case, because we're just "atomically" updating the DELETING state
3. break the conditional-expression into a conditional-statement. This was actually a significant impediment to my understanding this code for about 10 minutes. I think a statement instead of an expression is better because it's imperative code, not just deciding between two expressions for their return values, which is where I would normally use an if-expression.

radix:
No other points. LGTM other than my nitpicks :) Thank you very much!

otter/controller.py
manishtomar: Good points. Will make the change.

radix:

manishtomar:

radix:
LGTM after merged forward. thanks @manishtomar :)

PR #1262, opened 2015-03-27
scheduler test fix

manishtomar:
The test schedules policy execution 5 seconds from now. It should consider that 5 seconds when waiting for scheduler to kick in.


radix:
LGTM :) +1

PR #1268, opened 2015-03-27
view_manifest gets deleting groups

manishtomar:
`view_manifest` supports getting DELETING groups by passing `get_deleting=True`. All group viewing methods return `NoSuchScalingGroupError` when trying to access DELETING group including listing groups. This way REST API's accessing deleting groups will get 404. 
- refactored cass tests
    - moved view_manifest tests to separate class
    - moved modify_state tests from test_interface to cass tests
- as a result of refactor, realized that returned manifest without policies was not conforming to model schema. Hence, made "scalingPolicies" optional in schema

Closes 2nd and 3rd task of #1179.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -112,7 +112,7 @@ _cql_create_group = (
 _cql_view_manifest = (
     'SELECT "tenantId", "groupId", group_config, '
     'launch_config, active, pending, "groupTouched", '
-    '"policyTouched", paused, desired, created_at '
+    '"policyTouched", paused, desired, created_at, status '
     'FROM {cf} '
     'WHERE "tenantId" = :tenantId AND "groupId" = :groupId')
 _cql_insert_policy = (
@@ -641,7 +641,8 @@ class CassScalingGroup(object):
             return d
         return wrapper
 
-    def view_manifest(self, with_policies=True, with_webhooks=False):
+    def view_manifest(self, with_policies=True, with_webhooks=False,
+                      with_status=False):
         """
         see :meth:`otter.models.interface.IScalingGroup.view_manifest`
         """
@@ -666,12 +667,15 @@ class CassScalingGroup(object):
                 group)
 
         def _generate_manifest_group_part(group):
-            return {
+            m = {
                 'groupConfiguration': _jsonloads_data(group['group_config']),
                 'launchConfiguration': _jsonloads_data(group['launch_config']),
                 'id': self.uuid,
                 'state': _unmarshal_state(group)
             }
+            if with_status:
+                m['status'] = group['status']
+            return m
 
         view_query = _cql_view_manifest.format(
             cf=self.group_table)
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -503,9 +503,7 @@ def verified_view(connection, view_query, del_query, data, consistency,
     `exception_if_empty` if group's status is DELETING
 
     :param bool get_deleting: Should it return deleting group?
-
-    TODO: Should there be seperate argument for view_consistency and
-    del_consistency.
+    :return: Deferred that fires with result of executing view query
     """
     def _check_resurrection(result):
         if len(result) == 0:
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -90,7 +90,7 @@ def serialize_json_data(data, ver):
 # Otherwise it won't.
 #
 # Thus, selects have a semicolon, everything else doesn't.
-_cql_view = ('SELECT {column}, created_at, status FROM {cf} '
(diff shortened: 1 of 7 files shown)


otter/models/cass.py
```diff
@@ -493,11 +495,14 @@ def assemble_webhooks_in_policies(policies, webhooks):
...
-    not have "created_at" in it.  Any resurrected entry is deleted and
-    `exception_if_empty` is raised.
+    Ensures the view query on the group does not get resurrected row,
+    i.e. one that does not have "created_at" in it.  Any resurrected entry is
+    deleted and `exception_if_empty` is raised. Also raises
+    `exception_if_empty` if group's status is DELETING
+
+    :param bool get_deleting: Should it return deleting group?
 
     TODO: Should there be seperate argument for view_consistency and
```
cyli: Non-blocking, since this wasn't touched, but this TODO seems outdated, and could probably be deleted.

otter/models/cass.py
```diff
@@ -505,13 +510,16 @@ def verified_view(connection, view_query, del_query, data, consistency,
...
-        if result[0].get('created_at'):
-            return result[0]
-        else:
+        group = result[0]
+        if group.get('created_at') is None:
             # resurrected row, trigger its deletion and raise empty exception
             log.msg('Resurrected row', row=result[0], row_params=data)
             connection.execute(del_query, data, consistency)
             raise exception_if_empty
+        # Do not return group if group is deleting and we are told to do so
```
cyli: Nitpick - "Do not return group if it is deleting unless we are told to do so" ?

otter/models/cass.py
```diff
@@ -641,7 +649,12 @@ def wrapper(*args):
             return d
         return wrapper
 
-    def view_manifest(self, with_policies=True, with_webhooks=False):
+    def _group_status(self, status, deleting):
```
cyli: Do we need to handle the case where the status is `None` or an invalid value?  There seem to be some `DISABLED` statuses currently in the DB.

otter/test/models/test_cass_models.py
```diff
@@ -2274,6 +2092,185 @@ def not_empty_error(lockpath, recursive):
...
+
+    def test_with_webhooks(self):
+        """
+        Viewing manifest with_webhooks=True returns webhooks inside policies
+        that matches the `model_schemas.manifest`
+        """
+        self.verified_view.return_value = defer.succeed(self.vv_return)
+
+        # Getting policies
+        policies = [group_examples.policy()[i] for i in range(3)]
```
cyli: Non-blocking - could just get the list once, and pick the first 3 elements.  :)  `group_examples.policy()[:3]`?

otter/test/models/test_cass_models.py
```diff
@@ -3353,7 +3312,28 @@ def test_list_states_deletes_resurrected_groups(self):
...
+                                        active={},
+                                        pending={},
+                                        group_touched='0001-01-01T00:00:00Z',
+                                        policy_touched={},
+                                        paused=False)])
+
+    def test_list_states_filters_deleting_groups(self):
+        """
+        Groups that are deleting based on "deleting" column to be true are not
+        returned
```
cyli: Should we also assert that no call to delete the deleted group is made?  (It's only filtered, and not further deleted)

cyli:
Awesome cass test cleanup!  Removed several useless tests, and much less repeat code.  :)

Would it make sense to add one more test to the manifest tests such that if `get_deleting=True`, then the status appears in the manifest even if the group is not in deleting?

otter/models/cass.py
manishtomar: Very good point @cyli. Thanks :)

manishtomar: I was thinking either
a) Changing status of all `DISABLED` groups to `ERROR` and rest to `ACTIVE`
OR
b) Change code to return `ERROR` on `DISABLED` and `ACTIVE` on `None`.
What do you think?

radix: What's `DISABLED`? Maybe we should just properly support it (perhaps treat it like `DELETING`, but without the the real deletion done by the converger)? Or is it really just some dead status that we don't want to support any more?

manishtomar: While collecting metrics, I found many groups that had incorrect launch config or the tenant itself was fake. So, I changed the status of these groups to `DISABLED` and [filtered](https://github.com/rackerlabs/otter/blob/master/otter/metrics.py#L266) these out from when collecting metrics. I was not sure what statues we all decide on at that time. I think changing these groups to ERROR is fine.

radix: I see! ERROR makes sense :)

manishtomar:

radix:
LGTM :)

PR #1269, opened 2015-03-28
empty exception on DELETING groups

### LINKED ISSUE #1265: Return 404 on DELETED groups
As part of implementing #1179, otter is changing group status instead of actually deleting the group. Hence, to keep API compatible it should return 404 on all the REST calls on the DELETED group. This was suggested in https://github.com/rackerlabs/otter/pull/1261#issuecomment-87070437. 

This can be done by returning `NoSuchScalingGroupError` on all model APIs except `get_status` API.

manishtomar:
Closes #1265. Raises `NoSuchScalingGroupException` on group with "DELETING" status. This will return 404 on any DELETING group.

EDIT: I've to update the queries. Putting it "in progress" now.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -42,6 +42,7 @@ from otter.models.interface import (
     NoSuchWebhookError,
     PoliciesOverLimitError,
     ScalingGroupOverLimitError,
+    ScalingGroupStatus,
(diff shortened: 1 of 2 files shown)


manishtomar:
Closing. Will put updated changes in #1268 as they are tightly coupled now.

PR #1271, opened 2015-03-29
Converger deleting group

manishtomar:
Converger deletes all servers for DELETING groups and then deletes the group. 

4th task of #1179. Depends on #1268. The tests will pass after that PR is merged.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -28,8 +28,9 @@ from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
 from otter.convergence.model import ServerState, StepResult
 from otter.convergence.planning import plan
-from otter.models.intents import GetScalingGroupInfo, ModifyGroupState
-from otter.models.interface import NoSuchScalingGroupError
+from otter.models.intents import (
+    DeleteGroup, GetScalingGroupInfo, ModifyGroupState)
+from otter.models.interface import NoSuchScalingGroupError, ScalingGroupStatus
 from otter.util.fp import assoc_obj
 from otter.util.zk import CreateOrSet, DeleteNode, GetChildrenWithStats
 
@@ -94,7 +95,7 @@ def execute_convergence(tenant_id, group_id, log,
     :param get_all_convergence_data: like :func`get_all_convergence_data`, used
         for testing.
 
-    :return: Effect of StepResult based on order of severity
+    :return: Effect of most severe StepResult
     :raise: :obj:`NoSuchScalingGroupError` if the group doesn't exist.
     """
     # Huh! It turns out we can parallelize the gathering of data with the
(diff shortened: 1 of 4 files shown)


radix:
Won't the GetScalingGroupInfo call blow up since it'll raise exceptions when the group status is DELETING? I was expecting there to be an extra "get_deleted=True" parameter to be passed which would indicate the group should be returned even if it's deleted.

manishtomar:
> Won't the GetScalingGroupInfo call blow up since it'll raise exceptions when the group status is DELETING?

Yes @radix . It will blow up. Sorry I didn't realize it. I'll add the `get_deleting` parameter too.

radix:

otter/convergence/service.py
```diff
@@ -114,6 +115,10 @@ def execute_convergence(tenant_id, group_id, log,
     launch_config = manifest['launchConfiguration']
     now = yield Effect(Func(time.time))
 
+    group_status = ScalingGroupStatus.lookupByName(manifest['status'])
+    if group_status == ScalingGroupStatus.DELETING:
+        group_state.desired = 0
```
radix: This is a side-effect in an otherwise pure function. Also, it's not in a modify_state call, so it's not actually being persisted (unless I'm missing something), so I'm not sure why you're doing it. Doesn't the controller code already set the desired to 0 when deleting the group?

otter/convergence/service.py
```diff
@@ -121,7 +126,8 @@ def execute_convergence(tenant_id, group_id, log,
     log.msg('execute-convergence',
             servers=servers, lb_nodes=lb_nodes, steps=steps, now=now,
             desired=desired_group_state, active=active)
-    yield _update_active(scaling_group, active)
+    if group_status != ScalingGroupStatus.DELETING:
```
radix: Did you make this conditional to avoid race conditions or something?

I'm imagining a scenario where we're deleting a group and deleting all CLBs/servers, and making this conditional will mean someone watching group state will see "active" servers that aren't actually active any more, for example after they have been removed from the CLB but before they have been deleted.

It's unlikely to be an important detail, since the group is destined to be deleted anyway, but there may be strange situations where it takes a very long time to clean up a group and it may be useful to track the status through the API. Leaving the _update_active unconditional would allow this to happen, so I'm just curious about why we're conditionalizing it.

otter/convergence/service.py
manishtomar: > Doesn't the controller code already set the desired to 0 when deleting the group?

No. The controller only sets DELETING status. This code does not persist the updated desired state. It only uses it when calculating the desired group state. I agree it leaks side-effect. I could instead do 
```python
group_status = ScalingGroupStatus.lookupByName(manifest['status'])
desired_group_state = get_desired_group_state(
   group_id, launch_config, 
   0 if group_status == ScalingGroupStatus.DELETING else group_state.desired)
```
This doesn't look like having side-effect and I think it's clearer. What do you think?

cyli: I was wondering about that too, and seeing where group_state was being written or something.  I think that makes more sense, and makes it clearer that we aren't changing anything in the, just deleting the group resources

radix: Yeah, that sounds better @manishtomar, thanks :)

otter/convergence/service.py
manishtomar: Currently, user facing APIs do not get DELETING groups to be backward-compatible. This was the reason condition was added otherwise `_update_active` raised `NoScalingGroupError`.

radix: Ah! Of course, I see now. Okay, that's fine, don't worry about it :)

radix:
@manishtomar LGTM after you make that `desired` change.

manishtomar:

manishtomar:
@radix desired change made and integration tests pass. Ok to merge?

radix:
Yep! I'll merge. Thanks @manishtomar  :D

PR #1272, opened 2015-03-30
update to the latest TxKazoo API

radix:
This won't work until 0.0.6 is released, but here's the code.

I made no attempt to clean up the super mocky unit tests, and in fact I added more mocks. Sorry. :-(

I've verified that the code works when running integration tests on my dev VM.


manishtomar:
LGTM once the build passes :)

radix:
Thanks @manishtomar :)

Issue #1238, opened 2015-03-24
Can't use otter `delete server` REST API on servers that were created with convergence


radix:
The `active` data stored in cassandra for convergence-created servers doesn't have all the data that is usually there, and the deletion code depends on this. Here's the exception I'm getting:

```
Traceback (most recent call last):
  File "/opt/otter/otter/supervisor.py", line 384, in start
  self.log, self.trans_id, self.scaling_group, self.server_info)
  File "/opt/otter/otter/supervisor.py", line 196, in execute_delete_server
    return d.addCallback(got_request_bag)
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 307, in addCallback
    callbackKeywords=kw)
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 296, in addCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 578, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/opt/otter/otter/supervisor.py", line 192, in got_request_bag
    instance_details = server['id'], server['lb_info']
exceptions.KeyError: 'lb_info'
```

manishtomar:
The `delete_server` implementation needs to change: When deleting out of group not from Nova, otter should remove group metadata. Otherwise, it should just delete the server and update it's active list.

radix:
@manishtomar what do you mean "not from Nova"?

manishtomar:
@radix I meant remove from the group but do not delete the server. In our case it is `DELETE .../groupId/servers/serverId?purge=false`

radix:
@manishtomar the problem is that LB information is not available, and otter wants to remove the server from the load balancer.

manishtomar:
Since otter was the one that added to the LB, I think its fine that otter removes it from LB. In convergence it can get that from metadata.

radix:
@manishtomar ah, so you're saying when deleting a server, otter should hit nova directly to determine what load balancer it's associated with it, instead of relying on the cached data in `active`?

manishtomar:
Actually, I had not thought of that but it does sound like a good plan. I was thinking about extracting LB information from server metadata.

radix:
[...]

PR #1273, opened 2015-03-30
Add some tests to make sure that draining servers without LBs are deleted right away

cyli:
Since the current strategy for deleting a server from a group involves just setting it to DRAINING for now (see #1238)


lvh:
LGTM!

PR #1277, opened 2015-03-31
separate column for deleting status

manishtomar:
Based on https://github.com/rackerlabs/otter/pull/1261#discussion_r27331263 this adds another column for DELETING status. This way DELETING status does not interfere with ACTIVE or ERROR and always takes precedence. Please note that this PR does not contain code to _read_ status. That will be in separate PRs under `view_manifest` (most likely a fix to #1268)

This PR depends on supporting boolean types in silverberg that will be released in 0.1.11 version after merging https://github.com/rackerlabs/silverberg/pull/49. Hence, please review that too :)

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -42,6 +42,7 @@ from otter.models.interface import (
     NoSuchWebhookError,
     PoliciesOverLimitError,
     ScalingGroupOverLimitError,
+    ScalingGroupStatus,
     UnrecognizedCapabilityError,
     WebhooksOverLimitError,
     next_cron_occurrence)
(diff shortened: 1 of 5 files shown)


otter/models/cass.py
```diff
@@ -808,7 +809,24 @@ def _do_update(ts, _):
                  'status': status.name},
                 DEFAULT_CONSISTENCY)
 
-        return self.view_config().addCallback(_do_update)
```
cyli: Did this line get deleted by accident?

radix:

cyli:

cyli:

otter/models/cass.py
```diff
@@ -808,7 +809,24 @@ def _do_update(ts, _):
...
+        @self.with_timestamp
+        def set_deleting(ts, _):
```
radix: Given that we're not using the config information here (`_`), we should be able to avoid calling `view_config` in this code path entirely, I think?

manishtomar: `view_config` is called to raise nogrouperror when `update_status` is called on non-existent group. However, I just realized that it will raise nogrouperror on DELETING group as well which is probably ok since setting any status on a deleting group is not helpful? What do you think?

radix: yep, that's fine. LGTM :)

Issue #1228, opened 2015-03-23
test_system_user_delete_some_servers_out_of_band  is mimic specific?


manishtomar:
The above test expects server to be building for 30 seconds and deletes while it is building. This is not a predictable behavior with real Nova. So, maybe this test should just be mimic specific?

PR #1279, opened 2015-03-31
Decorator that allows skipping of integration tests based on whether mimic is used or not

cyli:
Also, update the docstring on one of the tests that should never pass for convergence.

May help with #1228.


autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -134,6 +135,17 @@ def _set_up_clients():
 autoscale_client, server_client, lbaas_client, rcv3_client = _set_up_clients()
 
 
+def skip_if_mimic_is(boolean):
```
manishtomar: Could you rename the `boolean` to `skip` or something else that describes its intent?

cyli: Sure :)

lvh:
LGTM!

PR #1280, opened 2015-03-31
APIs needed for deleting one server in convergence

cyli:
This includes an API for setting a metadata item (which we need to set DRAINING) and for getting server details (to determine whether a server is a member of our group)

So that this can be called for the convergence codepath of #1238.

For more context, I plan on using the APIs here:  https://github.com/rackerlabs/otter/compare/delete_server_with_convergence

### DIFF
--- a/otter/test/test_cloud_client.py
+++ b/otter/test/test_cloud_client.py
@@ -515,7 +515,7 @@ class NovaClientTests(SynchronousTestCase):
             service_request_eqf(
                 stub_pure_response(json.dumps(failure_body), 413)))])
 
-        with self.assertRaises(NoSuchServerError) as cm:
+        with self.assertRaises(NovaRateLimitError) as cm:
             sync_perform(dispatcher, real)
 
         self.assertEqual(cm.exception,
--- a/otter/test/test_cloud_client.py
+++ b/otter/test/test_cloud_client.py
@@ -32,6 +32,7 @@ from otter.cloud_client import (
     add_bind_service,
     change_clb_node,
     concretize_service_request,
+    get_server_details,
     perform_tenant_scope,
     service_request,
     set_nova_metadata_item)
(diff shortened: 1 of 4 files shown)


otter/test/test_cloud_client.py
```diff
@@ -405,3 +414,162 @@ def test_change_clb_node(self):
 
         # all the common failures
         self.assert_parses_common_clb_errors(expected.intent, eff)
+
+
+class NovaClientTests(SynchronousTestCase):
+    """
+    Tests for Nova client functions, such as :obj:`set_nova_metadata_item`.
+    """
+    def _setup_for_set_nova_metadata_item(self):
```
radix: I guess idiomatically I would expect two different test suites (one for `set_nova_metadata_item` and one for `get_server_details`) , and these _setup methods would be `setUp` methods that assign attributes on `self`. Was this a conscious decision to avoid that pattern? I do think that `setUp` methods can often get pretty gross (and do too much wrt what each tests needs), so I would understand a desire to avoid them.

Just curious, not asking you to change anything :)

radix:
LGTM! I'll let you merge after you see my question :)

otter/test/test_cloud_client.py
cyli: I have definitely become slightly more predisposed to avoid `setUp` methods, because I used to write a lot of `setUp` code which put tons of attributes on `self` that I constantly had to look up every time I added a test.  So yes, this was sort of a desire to avoid them, but the main reason for the two different setup functions was that there are some common Nova errors (like rate limiting, no such server) that can share the same assertions.

Maybe a better way to do this would be to put the assertions in their own mixin?  And then go one test suite per client function, with it's own `setUp`?

radix: nope, that's fine. I'll merge!

PR #1282, opened 2015-03-31
Parse server metadata to recognize "DRAINING" state

cyli:
In the convergence models, the server state should be DRAINING if the draining metadata is on the server.

However, DELETED and ERROR states should trump DRAINING, since "DRAINING" just means that the server is pretty much worker but Autoscale would like to delete it.

Part of #1238

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -80,7 +80,9 @@ class ServerState(Names):
 
     DRAINING = NamedConstant()
     """"
-    Autoscale is deleting the server.
+    Autoscale is deleting the server.  This state is meant to supercede Nova's
+    ``BUILD`` and ``ACTIVE`` states, because this state means that the server
+    is basically functional, but that Autoscale would like to delete it.
     """
 
     DELETED = NamedConstant()
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -221,14 +221,21 @@ class NovaServer(object):
 
         :return: :obj:`NovaServer` instance
         """
+        server_state = ServerState.lookupByName(server_json['status'])
+        metadata = server_json.get('metadata', {})
+
+        if (server_state in (ServerState.ACTIVE, ServerState.BUILD) and
+                metadata.get(DRAINING_METADATA[0]) == DRAINING_METADATA[1]):
+            server_state = ServerState.DRAINING
+
         return cls(
             id=server_json['id'],
-            state=ServerState.lookupByName(server_json['status']),
+            state=server_state,
             created=timestamp_to_epoch(server_json['created']),
             image_id=server_json.get('image', {}).get('id'),
             flavor_id=server_json['flavor']['id'],
             links=freeze(server_json['links']),
-            desired_lbs=_lbs_from_metadata(server_json.get('metadata')),
+            desired_lbs=_lbs_from_metadata(metadata),
             servicenet_address=_servicenet_address(server_json))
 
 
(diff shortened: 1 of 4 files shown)


radix:
hmm.

I'm tempted to say we should maintain the same model that nova provides. perhaps we should keep .state a pure reflection of server['state'], and then add a `draining` attribute that is True when that metadata item is there?

What do you think?

cyli:
@radix  I can see your point in that state was originally supposed to reflect the Nova server state.  And instead, this seems to have mutated it.

How would you feel about maybe having `.nova_state` and `.autoscale_state` instead?  The first being the pure reflection of `server['state']`, and the second being a server's autoscale state?

I think having the DRAINING state makes sense in the context of autoscale - there seems to be a real state transition (since building or active servers can go into draining, errored and deleted servers cannot, and building/active/draining servers can all go into errored and deleted states).

Also, planning actually does something different with each of the 5 disjoints sets of servers: active servers that are not draining, building servers that are not draining, draining servers, errored servers, and deleted servers.  Having 5 states makes partitioning the servers by state nice in the planner.

That said, though, it's not like it's hard to partition by the 4 Nova states in the planner, and then further spit the (active, building) sets into (active not draining, building not draining, active or building and draining).  :)

radix:
@cyli 

I admit that this is definitely bikeshedding on _my_ part -- you already have working code, and I'm just being a hassle. But I'll describe some more bikeshedding in slightly more detail, and you can feel free to completely ignore the suggestions, since I realize the long-term development of this code may not be particularly interesting or relevant.

TL;DR = perhaps leave NovaServer general and then just add a `convergence_relevant_status` function that takes a NovaServer and returns some status representation that we care about.

I think I would prefer NovaServer to maintain a general, non-convergence-specific representation of a nova server, and perhaps even move it up to cloud_client.py eventually (like, it'd perhaps make sense for the new `get_server_details` to return it directly, and for `get_all_servers` to also move up there). In that case, I think NovaServer should try to reflect the state of the nova server and all its possible combinations of data.  Then, for the purposes of getting "the interesting abstract status of a NovaServer", we could have a function of `NovaServer -> convergence_status` which basically return DRAINING in preference to any other status.

I strongly discourage you from taking this too seriously. I just wanted to describe the idea in slightly more detail.

cyli:
@radix 

No worries, I like thought experiments.  :)  Hmm... in terms of pure bikeshedding - I agree with you in that what's in the model currently is something I think of as an Autoscale-controlled Nova server - it's not really a general Nova server, since it already has an autoscale-specific layer on it: the desired load balancers.

A pure Nova server model would probably have just the metadata and the Nova statuses.  If we were to go with your suggestion of moving NovaServer to `cloud_client`, I'd suggest we add a new model for AutoscaledServer that wrapped a NovaServer, and interpreted the metadata to get the desired load balancers, autoscaling state, and group ID on top of it.  It could interpret a NovaServer with a `OS-EXT-STS:task_state` of "deleting" to be in DELETED state too, for instance.

This would make the separation between the Nova client and autoscale itself more clear - because really autoscale doesn't care at all about NovaServers that are not AutoscaledServers.

---

In terms of immediate work though to move this PR forward, how would you feel about a property rather than a function?  And if we are going to return a convergence status, should that be a different set of NamedConstants?  If we wanted to keep things general, convergence statuses are not Nova server statuses (Nova doesn't have draining), and hence shouldn't be conflated.  I'm just resistant to it because I'm lazy, and it's just a superset of the Nova statuses - I don't want to go change all the tests.  :)

radix:
@cyli you make an excellent point, and are of course completely right. As far as this PR, I think you should leave it exactly as-is.

radix:
LGTM :)

PR #1286, opened 2015-04-01
make the converger interval configurable as `converger.interval`

radix:
Mostly to allow us to tweak this for integration test purposes, where we probably want to set it to something very low.


cyli:
Can this option be added to https://github.com/rackerlabs/otter/blob/master/config.example.json also?  Other than that, LGTM if ci +1's.  :)

radix:
Good point, @cyli. I've added it with a reasonable-ish default of 30.

radix:

PR #1289, opened 2015-04-02
use ZK watches to trigger convergence faster

### LINKED ISSUE #1236: Trigger convergence based on a watch of /groups/divergent in ZooKeeper
This would allow convergence to start faster when a policy is triggered or configuration is created/updated.

Honestly this probably isn't important to real-world usage (assuming the current check interval of 10s), but it would make the convergence integration tests go a lot faster.

radix:
Fixes #1236.

I had to modify ZKPartitioner to expose `get_current_state` and `get_current_buckets`, so we can make sure not to trigger convergence when we haven't acquired our buckets.

I also had to disable one test that was previously marked as `convergence='yes'`, because it was erroneously detecting a group that hadn't been converged yet as one that had been successfully force-deleted. This test relies on #1179 is fixed.

Other than that, this PR is pretty straightforward, and drastically speeds up the gating integration tests on my machine (~11m to ~5m).

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -446,6 +446,7 @@ class Converger(MultiService):
             # TODO: don't use private stuff :(
             # TODO: only call buckets_acquired if we own some of the children
             # TODO: maybe don't do so much work?
+            # TODO: nagle?
             my_buckets = self.partitioner._get_current_buckets()
             self.buckets_acquired(my_buckets)
 
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -14,6 +14,8 @@ from effect.do import do, do_return
 from effect.ref import Reference
 from effect.twisted import exc_info_to_failure, perform
 
+from kazoo.recipe.partitioner import PartitionState
+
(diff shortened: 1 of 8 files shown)


radix:

radix:
jenkins seems to have gone from running `convergence='yes'` tests in 364 seconds to 166 seconds. yay. :)

otter/convergence/service.py
```diff
@@ -440,6 +456,24 @@ def buckets_acquired(self, my_buckets):
...
         return result
 
+    def divergent_changed(self, children):
+        """
+        ZooKeeper children-watch callback that lets this service know when the
+        divergent groups have changed. If any of the divergent flags are for
+        tenants associated with this service's buckets, a convergence will be
+        triggered.
+        """
+        # It'd sure be nice if we could know *exactly which* flag was just
```
cyli: Random non-blocking thought - can we store all the previous divergent groups, and compare it to the new divergent groups, to see which groups have been added?  Much like the current set comparison.

cyli:
LGTM!

PR #1291, opened 2015-04-02
DeleteServer returns RETRY instead of doing an additional gathering at the end of convergence

radix:
It turns out I had a big brain-fart when fixing #1234. It's fine for DeleteServer to return RETRY to indicate that we should do another gathering after a successful deletion. And so we don't need to explicitly perform one last gathering after a successful convergence.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -220,7 +220,7 @@ class DeleteServer(object):
             next_interval=exponential_backoff_interval(2))
 
         def report_success(result):
-            return StepResult.SUCCESS, []
+            return StepResult.RETRY, ['must re-gather after deletion']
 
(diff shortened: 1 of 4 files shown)


radix:

otter/convergence/steps.py
```diff
@@ -220,7 +220,7 @@ def as_effect(self):
             next_interval=exponential_backoff_interval(2))
 
         def report_success(result):
-            return StepResult.SUCCESS, []
+            return StepResult.RETRY, ['must re-gather after deletion']
```
cyli: Nitpick - maybe "must re-gather after deletion in order to update the active cache" ?

cyli:
Aside from the nitpick, LGTM!

PR #1293, opened 2015-04-02
import SequenceDispatcher

radix:
an unfortunate series of merges led to master having a missing import statement causing test failures. just adding back the import.


cyli:
LGTM once travis +1's :)

PR #1295, opened 2015-04-03
Convergence delete server

cyli:
Part of fixing #1238.

This provides a convergence-specific implementation for removing a server from the scaling group - we can't just check the DB for the server, because that is a cache of servers.  Nova is the source of truth.

And we just have to set the server metadata to DRAINING and kick off a convergence cycle (which will be done in the next PR, which will have a `remove_server_from_scaling_group` function that dispatches to either the supervisor function or this one which.

The branch I'm working towards for the next PR (it still has bits of code from this PR) is:  https://github.com/rackerlabs/otter/compare/delete_server_with_convergence

Review questions:
- Should this live in the controller?  I didn't know where else to put it - maybe it should be in a different file instead
- Wondering about whether it's worthwhile to do the check in parallel
- I just randomly picked a retry value - not sure if we should do 3 times with exponential backoff, since we probably want to block the delete request until all these checks and metadata setting are finished.

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -20,13 +20,14 @@ Storage model for state information:
  * last touched information for group
  * last touched information for policy
 """
+import json
 from datetime import datetime
 from decimal import Decimal, ROUND_UP
-import iso8601
-import json
 
 from effect import Effect, parallel_all_errors
 from effect.do import do, do_return
+
+import iso8601
 
 from six import reraise
 
@@ -39,10 +40,10 @@ from otter.cloud_client import (
     get_server_details,
     set_nova_metadata_item)
 from otter.convergence.composition import tenant_is_enabled
+from otter.convergence.model import DRAINING_METADATA, group_id_from_metadata
 from otter.convergence.service import get_convergence_starter
 from otter.json_schema.group_schemas import MAX_ENTITIES
 from otter.log import audit
-from otter.convergence.model import DRAINING_METADATA, group_id_from_metadata
 from otter.models.intents import GetScalingGroupInfo
 from otter.supervisor import (
     CannotDeleteServerBelowMinError,
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -25,15 +25,34 @@ from decimal import Decimal, ROUND_UP
 import iso8601
 import json
 
-from twisted.internet import defer
+from effect import Effect, parallel_all_errors
+from effect.do import do, do_return
 
-from otter.convergence.composition import tenant_is_enabled
+from six import reraise
+
+from toolz.dicttoolz import get_in
+
+from otter.cloud_client import (
+    NoSuchServerError,
+    get_server_details,
+    set_nova_metadata_item)
 from otter.convergence.service import get_convergence_starter
 from otter.json_schema.group_schemas import MAX_ENTITIES
 from otter.log import audit
-from otter.supervisor import exec_scale_down, execute_launch_config
+from otter.convergence.model import DRAINING_METADATA, group_id_from_metadata
+from otter.models.intents import GetScalingGroupInfo
+from otter.supervisor import (
+    CannotDeleteServerBelowMinError,
+    EvictServerFromScalingGroup,
+    ServerNotFoundError,
+    exec_scale_down,
+    execute_launch_config)
 from otter.util.config import config_value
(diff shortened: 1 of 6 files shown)


otter/test/test_controller.py
```diff
@@ -1139,3 +1163,257 @@ def test_real_convergence_zero_delta(self):
...
+        """
+        dispatcher = SequenceDispatcher([
+            (get_server_details('server_id').intent,
+                lambda _: self.server_details),
+            (EvictServerFromScalingGroup(scaling_group=self.group,
+                                         server_id='server_id'),
+                lambda _: raise_(ValueError('oops')))
+        ])
+        self.assertRaises(ValueError, self._remove, True, False, dispatcher)
+        self.assertEqual(dispatcher.sequence, [])
```
radix: I see you picked up on this :) I'm actually thinking of adding an explicit `SequenceDispatcher.completed` method that returns True when the dispatcher has been exhausted, since it turns out it's a really important part of using this (I've already had tests that continued to pass when I forgot/removed some code at the end of a function...)

otter/controller.py
```diff
@@ -348,3 +371,88 @@ def calculate_delta(log, state, config, policy):
...
+    if purge:
+        eff = set_nova_metadata_item(server_id, *DRAINING_METADATA)
+    else:
+        eff = Effect(
+            EvictServerFromScalingGroup(scaling_group=group,
+                                        server_id=server_id))
+    yield retry_effect(eff, retry_times(3), exponential_backoff_interval(2))
+
+    if not replace:
+        state.desired -= 1
```
radix: I wouldn't mention this in "normal" code, but since this is an Effect-returning function: this is a side-effect that's not encapsulated in an Effect. We have `ModifyGroupState` for changing groups, which might be suitable here, but given that you're returning `state` I'm not sure what your plans are (sorry I haven't read your other in-progress branch yet). You could also return a *new* state object -- you can just return `assoc_obj(state, desired=state.desired - 1)`.

otter/supervisor.py
```diff
@@ -683,3 +685,23 @@ def scrub_otter_metadata(_):
...
+@attributes(['scaling_group', 'server_id'])
+class EvictServerFromScalingGroup(object):
+    """
+    An Effect intent which indicates that a server should be evicted from a
+    particular group.
+    """
+
+
+@deferred_performer
+def perform_evict_server(log, transaction_id, dispatcher, intent):
```
radix: I assume you put `transaction_id` up here because you didn't want to pass it into the `convergence_remove_server_from_group` function? I'm wondering whether the tradeoff in complexity of setting up the dispatcher is worth it. Not really sure, I guess I'll go look in your other branch :)

otter/controller.py
cyli: Oh I forgot about assoc_obj - great suggestion thanks!

otter/controller.py
```diff
@@ -348,3 +371,88 @@ def calculate_delta(log, state, config, policy):
...
+        raise ServerNotFoundError(group.tenant_id, group.uuid, server_id)
+
+    group_id = group_id_from_metadata(
+        get_in(('server', 'metadata'), server_info, {}))
+    if group_id != group.uuid:
+        raise ServerNotFoundError(group.tenant_id, group.uuid, server_id)
+
+
+@do
+def _can_maybe_scale_down(group, server_id):
```
manishtomar: Nitpick: "can" and "maybe" are probably repetetive. You could just have `_can_scale_down`?

otter/supervisor.py
manishtomar: Not sure if its possible, but will be nice to receive `supervisor` as argument to the performer.

otter/controller.py
cyli: Actually, what library is `assoc_obj` part of?

manishtomar: `attrs`

cyli:
I went with not using `attrs`, because we already use `characteristic` everywhere, and I wasn't sure if it'd be confusing to mix them together in the same codebase.  Creating a new GroupState isn't very onerous, though, and if we wanted to convert everything to `attrs` instead we can do that in a separate PR.

radix:

radix:
Thanks for addressing those earlier review comments, @cyli. LGTM!

Issue #1216, opened 2015-03-20
log "currently converging"


radix:
While we have the "currently converging" lock set, we should log its current value on every converge-all-groups cycle.

radix:
This was fixed by #1297.

Issue #1296, opened 2015-04-03
When starting convergence due to a watch trigger, don't call GetChildren


radix:
Since the children-watch we have in place on the /groups/divergent gives us the list of children, we don't need to call GetChildren to list them when we start convergence. This should only require a small refactoring to allow passing in our own list of children instead of listing them.

PR #1297, opened 2015-04-03
only `stat` dirty nodes that we really need to 

### LINKED ISSUE #1288: reduce unnecessary ZK interaction in convergence
As @manishtomar originally pointed out in the review of dirty-flags, `GetChildrenWithStats` is pretty much the wrong abstraction for us -- we should be listing children, filtering out the ones we don't care about, and then statting the ones we do care about.

And make sure that when we filter out the ones we don't care about, we filter out groups that aren't associated with our buckets _and_ groups that we are already converging.

radix:
Fixes #1288
Fixes: #1216 

`master` currently uses `GetChildrenWithStats` to find the dirty groups, which causes a bunch of `stat` calls that we don't need. These fall into two categories:

1. stats on dirty nodes that aren't allocated to us according to the bucket set partitioner
2. stats on dirty nodes that ARE allocated to us, but which we're already converging on.

So now we just `GetChildren` instead of `GetChildrenWithStats`, and only `Stat` it after filtering out groups that aren't ours and aren't already converging.

To do the second part, I needed to de-encapsulate the "currently converging" set from the group_locks object. Now the Converger service has an attribute `self.currently_converging = Reference(pset())` instead of a `self.group_locks = make_lock_set()`. This *also* allows us to log the currently converging set easily, so I went ahead and did that, fixing #1216.

There's one more useful optimization to make that I haven't done yet: #1296.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -5,6 +5,7 @@ The top-level entry-points into this module are :obj:`ConvergenceStarter` and
 :obj:`Converger`.
 """
 
+import sys
 import time
 from functools import partial
 from hashlib import sha1
@@ -33,7 +34,7 @@ from otter.convergence.planning import plan
 from otter.models.intents import GetScalingGroupInfo, ModifyGroupState
 from otter.models.interface import NoSuchScalingGroupError
 from otter.util.fp import assoc_obj
-from otter.util.zk import CreateOrSet, DeleteNode, GetChildrenWithStats
+from otter.util.zk import CreateOrSet, DeleteNode, GetChildren, GetStat
 
 
 def server_to_json(server):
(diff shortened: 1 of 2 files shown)


radix:

otter/convergence/service.py
```diff
@@ -293,45 +279,39 @@ def get_my_divergent_groups(my_buckets, all_buckets):
     :returns: list of dicts, where each dict has ``tenant_id``,
         ``group_id``, and ``version`` keys.
```
radix: I realize I missed this documentation referring to `version` -- I've fixed it in an upcoming PR to refer to `dirty-flag`.

lvh:
LGTM!

PR #1298, opened 2015-04-03
be nicer about deleting the group lock after deleting a group

radix:
This implements my suggestion in https://github.com/rackerlabs/otter/issues/1192#issuecomment-89151150

I realize that the problem of deleting locks while they're still being used will be far less likely when Manish's work is done, but I think this is still a better strategy for lock cleanup when it comes to bad cases.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -51,6 +51,7 @@ from otter.util.config import config_value
 from otter.util.cqlbatch import Batch, batch
 from otter.util.deferredutils import with_lock
 from otter.util.hashkey import generate_capability, generate_key_str
+from otter.util.retry import repeating_interval, retry, retry_times
 
 
 LOCK_PATH = '/locks'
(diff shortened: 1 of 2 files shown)


radix:

radix:

otter/models/cass.py
```diff
@@ -1231,15 +1232,18 @@ def _delete_group():
             return d
 
         def _delete_lock_znode(result):
-            d = self.kz_client.delete(LOCK_PATH + '/' + self.uuid,
-                                      recursive=True)
```
lvh: Why can we get rid of the recursive here? ISTR there's nothing there, but...

otter/models/cass.py
```diff
@@ -1231,15 +1232,18 @@ def _delete_group():
...
+                can_retry=retry_times(60),
+                next_interval=repeating_interval(5),
+                clock=self.reactor,
+            )
             d.addErrback(
                 lambda f: self.log.msg(
                     "Error cleaning up lock path (when deleting group)",
                     exc=f.value,
                     otter_msg_type="ignore-delete-lock-error"))
-            return d
```
lvh: The asynchrony here (well, not waiting until deleted) is intentional, right?

radix: Yep

lvh:
Sorry about those gnarly mock tests. LGTM assuming my questions pan out. Thanks for working on this!

otter/models/cass.py
radix: There _is_ stuff inside of these lock directories, and the problem is that deleting them recursively while someone else is trying to acquire the lock (or while it's held) can wreak havoc. There is content inside the lock directory only while the lock is being acquired or held.

Deleting it with recursive=False means that, if the lock is in use, the deletion will fail, which is why I added the 5 minutes of retries every 5 seconds -- this is very likely to delete the lock given our usage patterns, but will eventually give up and simply leave it there.

This PR is basically a stopgap to introducing a more persistent lock-cleanup solution based on the idea of a "child reaper" a la Curator (mentioned in the issue associated with this PR). These "child reapers" act very similar to this code, but they do it for all entries in a directory -- i.e., they list all children and constantly try to delete them non-recursively. This is pretty much the only way to definitely keep a directory full of locks clean, according to my discussion with ZK community members.

lvh: OK, awesome, sounds good.

Issue #1203, opened 2015-03-16
test convergence recreating servers deleted out of band


radix:
After otter builds a server in a group (and that convergence cycle finished successfully), if the server is deleted and convergence is somehow triggered again, otter should rebuild the server.

manishtomar:
This basically tests #881

radix:
@manishtomar - actually #881 is about load balancers, so that would be a separate test.

manishtomar:
Yeah. You are right. Created #1306 to track CLB issue.

PR #1300, opened 2015-04-03
First convergence-specific test I'm aware of.

sam-falvo:
This test aims to exercise the case where a user deletes a Nova server
from a scaling group out-of-band.

It's not finished yet, but it runs through to completion and doesn't
mess anything up.

It works by creating a scaling group with N_SERVERS (constant in the
source code).  It then deletes half of those servers through Nova's API.
What comes next is a subject for a future PR.

This is relevant to #1203.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -23,6 +23,53 @@ class BreakLoopException(Exception):
     """This serves to break out of a `retry_and_timeout` loop."""
 
 
+def create_scaling_group_dict(
+    image_ref=None, flavor_ref=None, min_entities=0, name=None
+):
+    """This function returns a dictionary containing a scaling group's JSON
+    payload.  Note: this function does NOT create a scaling group.
+
+    :param str image_ref: An OpenStack image reference ID (typically a UUID).
+        If not provided, the content of the AS_IMAGE_REF environment variable
+        will be taken as default.  If that doesn't exist, "" will be used.
+    :param str flavor_ref: As with image_ref above, but for the launch config's
+        flavor setting.
+    :param int min_entities: The minimum number of servers to bring up when
+        the scaling group is eventually created or operating.  If not
+        specified, 0 is assumed.
+    :param str name: The scaling group name.  If not provided, a default is
+        chosen.
+    :return: A dictionary containing a scaling group JSON descriptor.  Inside,
+        it will contain a default launch config with the provided (or assumed)
+        flavor and image IDs.
+    """
+
+    if not image_ref:
+        image_ref = os.environ['AS_IMAGE_REF']
+    if not flavor_ref:
+        flavor_ref = os.environ['AS_FLAVOR_REF']
+    if not name:
+        name = "automatically-generated-test-configuration"
+
+    return {
+        "launchConfiguration": {
+            "type": "launch_server",
+            "args": {
+                "server": {
+                    "flavorRef": flavor_ref,
+                    "imageRef": image_ref,
+                }
+            }
+        },
+        "groupConfiguration": {
+            "name": name,
+            "cooldown": 0,
+            "minEntities": min_entities,
+        },
+        "scalingPolicies": [],
+    }
+
+
 @attributes([
     Attribute('group_config', instance_of=dict),
     Attribute('pool', default_value=None),
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -3,6 +3,7 @@
 from __future__ import print_function
 
 import json
+import os
 
 from characteristic import Attribute, attributes
 
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -18,6 +18,7 @@ from otter.integration.lib.autoscale import (
     BreakLoopException,
     ScalingGroup,
     ScalingPolicy,
+    create_scaling_group_dict,
 )
 from otter.integration.lib.identity import IdentityV2
 from otter.integration.lib.resources import TestResources
@@ -77,23 +78,10 @@ class TestConvergence(unittest.TestCase):
 
         rcs = TestResources()
 
-        scaling_group_body = {
-            "launchConfiguration": {
-                "type": "launch_server",
-                "args": {
-                    "server": {
-                        "flavorRef": flavor_ref,
-                        "imageRef": image_ref,
-                    }
-                }
-            },
-            "groupConfiguration": {
-                "name": "converger-test-configuration",
-                "cooldown": 0,
-                "minEntities": N_SERVERS,
-            },
-            "scalingPolicies": [],
-        }
+        scaling_group_body = create_scaling_group_dict(
+            image_ref=image_ref, flavor_ref=flavor_ref,
+            min_entities=N_SERVERS
+        )
 
         self.scaling_group = ScalingGroup(
             group_config=scaling_group_body,
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -53,9 +53,6 @@ def find_end_points(rcs):
         rcs.endpoints["otter"] = 'http://localhost:9000/v1.0/{0}'.format(
             rcs.access['access']['token']['tenant']['id'])
 
-    rcs.endpoints["loadbalancers"] = auth.public_endpoint_url(
-        sc, "cloudLoadBalancers", region
-    )
     rcs.endpoints["nova"] = auth.public_endpoint_url(
         sc, "cloudServersOpenStack", region
     )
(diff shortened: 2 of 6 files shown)


stretchservers:

sam-falvo:
So, I'm getting this:

./otter/integration/tests/test_convergence.py:92:1: C901 'TestConvergence.test_reaction_to_oob_server_deletion'

in the build output.  Not sure how to address this.

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,228 @@
...
+                return (
+                    treq.delete(
+                        "{}/servers/{}".format(str(rcs.endpoints["nova"]), i),
+                        headers=headers(str(rcs.token)),
+                        pool=self.pool
+                    ).addCallback(check_success, [204])
+                    .addCallback(lambda _: rcs)
+                )
+
+            deferreds = map(lambda i: delete_server_by_id(i), ids)
```
manishtomar: You could do `map(delete_server_by_id, ids)` directly.

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,212 @@
...
+
+username = os.environ['AS_USERNAME']
+password = os.environ['AS_PASSWORD']
+endpoint = os.environ['AS_IDENTITY']
+flavor_ref = os.environ['AS_FLAVOR_REF']
+image_ref = os.environ['AS_IMAGE_REF']
+region = os.environ['AS_REGION']
+
+
+def find_end_points(rcs):
```
manishtomar: This is not convergence specific and should probably belong in `otter.integration.lib.identity`. Also, non-blocking: is it possible for `self.identity.authenticate_user` to return `TestResources` instead of passing it and mutating it in this function?

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,212 @@
...
+        eight servers, then use Nova to delete four of them.  We should be able
+        to see, over time, more servers coming into existence to replace those
+        deleted.
+        """
+
+        N_SERVERS = 4
+
+        rcs = TestResources()
+
+        scaling_group_body = {
```
manishtomar: I expect to be reused in many tests. Could it go somewhere in `otter.integration.lib.autoscale`?

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,212 @@
...
+        # same.  So just return the 1st, which is just our rcs value.
+        return gatherResults(deferreds).addCallback(lambda rslts: rslts[0])
+
+    def _wait_for_autoscale_to_catch_up(self, _, rcs, timeout=60, period=1):
+        """Wait for the converger to recognize the reality of this tenant's
+        situation and reflect it in the scaling group state accordingly.
+        """
+
+        def check(state):
+            if state[0] != 200:
```
manishtomar: Nitpick: Could you take arguments like `def check((code, response))` and check for code or
```python
code, response = state
if code != 200:
```
This makes it easier to understand.

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,212 @@
...
+                raise BreakLoopException(
```
manishtomar: Shouldn't this be a `ScalingGroupNotFound` kind of exception instead of generic `BreakLoopException` since its representation that situation?

manishtomar:
I was wondering if you had thought of writing a simpler test which deletes a single server, executes scaling policy of +1 and expects "old + 2" servers instead of deleting half and expecting old + 1?

otter/integration/tests/test_convergence.py
```diff
@@ -0,0 +1,212 @@
...
+            endpoint=endpoint, pool=self.pool
+        )
+
+    def tearDown(self):
+        """Destroy the HTTP connection pool, so that we close the reactor
+        cleanly.
+        """
+
+        def _check_fds(_):
+            fds = set(reactor.getReaders() + reactor.getReaders())
```
kivattik: Should this be set(getReaders() + getWriters())?

otter/integration/tests/test_convergence.py
kivattik: Unless we are specifically checking for a 404, can we conclude the scaling group disappeared? In this case, we might keep the BreakLoop and change the message to a more generic "Unable to get scaling group state, no longer waiting."

otter/integration/tests/test_convergence.py
kivattik: Are you suggesting creating the TestResources object within the identity function rather than the test case? If so, I think it is better to keep the creation in the test case both for the logical association of the resources with the specific case and because it is consistent with the many other functions that take the rcs parameter and mutate it.

sam-falvo: I'm going to object, because find_end_points() is a function which is test specific.  You'll note that the endpoints sought by this implementation differ from the other one you've referenced.

otter/integration/tests/test_convergence.py
sam-falvo: @manishtomar I can, but what would it buy?  The BreakLoopException is already defined in a central location, and is serving its purpose as a means to break out of the retry loop.

@kivattik Yes, 404 is expressly checked for; you just don't see it here.  If you look in the definition of `get_scaling_group_state()`, you'll see that it checks expressly for 200 and 404 response codes.  All other response codes will raise an exception.

kivattik:
Just to get things moving, I think that the two comments that should be addressed before merging are the  getReaders()/Writers() and manisht's comment on parameterizing the check for status code.

sam-falvo:
@manishtomar Yes, I thought about it; but I believe this test is more comprehensive.  Restoring a single server is relatively easy; restoring multiple servers is a genuine test of convergence's capabilities.

otter/integration/tests/test_convergence.py
sam-falvo: Fixed in 5ae1b7d

otter/integration/tests/test_convergence.py
sam-falvo: You mean add the dictionary, or add a constructor?

I'm hesitant to add the dictionary because each test will need specializations on this structure.

If a constructor, I'm not sure what that saves over just using hash notation.

otter/integration/tests/test_convergence.py
sam-falvo: @kivattik set(..) bug fixed in 15b4207

otter/integration/tests/test_convergence.py
sam-falvo: Fixed in c586319, but I'm not sure what it actually buys.  The total LOC isn't much smaller in retrospect.  Spent several hours debugging this, you better appreciate it!!  ;-)

otter/integration/tests/test_convergence.py
manishtomar: For now, I would just suggest a function `otter.integration.lib.autoscale.scaling_group_body_template()` that returns the `dict`. Ideally, it would be nice to have `create_scaling_group(img_ref=None, flavor_ref=None)` function that defaults to configured (via env or config file) image ref, flavor ref and other group properties if it is not provided.

manishtomar:
> @manishtomar Yes, I thought about it; but I believe this test is more comprehensive.

I agree. Although I would argue that this does not bring something more to the table since same code path is exercised. Please feel free to disagree with me as this is developer in me talking :) and I am sure you've better reasoning to do this. 

One more reason I was thinking to do a simpler test is because you had talked about not able writing the test in time you would've liked. So, it might be better to write simpler tests first and move to complicated ones later on. This is just a thought. 

Please fix the build and I will take a look at it again. Thanks :)

otter/integration/tests/test_convergence.py
sam-falvo: Fixed in 5bda2ee

sam-falvo:
To answer your question, I started out with N_SERVERS=2, specifically because that was the simpler case (and meets your definition above).  I increased it to N_SERVERS=4 once I was comfortable that it would work with more sophisticated cases.

otter/integration/lib/autoscale.py
```diff
@@ -23,6 +24,53 @@ class BreakLoopException(Exception):
...
+        flavor setting.
+    :param int min_entities: The minimum number of servers to bring up when
+        the scaling group is eventually created or operating.  If not
+        specified, 0 is assumed.
+    :param str name: The scaling group name.  If not provided, a default is
+        chosen.
+    :return: A dictionary containing a scaling group JSON descriptor.  Inside,
+        it will contain a default launch config with the provided (or assumed)
+        flavor and image IDs.
+    """
```
manishtomar: +1. Nice docs!

PR #1301, opened 2015-04-05
mock_group tests refactor

manishtomar:
As per `group.modify_state` docs it is supposed to return None. So, changing the mock `modify_state` to return None instead of new state. Changed affected tests


radix:
LGTM :)

PR #1302, opened 2015-04-06
delete out of band test is mimic specific

### LINKED ISSUE #1228: test_system_user_delete_some_servers_out_of_band  is mimic specific?

manishtomar:
Closes #1228


lvh:
LGTM!

Issue #1215, opened 2015-03-20
clean up convergence if a group is removed from /groups/divergent


radix:
If a group ever gets wedged in a converging state, we may want to manually delete its dirty flag in /groups/divergent/{tenant_id}_{group_id}. This will prevent nodes from *starting* convergence, but it won't stop any current convergence processes. *edit: actually, I was wrong about this, see the next comment*

The converger service should notice if any of its currently-converging groups have been removed from /groups/divergent, and cancel its outstanding convergence process.

radix:
Er, while I think there is a bug here, the nature of it is wrong (I'm not sure if I just had a brainfart or if I changed the design of `currently_converging` at some point).

Anyway, deleting dirty flags *will* stop convergence, but since the group is left in currently_converging, any *further* attempts to converge the group will stall, since the node will refuse to start convergence on a node that it thinks it is already converging on.

radix:
No, no, there's not a bug, I have no idea what I'm even doing. I keep forgetting what the scope of the currently-converging lock is. :(

PR #1303, opened 2015-04-06
Remove the comment about bug #1215, which was invalid

radix:
a group's membership in currently_converging does not persist beyond a single cycle, so this bug was just persistent confusion on my part.


lvh:
LGTM!

PR #1307, opened 2015-04-08
scale down fix for mimic

manishtomar:
The boolean mimic config was incorrectly read as true everytime causing smaller timeout and interval in retry. Also, it looked like timeout was fixed to 60 seconds independent of mimic config. So, fixed reading
config and simpler code for scaling down timeout, interval.

### DIFF
--- a/autoscale_cloudcafe/autoscale/behaviors.py
+++ b/autoscale_cloudcafe/autoscale/behaviors.py
@@ -480,13 +480,11 @@ class AutoscaleBehaviors(BaseBehavior):
         if interval_time is None:
             interval_time = self.autoscale_config.interval_time
 
-        if time_scale:
+        if time_scale and self.autoscale_config.mimic:
             # scale time down if using mimic - no shorter than 1 second, though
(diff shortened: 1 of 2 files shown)


autoscale_cloudcafe/autoscale/behaviors.py
```diff
@@ -480,13 +480,11 @@ def retry(self, callable, timeout=None, interval_time=None,
         if interval_time is None:
             interval_time = self.autoscale_config.interval_time
 
-        if time_scale:
+        if time_scale and self.autoscale_config.mimic:
             # scale time down if using mimic - no shorter than 1 second, though
```
cyli: Since we are no longer doing `max(val, 1)`, the last half of this comment can probably be truncated.  :)

cyli:
LGTM, thank you for fixing this!  I should have caught this in the previous PR.

autoscale_cloudcafe/autoscale/behaviors.py
manishtomar: Maybe I should add that `timeout = max(timeout * scale_down_factor, 1)`?

PR #1309, opened 2015-04-09
error when capacity tries to go outside min/max

manishtomar:
When executing policy on convergence tenants, error is raised if no change in servers is detected. This allows 2 more tests to pass.


manishtomar:

radix:
LGTM

PR #1340, opened 2015-04-15
Scale up by 3, delete 1 OOB, then scale up by 2.

### LINKED ISSUE #1314: CATC5: Scale up after out of band delete
OOBD: Scale up after out of band delete, non constrained
	- Create a group with min N servers
	- Scale group to (N+ x) servers
	- Delete one of the servers out of band
	- Scale up by (y) servers
        - Expect N + x + y servers eventually (i.e. the deleted server is replaced and the y scale is executed on top of that.)

sam-falvo:
We expect 4 servers after this sequence of operations.

Fixes #1314 

Also replaces some legacy code in autoscale.py with easier-to-read, but semantically identical, code.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -24,6 +24,16 @@ class BreakLoopException(Exception):
     """This serves to break out of a `retry_and_timeout` loop."""
 
 
+def extract_active_ids(group_status):
+    """Extracts all server IDs from a scaling group's status report.
+
+    :param dict group_status: The successful result from
+        ``get_scaling_group_state``.
+    :result: A list of server IDs known to the scaling group.
+    """
+    return [obj['id'] for obj in group_status['group']['active']]
+
+
 def create_scaling_group_dict(
     image_ref=None, flavor_ref=None, min_entities=0, name=None
 ):
@@ -292,7 +302,7 @@ class ScalingGroup(object):
         else:
             report = (
                 "Scaling group failed to reflect {} servers removed."
-                .format(len(ids_deleted))
+                .format(len(removed_ids))
             )
 
         return retry_and_timeout(
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -195,7 +195,7 @@ class ScalingGroup(object):
         """
 
         def check((code, response)):
-            if code != 200:
+            if code == 404:
                 raise BreakLoopException("Scaling group not found.")
             servers_active = len(response["group"]["active"])
             if servers_active == servers_desired:
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -246,6 +246,63 @@ class ScalingGroup(object):
             .addCallback(record_results)
         )
 
+    def wait_for_deleted_id_removal(
+        self, removed_ids, rcs, timeout=60, period=1, total_servers=None
+    ):
+        """Wait for the scaling group to reflect the true state of the tenant's
+        server states.  Out-of-band server deletions or servers which
+        transition to an ERROR state should eventually be removed from the
+        scaling group's list of active servers.
+
+        :param list removed_ids: A list of server IDs that we should expect
+            the scaling group to realize are gone eventually.
+        :param TestResources rcs: The test resources necessary to invoke API
+            calls and manage state.
+        :param int timeout: The number of seconds to wait before giving up.
+        :param int period: The number of seconds between each check for ID
+            removal.
+        :param int total_servers: If provided, the total number of servers in
+            your scaling group.  This parameter is used for error reporting
+            purposes only.
+        :return: It'll return the value of ``rcs`` if successful.  An exception
+            will be raised otherwise, including timeout.
+        """
+
+        def check((code, response)):
+            if code == 404:
+                raise BreakLoopException(
+                    "Scaling group appears to have disappeared"
+                )
+
+            active_ids = extract_active_ids(response)
+            for deleted_id in removed_ids:
+                if deleted_id in active_ids:
+                    raise TransientRetryError()
+
+            return rcs
+
+        def poll():
+            return self.get_scaling_group_state(rcs).addCallback(check)
+
+        if total_servers:
+            report = (
+                "Scaling group failed to reflect {} of {} servers removed."
+                .format(len(removed_ids), total_servers)
+            )
+        else:
+            report = (
+                "Scaling group failed to reflect {} servers removed."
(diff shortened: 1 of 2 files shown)


stretchservers:

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +68,54 @@ def _check_fds(_):
             return deferLater(reactor, 0, _check_fds, None)
         return self.pool.closeCachedConnections().addBoth(_check_fds)
 
+    def test_reaction_to_oob_deletion_after_scale_up(self):
+        """Exercise out-of-band server deletion, but then scale up afterwards.
+        The goal is to spin up, say, three servers, then use Nova to delete
+        one of them directly, without Autoscale's knowledge.  Then we scale up
+        by, say, two servers.  If convergence is working as expected, we expect
+        four servers at the end.
```
radix: I think this is off:

3 servers running
-1 deleted OOB
+2 from policy

Convergence should fix the deletion by replacing it, _and_ add 2 for the policy execution, to arrive at 5, not 4.

otter/integration/tests/test_convergence.py
```diff
@@ -129,6 +178,16 @@ def _choose_half_the_servers(self, (code, response)):
         self.n_killed = self.n_servers / 2
         return ids[:self.n_killed]
 
+    def _choose_random_servers(self, (code, response), n):
+        """Selects ``n`` randomly selected servers from those returned by the
+        ``get_scaling_group_state`` function.
+        """
+
+        if code != 200:
+            raise Exception("Got 404; dude, where's my scaling group?")
```
radix: I like the message :-), but the 404 part is inaccurate; the real code should be interpolated here.

otter/integration/tests/test_convergence.py
```diff
@@ -129,6 +178,16 @@ def _choose_half_the_servers(self, (code, response)):
...
+        ids = map(lambda obj: obj['id'], response['group']['active'])
```
radix: Just a python style thing: list comprehensions are often more concise than `map` when specifically used in conjunction with `lambda`. I am all for maps and lambdas, but consider:

```
[obj['id'] for obj in response['group']['active']
```
vs
```
map(lambda obj: obj['id'], response['group']['active'])
```

Please just consider this comment as informative. Conciseness isn't everything, anyway.

radix:
LGTM if it actually passes. (I would try it myself but I don't know how to run these new tests yet -- I can try to set it up tomorrow)

otter/integration/tests/test_convergence.py
sam-falvo: The 404 is accurate; if you look at the code for get_scaling_group_status in the ScalingGroup class, you'll see that it check_results([200, 404]).  200 and 404 are the only "valid" codes it returns.  Anything else is an exception.

So, this issue has come up repeatedly.  I'm not sure why, but it's very annoying.  What can I do to avoid this in the future?

radix: I see. I would suggest `if code == 404` for clarity, then.

sam-falvo: OK, I just pushed patches to the PR.  Github refuses to accept that I've addressed this issue.  :(  But I have.

otter/integration/lib/autoscale.py
```diff
@@ -194,10 +204,10 @@ def wait_for_N_servers(
             Otherwise, an exception is raised.
         """
 
-        def check(state):
-            if state[0] != 200:
+        def check((code, response)):
```
lvh: Tuple parameter unpacking is deprecated, and doesn't work in 3.x+; see [PEP 3113](http://legacy.python.org/dev/peps/pep-3113/). Suggestion: keep the "state" parameter, but destructure on the next line:

```python
def check(state):
    code, response = state
    ....
```

manishtomar: Oh. Didn't know that. Sorry @sam-falvo for suggesting this. You should probably change to suggested form.

cyli:
Non-blocking nitpick - `test_reaction_to_oob_deletion_after_scale_up`: wondering if renaming this to `test_reaction_to_oob_deletion_then_scale_up` may be clearer?  I thought that the deletion was supposed to happen after the scale up by the test name, although the docstring was pretty clear.

Also, non-blocking, was just curious - `test_reaction_to_oob_server_deletion` seems to be very similar (both create a group with a min number of servers, delete some, scale up, and ensure that the number is equal to the min + scaling change).  The docstring for `test_reaction_to_oob_server_deletion` does not mention a scale-up, and I notice that the only difference seems to be that it doesn't wait until the servers are active before deleting.  I'm wondering if the scale up is because the test fails in mimic without it?

otter/integration/lib/autoscale.py
lvh: Don't get me wrong, the destructuring is way better than `state[n]`; it should probably just not happen in the argspec :)

cyli:
Tested this, works fine except force-delete is still broken in convergence, so getting errors with cleanup.  So other than @lvh's comment LGTM (unless you want to address my naming nitpick :)).

otter/integration/lib/autoscale.py
sam-falvo: Fixed in ad7412f

sam-falvo:
Name change to test_reaction_to_oob_deletion_then_scale_up fixed in 9c8ca64

sam-falvo:
Regarding oob_server_deletion function, it does wait for the number of servers to come up before deleting:

```
            .addCallback(
                self.scaling_group.wait_for_N_servers,
                N_SERVERS, timeout=1800
            ).addCallback(self.scaling_group.get_scaling_group_state)
```

Am I missing something?

The reason I scale up is because the /converge endpoint doesn't exist yet.  I needed something that was as close to idempotent as I could find which would cause the converger to (eventually) kick in for this tenant.

The test exercises recognition (in a reasonable period of time) of _multiple_ servers being deleted, not just one.

cyli:
@sam-falvo Ah ok, sorry, I missed that.  Ok - would it make sense then to name them very similar things/have them have almost the same docstring, and then also mention what the difference is between them?  They looked like otherwise the same test to me except for the numbers.  (For that matter, could the code be factored out since they do very similar things?)

sam-falvo:
I'd like not to do that.  Let's see if things are comparable after I replace the scaling operation in one of the tests with a POST(GET(scaling group state), scaling group state) operation to kick off the converger without altering the number of servers.

cyli:
Ok - after IRL discussion I see the difference between the two tests now.

PR #1342, opened 2015-04-16
Convergence tests: Scale to clb max after oob delete

sam-falvo:
This exercises two variants of the core test logic:

1) When SG's max_entities == CLB's max
2) When SG's max_entities > CLB's max

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -25,7 +25,8 @@ class BreakLoopException(Exception):
 
 
 def create_scaling_group_dict(
-    image_ref=None, flavor_ref=None, min_entities=0, name=None
+    image_ref=None, flavor_ref=None, min_entities=0, name=None,
+    use_lbs=None
 ):
     """This function returns a dictionary containing a scaling group's JSON
     payload.  Note: this function does NOT create a scaling group.
@@ -40,6 +41,11 @@ def create_scaling_group_dict(
         specified, 0 is assumed.
     :param str name: The scaling group name.  If not provided, a default is
         chosen.
+    :param list use_lbs: Specifies a list of one or more cloud or RackConnect
+        load balancer JSON *dictionary* objects.  These are *not* instances of
+        ``otter.lib.CloudLoadBalancer``.  However, you can get the dicts by
+        invoking the o.l.CLB.scaling_group_spec() method on such objects.  If
+        not given, no load balancers will be used.
     :return: A dictionary containing a scaling group JSON descriptor.  Inside,
         it will contain a default launch config with the provided (or assumed)
         flavor and image IDs.
@@ -52,7 +58,7 @@ def create_scaling_group_dict(
     if not name:
         name = "automatically-generated-test-configuration"
 
-    return {
+    obj = {
         "launchConfiguration": {
             "type": "launch_server",
             "args": {
@@ -69,6 +75,11 @@ def create_scaling_group_dict(
         },
         "scalingPolicies": [],
     }
+
+    if use_lbs:
+        obj["launchConfiguration"]["args"]["loadBalancers"] = use_lbs
+
+    return obj
 
 
 @attributes([
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -2,9 +2,9 @@
 
 from __future__ import print_function
 
+import datetime
 import json
 import os
-import datetime
 
 from characteristic import Attribute, attributes
 
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -174,10 +174,12 @@ class TestConvergence(unittest.TestCase):
                 .addCallback(self._choose_random_servers, 2)
                 .addCallback(self._delete_those_servers, rcs)
                 .addCallback(self.second_scaling_policy.execute)
+                .addCallback(lambda _: self.removed_ids)
                 .addCallback(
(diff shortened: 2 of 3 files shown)


stretchservers:

sam-falvo:
Travis CI is puking pretty hard.

jenkins please run unit tests

sam-falvo:
jenkins please run unit tests

otter/integration/lib/autoscale.py
```diff
@@ -317,25 +332,50 @@ def poll():
...
+            change_type = "changePercent"
+            change_factor = self.scale_percent
+
         self.policy = [{
-            "name": "integration-test-policy",
+            "name": name_time,
             "cooldown": 0,
             "type": "webhook",
-            "change": self.scale_by
+            change_type: change_factor
```
cyli: Non-blocking - might be easier to have `change_type` and `change_factor` as attributes instead, since having one override another and maybe taking priority seems more confusing.  I can file a ticket if you agree, or not if you don't agree.

otter/integration/tests/test_convergence.py
```diff
@@ -62,6 +63,130 @@ def _check_fds(_):
...
+            #
+            # - If Otter converges, it's entirely possible for it to notice
+            # that two servers are gone, for it to attempt to provision
+            # replacements, and then provision the third and final server, all
+            # before we get a response back from
+            # scaling_group.get_scaling_group_state.
+            #
+            # So, we need an intermediate step that *guarantees* our test the
+            # ability to inspect Otter's behavior.  This is that step.
+            self.second_scaling_policy = ScalingPolicy(
```
cyli: Non-blocking - apologies, I don't think I understand this explanation very well - wouldn't scaling up by 1, waiting for those 2 IDs to be removed first, then waiting for there to be 25 servers be sufficient?  If otter does not have convergence, those 2 IDs will still be in the active list and never removed.

That's what `test_reaction_to_oob_deletion_then_scale_up` does, right?

Not that this breaks the test - was just wondering if it was necessary.  Again, we can file a cleanup ticket to remove the extra call, if this is a valid comment.

otter/integration/tests/test_convergence.py
```diff
@@ -62,6 +63,130 @@ def _check_fds(_):
...
+        return self._perform_oobd_clb_test(50)
+    test_scaling_to_clb_max_after_oob_delete_type2.timeout = 1800
+
+    def _perform_oobd_clb_test(self, scaling_group_max_entities):
+        rcs = TestResources()
+
+        def create_clb_first():
+            self.clb = CloudLoadBalancer(pool=self.pool)
+            return (
+                self.identity.authenticate_user(rcs)
```
cyli: Should this specify `tenant_id="000000"`, or is that a different PR?

otter/integration/tests/test_convergence.py
```diff
@@ -62,6 +63,130 @@ def _check_fds(_):
...
+    test_scaling_to_clb_max_after_oob_delete_type1.timeout = 1800
+
+    def test_scaling_to_clb_max_after_oob_delete_type2(self):
+        """This test starts with a scaling group with no servers.  We scale up
+        to 24 servers, but after that's done, we delete 2 directly through
+        Nova.  After that, we scale up once more by 1 server, thus max'ing out
+        the CLB's ports.  We expect that the group will return to 25 servers,
+        and does not overshoot in the process.
+
+        Further, we want to make sure the deleted servers are removed from the
```
cyli: Is this going to happen in a different PR?  Should an issue be filed to address it in these two tests?

cyli:
I'd love answers to the above questions, but other than that this LGTM.  The tests do not pass for me, but I think this is a bug in convergence code - I'm getting mimic reflecting 24 servers + 24 load balancer nodes, but autoscale state saying that there are 21 active servers and 3 pending servers, which means that active is not being written to one last time.

The test logic looks right, again with the exception of the non-blocking questions above.

otter/integration/lib/autoscale.py
sam-falvo: I'm not sure what you mean by making it an attribute.  My guess is that you mean changing this logic to just use `self.change_type` and `self.change_factor` directly and eliminating the whole if/elif block?

otter/integration/tests/test_convergence.py
sam-falvo: Yes, let's make this and similar tasks a separate clean-up ticket.  :)

otter/integration/tests/test_convergence.py
sam-falvo: Can you clarify what your pronoun "this" refers to?

otter/integration/tests/test_convergence.py
sam-falvo: Separate PR.  :)

sam-falvo:
Due to time constraints, I'm going to land the PR, but feel free to follow up with additional comments in new tickets, please.  Thanks!

otter/integration/tests/test_convergence.py
cyli: Sorry, the comment about making sure deleted servers are removed from the load balancers.

otter/integration/lib/autoscale.py
cyli: Yep, that's what I mean.  :)

cyli: https://github.com/rackerlabs/otter/issues/1358

otter/integration/tests/test_convergence.py
cyli: https://github.com/rackerlabs/otter/issues/1359

Issue #1223, opened 2015-03-21
some webhook keys disappearing


manishtomar:
I noticed few instances of webhook keys being there in `policy_webhooks` but not in `webhook_keys`. From the logs, the webhook key was created *and* used successfully on Feb 23rd but for some reason not available today. I have currently added these using `python load_cql.py --webhook-migrate`. We might need to check again sometime in future by `python load_cql.py --webhook-index-only` and see if it still happens.

manishtomar:
Correct fix in #1345

cyli:
This seems to have been fixed.  Can we close?

manishtomar:
Yes. I guess I forgot to close with last comment. Closing.

PR #1343, opened 2015-04-20
missing webhook keys mitigation - Issue #1223

manishtomar:
Mitigating #1223. Getting webhook key from policy_webhooks if it is not found in webhook_keys table. This is done temporarily until the root cause is known. It also logs error when such a situation arises. Did some refactoring:
   - implemented getting using effect. the main function only performs it 
   - moved `get_cql_dispatcher` to cass.py itself as their was cyclic import issues


manishtomar:
jenkins please run unit tests

manishtomar:
Closing this as the root cause has *just* now been discovered!

PR #1348, opened 2015-04-21
Use convergence prepare launch config

cyli:
Just eliminate the worker-specific code for generating metadata and just use convergence's `prepare_server_launch_config`, partially so we can start getting reach to maybe switch over to the new metadata value, partially so we can test scrubbing convergence specific metadata.

Scrubbing convergence-specific metadata is important for disowning servers as part of #1238.


cyli:

radix:
Nice! Glad to see negative diffstats :D

PR #1351, opened 2015-04-22
Move performer and intent from supervisor to worker_intents

cyli:
As suggested by @radix to avoid circular imports when making a dispatcher that handles `EvictServerFromScalingGroup`

This is just a move - no code changes except for imports from the controller.


radix:
LGTM :) thanks @cyli

cyli:
@radix:  thanks for the suggestion!

PR #1352, opened 2015-04-22
Add eviction dispatcher

cyli:
Make `get_full_dispatcher` return a dispatcher which can handle `EvictServerFromScalingGroup` intents.

### DIFF
--- a/otter/effect_dispatcher.py
+++ b/otter/effect_dispatcher.py
@@ -21,7 +21,7 @@ from .models.intents import get_model_dispatcher
 from .util.pure_http import Request, perform_request
 from .util.retry import Retry, perform_retry
 from .util.zk import get_zk_dispatcher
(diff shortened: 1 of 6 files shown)


otter/effect_dispatcher.py
```diff
@@ -84,3 +86,15 @@ def get_cql_dispatcher(reactor, connection):
             CQLQueryExecute: partial(perform_cql_query, connection)
         })
     ])
+
+
+def get_eviction_dispatcher(supervisor):
```
radix: To follow the model elsewhere I'd suggest this be moved to worker_intents.py. I think it's fine to leave the name the same for now since we don't know what else would go in here yet, we can rename it later if we decide to add more worker intents to it.

radix:
Other than the location nit, this LGTM :)

PR #1347, opened 2015-04-21
CATC - 6: Unconstrained out of band delete followed by scale

kivattik:
fixes #1315

PR #1353, opened 2015-04-22
Lk improve catc6

kivattik:
Improved the completeness of the scale down after OOB delete test by varying the delete and scale amounts.

Depends on #1347

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -166,7 +166,8 @@ class ScalingGroup(object):
             ),
             headers=headers(str(rcs.token)),
             pool=self.pool
-        ).addCallback(check_success, [204, 404]))
+        ).addCallback(check_success, [204, 404, 403]))  # HACKITY HACK - Set to
+        # 403 until force delete is fixed in convergence
 
     def get_scaling_group_state(self, rcs):
         """Retrieve the state of the scaling group.
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -182,7 +182,7 @@ class TestConvergence(unittest.TestCase):
         )
     test_reaction_to_oob_server_deletion.timeout = 1800
 
-    def test_scale_down_after_oobd_non_constrained(self):
+    def test_scale_down_after_oobd_non_constrained_z_lessthan_y(self):
         """
         Validate the following edge case:
         - When scaling down after an out of band delete (OOBD) that is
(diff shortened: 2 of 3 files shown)


otter/integration/tests/test_convergence.py
```diff
@@ -183,6 +344,25 @@ def _choose_half_the_servers(self, (code, response)):
...
+        provided, all active servers will be returned.
+
+        If the number of servers requested is less than the number active,
+        an error will be raised.
+        """
+
+        if code != 200:
+            raise Exception("Not 200!; Asking for the state didn't work.")
+        ids = extract_active_ids(response)
+        self.active_servers = len(ids)
```
cyli: Out of curiosity, since, this seems to be the same kind of function as `_choose_random_servers` and `_choose_half_the_servers`, any reason this is recording in a different attribute name?  (Just wondering because accessing different attributes based on which choice function was used is a little hard to remember).

Also wondering if grabbing the first N are significant?  There is no contract that specifies what order the servers in the active list are in, so wondering if `_choose_random_servers` would be sufficient?

otter/integration/tests/test_convergence.py
```diff
@@ -321,20 +321,93 @@ def test_scale_down_after_oobd_non_constrained(self):
...
+        """
         rcs = TestResources()
 
+        N = 2
+        x = 7
+        z = 3
+        y = -3  # Scale down 3 servers
+
+        return self._scale_down_after_oobd_non_constrained_param(
+            rcs, min_servers=N, set_to_servers=x, oobd_servers=z,
```
cyli: Nitpick on terminology and variable naming - if we are setting desired capacity to x, then shouldn't the docstring be:

`Scale group to x servers (where x > N)`

instead of 

`Scale group to (N + x) servers`

otter/integration/tests/test_convergence.py
```diff
@@ -321,20 +321,93 @@ def test_scale_down_after_oobd_non_constrained(self):
...
+                - z > |y|
+
+        """
+
+        rcs = TestResources()
+
+        N = 2
+        x = 7
+        z = 3
+        y = -2  # Scale down 3 servers
```
cyli: Comment nitpick - this is scaling down by 2 - but wondering if this comment is necessary?  Or if so, should all of them be commented?

otter/integration/lib/autoscale.py
```diff
@@ -166,7 +166,8 @@ def delete_scaling_group(self, rcs):
             ),
             headers=headers(str(rcs.token)),
             pool=self.pool
-        ).addCallback(check_success, [204, 404]))
+        ).addCallback(check_success, [204, 404, 403]))  # HACKITY HACK - Set to
+        # 403 until force delete is fixed in convergence
```
cyli: Cool - nice to see this make the tests pass.  :)  #1179 is the issue btw, in case that's useful

cyli:
Aside from the doc comments, LGTM.  :)

otter/integration/lib/autoscale.py
kivattik: I've added the removal of the 403 to the force delete PR list of tasks so that it isn't forgotten.

otter/integration/tests/test_convergence.py
kivattik: Well, the eventual total is N + x servers, but that is confusing, so I'll clarify a bit.

kivattik:
Merging after the doc comments were fixed.

PR #1355, opened 2015-04-22
More retries in steps

### LINKED ISSUE #1171: RemoveNodesFromCLB should return success, [reasons]
Right now there's a function `_clb_check_bulk_delete` which should work like `_rcv3_check_bulk_delete` in that it should return either an `Effect` or `StepResult.SUCCESS, []`

radix:
- AddNodesToCLB now returns RETRY on success, so that cached active servers can be updated
- ditto adding nodes to RCv3. There's still one case that I'm not sure about (see comment)
- make RemoveNodesFromCLB return step results at all (it hadn't been updated to the new step result API). It returns SUCCESS just because a remove from LB should always immediately be followed by a remove from server (or if not, the server isn't ours anyway?) This fixes #1171.

The fact that pretty much all SUCCESSes are gradually becoming RETRY does make me wonder if we should move this "retry to update active cache" *back* to service.py. but there are a few cases where we don't need to? more analysis necessary, but I think for now we should make this change so that stuff works.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -331,7 +331,9 @@ class AddNodesToCLB(object):
                 _check_clb_422(_CLB_DUPLICATE_NODES_PATTERN)))
 
         def report_success(result):
-            return StepResult.SUCCESS, []
+            return StepResult.RETRY, [
+                'must re-gather after adding to CLB in order to update the '
+                'active cache']
 
         def report_api_failure(result):
             """
@@ -384,8 +386,9 @@ class RemoveNodesFromCLB(object):
                                _CLB_DELETED_PATTERN)))
         # 400 means that there are some nodes that are no longer on the
         # load balancer.  Parse them out and try again.
-        return eff.on(partial(
+        eff = eff.on(partial(
             _clb_check_bulk_delete, self.lb_id, self.node_ids))
+        return eff.on(lambda r: (StepResult.SUCCESS, []))
 
 
 def _clb_check_bulk_delete(lb_id, attempted_nodes, result):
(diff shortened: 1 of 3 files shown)


otter/convergence/steps.py
```diff
@@ -575,6 +580,8 @@ def _rcv3_check_bulk_add(attempted_pairs, result):
         next_step = BulkAddToRCv3(lb_node_pairs=to_retry)
         return next_step.as_effect()
     else:
+        # It's unclear when this condition is reached. Should we really be
+        # returning SUCCESS?
```
cyli: I think the motivation for this was that if some other error happens, someone on the RCv3 team will manually fix it?  But yeah, wondering if we should return Retry instead.

cyli:
Aside from travis linting issues, this LGTM - better to get the tests passing ASAP :)  

I agree with you that we may want to revisit something else always repopulating the active cache, or when to re-converge.  I'm not sure about the steps having to know whether they are the last steps in a process right now.  Or for instance that the remove-from-lb know that they are not the last step in the process, but they know that the planner happens to be implemented in such a way that if a server is removed from the last load balancer, the server also gets deleted (hence no need for retry).

But anyway, once linting is fixed, :shipit: :D

otter/convergence/steps.py
radix: Okay, now that I read the code closer, it definitely seems strange to treat it as a SUCCESS -- this is in the 409 case only.  We should perhaps even consider it a FAILURE.

radix:
@cyli I totally agree with your point about the strange knowledge that the steps have of the context in which they're used -- that's exactly what all this RETRY stuff is about. Anyway, thanks for the review :)

otter/convergence/steps.py
radix: filed #1356 about this.

PR #1357, opened 2015-04-22
Pass params through during request concretization

radix:
`concretize_service_request` was not passing params through to the underlying `pure_http.Request`. The CLB bulk delete step is relying on this.


cyli:
Nice catch!  LGTM

PR #1361, opened 2015-04-22
Add tenant override for convergence.

sam-falvo:
To use:

$ trial test.foo.bar  --  without Convergence
$ AS_CONVERGENCE_TENANT=000000 trial test.foo.bar  --  with convergence


stretchservers:

cyli:
+1

Issue #1363, opened 2015-04-23
deleting column index and query


manishtomar:
As mentioned in https://github.com/rackerlabs/otter/pull/1261#issuecomment-94983721, use `deleting` column in query directly to filter out deleting groups. To do that, index should be created on that column. Hence following tasks in order:
- [x] Run migration script that inserts `false` on all group's `deleting` column
- [x] Create index on `deleting` column
- [x] Use `deleting` column in code to filter non-deleting groups

manishtomar:
This was deployed today. Closing.

PR #1364, opened 2015-04-23
Deleting column and index in query

manishtomar:
3rd task of #1363.

Certain refactorings:
- Taking groups limit as argument to constructor instead of reading from config in `CassScalingGroupCollection` and simplified checking limits in `create_group`
- Moved `CassScalingGroup.get_policy` tests to a separate class
- Removed unwanted and changed some cass tests (Cannot help refactoring these tests every time I modify it) 

NOTE: This code must be deployed AFTER creating index in prod, otherwise all queries involving `deleting` column will fail.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -23,7 +23,6 @@ from pyrsistent import freeze
 
 from silverberg.client import ConsistencyLevel
 
-from toolz.curried import filter
 from toolz.dicttoolz import keymap
 
 from twisted.internet import defer
@@ -207,7 +206,7 @@ _cql_del_on_key = 'DELETE FROM {cf} WHERE "webhookKey"=:{name}webhookKey'
 
 _cql_count_for_tenant = (
     'SELECT COUNT(*) FROM {cf} '
-    'WHERE "tenantId"=:tenantId AND deleting=false;')
+    'WHERE "tenantId"=:tenantId {deleting};')
 _cql_count_for_policy = (
     'SELECT COUNT(*) FROM {cf} '
     'WHERE "tenantId" = :tenantId AND "groupId" = :groupId '
@@ -947,8 +946,8 @@ class CassScalingGroup(object):
             query = _cql_view_policy.format(cf=self.policies_table)
             d = self.connection.execute(query,
                                         {"tenantId": self.tenant_id,
-                                        "groupId": self.uuid,
-                                        "policyId": policy_id},
+                                         "groupId": self.uuid,
+                                         "policyId": policy_id},
                                         DEFAULT_CONSISTENCY)
             return d.addCallback(_extract_policy)
 
@@ -1173,9 +1172,9 @@ class CassScalingGroup(object):
             query = _cql_view_webhook.format(cf=self.webhooks_table)
             d = self.connection.execute(query,
                                         {"tenantId": self.tenant_id,
-                                        "groupId": self.uuid,
-                                        "policyId": policy_id,
-                                        "webhookId": webhook_id},
+                                         "groupId": self.uuid,
+                                         "policyId": policy_id,
+                                         "webhookId": webhook_id},
                                         DEFAULT_CONSISTENCY)
             d.addCallback(_assemble_webhook)
             return d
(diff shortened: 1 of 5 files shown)


radix:

otter/models/cass.py
```diff
@@ -206,7 +206,7 @@ def serialize_json_data(data, ver):
 
 _cql_count_for_tenant = (
     'SELECT COUNT(*) FROM {cf} '
-    'WHERE "tenantId" = :tenantId;')
+    'WHERE "tenantId"=:tenantId {deleting};')
```
radix: First I wanted to suggest changing {deleting} to something like {extra_clauses}, since it doesn't intrinsically have anything to do with deleting (other than where it's used)...

but then I figured it'd be better to just have two different constant CQL expressions, one for getting deleting groups and one for getting active groups, to keep CQL clause details out of the python code below...

but now I'm wondering why we ever want to get deleting groups in this query? It's not clear to me why `get_counts` should return information about deleting groups.

otter/models/cass.py
```diff
@@ -1585,20 +1590,30 @@ def add_webhook_keys(self, webhook_keys):
             CQLQueryExecute(query=batch(stmts), params=data,
                             consistency_level=ConsistencyLevel.ONE))
 
+    def _extract_count(self, r):
+        return r[0]['count']
+
+    def get_groups_count(self, log, tenant_id):
+        d = self.connection.execute(
```
radix: needs a docstring

otter/models/cass.py
```diff
@@ -1163,15 +1168,18 @@ def _assemble_webhook(cass_data):
...
+            d = self.connection.execute(query,
+                                        {"tenantId": self.tenant_id,
+                                         "groupId": self.uuid,
+                                         "policyId": policy_id,
+                                         "webhookId": webhook_id},
+                                        DEFAULT_CONSISTENCY)
+            d.addCallback(_assemble_webhook)
+            return d
+
+        return self.get_policy(policy_id).addCallback(fetch_webhook)
```
radix: Could this just call `view_config()` instead of `get_policy()`? That would reduce the extra roundtrips from 2 to 1.

otter/test/models/test_cass_models.py
```diff
@@ -2251,22 +2236,16 @@ def test_with_deleting(self):
...
-            'AND "groupId" = :groupId')
-        del_cql = ('DELETE FROM scaling_group '
-                   'WHERE "tenantId" = :tenantId AND "groupId" = :groupId')
-        exp_data = {'tenantId': self.tenant_id, 'groupId': self.group_id}
-        self.verified_view.assert_called_once_with(
-            self.connection, view_cql, del_cql,
-            exp_data, ConsistencyLevel.QUORUM,
-            matches(IsInstance(NoSuchScalingGroupError)),
-            self.mock_log, get_deleting=True)
+    def test_nogrouperror_on_deleting_group(self):
```
radix: Thanks for factoring out tests :)

otter/models/cass.py
manishtomar: This query is also used to get count of policies and webhooks from `scaling_policies` and `policy_webhooks` respectively which is why there is place for extra clause. If you prefer I could have two separate queries: One with `deleting=false` for groups and another without it for policies and webhooks.

otter/models/cass.py
manishtomar: Good idea. I thought doing that can allow this API to work on non-existent policy but that should not happen since corresponding webhook entry itself will not be there. Will change it

otter/models/cass.py
radix: I'm asking why these queries do want to know the deleting groups.

manishtomar: They do not want to know deleting groups. They want count of non-deleting groups. The code is also filtering with `deleting=false`.

radix: Now I'm more confused. I originally asked about `get_counts` (quoted below), which is where you're _not_ passing the "deleting=False" part of the query. Why does that need to know about the deleting groups?

> It's not clear to me why get_counts should return information about deleting groups.

radix: sorry, I kept misunderstanding because I didn't realize different tables were being passed to this query. It's all good :)

radix:
I'm fine with merging this, though if you want to wait until the migration is run that's fine :)

PR #1365, opened 2015-04-23
option to insert false in all group's deleting

manishtomar:
First task of #1363. Also refactored the script to easily add and run new migration functions. Tested this on devvm.

### DIFF
--- a/scripts/load_cql.py
+++ b/scripts/load_cql.py
@@ -8,21 +8,23 @@ from __future__ import print_function
 import argparse
 import re
 import sys
-from pprint import pprint
 
 from cql.apivalues import ProgrammingError
 from cql.connection import connect
 
 from effect.twisted import perform
 
-from silverberg.client import CQLClient
+from silverberg.client import CQLClient, ConsistencyLevel
 
 from twisted.internet import task
+from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.endpoints import clientFromString
 
 from otter.effect_dispatcher import get_cql_dispatcher
+from otter.metrics import get_scaling_groups
 from otter.models.cass import CassScalingGroupCollection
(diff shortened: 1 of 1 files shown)


scripts/load_cql.py
```diff
@@ -154,42 +153,61 @@ def execute_commands(cursor, commands, verbose):
...
+
+
+@inlineCallbacks
+def insert_deleting_false(reactor, conn):
+    """
+    Insert false to all group's deleting column
+    """
+    groups = yield get_scaling_groups(conn)
+    query = (
+        'INSERT INTO scaling_group ("tenantId", "groupId", deleting) '
```
radix: Is there a reason you're including `tenantId` and `groupId` here? Couldn't we just insert the `deleting` value?

manishtomar: Oh yes. Sorry don't know why I didn't do that.

manishtomar: Actually, I just realized `tenantId` and `groupId` is required to identify the row and column to insert to. There is no `WHERE` clause in insert statement.

radix: but I thought we're inserting it into all rows?

manishtomar: Yes. But still need to insert in each row separately. There is no way to insert to all rows. `INSERT` needs the primary key part of the table. This is why the code gets all the groups.

radix: Sorry, yet another difference with SQL that I wasn't aware of :)

radix: oh, and I was mentally reading "UPDATE" instead of "INSERT"...

manishtomar: Same deal with `UPDATE`. Its equivalent to `INSERT` since its basically a DHT.

radix:
LGTM :)

PR #1366, opened 2015-04-23
Gh1324 scale group over max after metadata removal

sam-falvo:
While I'm waiting for changes to be committed to mimic, I might as well post this PR to get early review as much as I can.


stretchservers:

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +69,66 @@ def _check_fds(_):
...
+        some random sampling of servers.
+        """
+
+        rcs = TestResources()
+
+        scaling_group_body = create_scaling_group_dict(
+            image_ref=image_ref, flavor_ref=flavor_ref,
+        )
+
+        self.scaling_group = ScalingGroup(
```
manishtomar: Can this go in `setUp`? I assume all tests will start with a group.

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +69,66 @@ def _check_fds(_):
...
+            scale_by=5,
+            scaling_group=self.scaling_group
+        )
+
+        return (
+            self.identity.authenticate_user(
+                rcs,
+                resources={
+                    "otter": ("autoscale", "http://localhost:9000/v1.0/{0}"),
+                    "nova": ("cloudServersOpenStack",),
```
manishtomar: These values should not be hard-coded in the test. It will become difficult to run these against prod/staging environments. The resources should be loaded based on environment variables / cmd-line args when the test suite starts.

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +69,66 @@ def _check_fds(_):
...
+                },
+                region=region,
+            ).addCallback(self.scaling_group.start, self)
```
manishtomar: Since all tests start of with a new group, can the creation of the group be in `setUp`? Any specific changes to the group can be made in the test. This way `rcs` can be stored in `self` and easily accessed throughout.

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +69,66 @@ def _check_fds(_):
...
+            .addCallback(
+                self.scaling_group.wait_for_deleted_id_removal,
+                rcs,
+                total_servers=25,
+            ).addCallback(
+                self.scaling_group.wait_for_N_servers, 25, timeout=1800
+            )
+        )
+
+    def test_aaa(self):  # XXX: Testing purposes only
```
cyli: Is this necessary for later?

sam-falvo: That will be removed once I am confident I no longer need to write:

`trial test_convergence.TestConvergence.test_rediculously_long_test_name_goes_here`

all the time.  :)  It's much easier to just write `trial test_convergence.TestConvergence.test_aaa`.  Also, easier to remember.

otter/integration/tests/test_convergence.py
sam-falvo: Is this a blocking issue?  If not, can we file a tech-debt ticket for this refactoring?

I haven't done this primarily because the scaling group seems configured differently for each test.  Changing the group takes as much LOC overhead as initially creating it.  I don't think it buys us anything.

I can see putting `rcs` into setup, but ... again, not sure that buys anything.

otter/integration/tests/test_convergence.py
cyli: Ah ok

otter/integration/tests/test_convergence.py
```diff
@@ -67,6 +70,72 @@ def _check_fds(_):
...
+            .addCallback(self.scale_beyond_max.start, self)
+            .addCallback(self.scale_up_to_max.execute)
+            .addCallback(
+                self.scaling_group.wait_for_N_servers, 25, timeout=1800
+            ).addCallback(self.scaling_group.get_scaling_group_state)
+            .addCallback(self._choose_random_servers, 3)
+            .addCallback(self._remove_metadata, rcs)
+            .addCallback(lambda _: rcs)
+            .addCallback(self.scale_beyond_max.execute)
+            .addErrback(expect_403)
```
cyli: There should probably be a check in here that we expect a 403, and not a success.  If `self.scale_beyond_max.execute` succeeds, the 403 check is skipped and the test might pass, and we probably want the test to fail, because the scaling policy should not execute?

otter/integration/tests/test_convergence.py
kivattik: I've added a function to do error code checking in #1369 which is up for review. Once that is merged, I think the _assert_error_status_code function is more general, though if it is going to be reused it should likely be moved with the other helper functions.

otter/integration/tests/test_convergence.py
kivattik: I will second Sam's request since we do often create scaling groups with different configurations. Having to ignore the default group and create a second was one of the frustrations with the previous framework. 

What cyli mentioned and that I agree with is that we could use a create_and_set_group function that would create a group with a Max/Min, set the desired capacity to a number of servers (usually different from Min) and then return once the correct number has been reached.

otter/integration/tests/test_convergence.py
sam-falvo: Is this comment intended for another PR?

sam-falvo: Fixed in 7367b27

otter/integration/tests/test_convergence.py
```diff
@@ -37,6 +41,9 @@
 # Get vs dict lookup because it will return None if not found,
 # not throw an exception.  None is a valid value for convergence_tenant.
 convergence_tenant = os.environ.get('AS_CONVERGENCE_TENANT')
+otter_key = os.environ['AS_AUTOSCALE_SC_KEY']
```
cyli: Might also be good to document these in the `config.sh-template`?

sam-falvo: fixed.

cyli:
LGTM, thanks @sam-falvo!

PR #1369, opened 2015-04-24
Add two test cases for catc7

kivattik:


otter/integration/tests/test_convergence.py
```diff
@@ -401,9 +403,8 @@ def test_scale_down_after_oobd_non_constrained_z_equal_y(self):
     def _scale_down_after_oobd_non_constrained_param(
             self, rcs, min_servers=0, max_servers=25, set_to_servers=0,
             oobd_servers=0, scale_servers=1):
-
         # This only applies if not constrained by max/min
-        converged_servers = set_to_servers + scale_servers
+        converged_servers = set_to_servers
```
cyli: I think this change breaks the previous 3 tests (`test_scale_down_after_oobd_non_constrained_z_lessthan_y`, `test_scale_down_after_oobd_non_constrained_z_greaterthan_y`, `test_scale_down_after_oobd_non_constrained_z_equal_y`), because the scaling policy that triggers convergence *does* get executed.

kivattik: Oops, yeah, that should only have been removed from the constrained version.

otter/integration/tests/test_convergence.py
```diff
@@ -428,6 +429,130 @@ def _scale_down_after_oobd_non_constrained_param(
...
+        (d.addCallback(
+            self.scaling_group.wait_for_N_servers,
+            min_servers if set_to_servers is None else set_to_servers,
+            timeout=120)
+         .addCallback(self.scaling_group.get_scaling_group_state)
+         .addCallback(self._choose_random_servers, oobd_servers)
+         .addCallback(self._delete_those_servers, rcs)
+         # The execution of the policy triggers convergence
+         .addCallback(self.policy_scale.start, self)
+          .addCallback(self.policy_scale.execute)
```
cyli: Non-blocking OCD nitpick - this line doesn't line up.  Actually I'm surprised the linter isn't complaining, but if the linter doesn't complain, I'm not going to block it.  :)

cyli:
Tests pass - this LGTM aside from the above (which again is non-blocking) :)

PR #1370, opened 2015-04-24
Specification based logging

manishtomar:
Implements part of #1107. Provides mapping from msg types to strings but does not validate. Validation can be done using JSON schema in other PRs.


otter/log/spec.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+# mapping from msg type -> message
+msg_types = {
+    "execute-convergence": "Executing convergence",
+    "execute-convergence-results": (
+        "Got result of {worst_status} after executing convergence"),
+    "launch-servers": "Launching {num_servers} servers",
+    "delete-server": "Deleting {server_id} server",
+    "add-server-clb": ("Adding {server_id} with IP address {ip_address} "
```
cyli: Non-blocking - we're using the bulk add/bulk delete apis for CLB, so wondering if these are going to end up being `server_ids` and `ip_addresses`.  But that can be fixed later when we do the actual logging.

otter/log/spec.py
```diff
@@ -0,0 +1,91 @@
...
+def try_msg_type(msg_type):
+    """
+    Try message type and return None if not found
+    """
+    return msg_type is not None and msg_types.get(msg_type, None)
+
+
+def get_validated_event(event):
+    """
+    Validate event's message as per msg_types and error details as
```
cyli: Non-blocking - might be good to include https://github.com/rackerlabs/otter/issues/1107#issuecomment-74938524 in the docstring so we know what kind of messages we should be using and getting.

otter/log/spec.py
```diff
@@ -0,0 +1,91 @@
...
+        observer(speced_event)
+
+    return validating_observer
+
+
+def try_msg_type(msg_type):
+    """
+    Try message type and return None if not found
+    """
+    return msg_type is not None and msg_types.get(msg_type, None)
```
cyli: This seems like it'd return `False` if `msg_type` is `None`?  The logic in `get_validated_event` seems to be expecting `None` or a string, though.

manishtomar: Good catch!

radix:
Very happy about this branch! I've been writing code for a while assuming this API, it's nice to see that it'll finally work as expected :)

otter/test/log/test_spec.py
```diff
@@ -0,0 +1,107 @@
...
+        otter_msg_type is added
+        """
+        e = {'isError': True, 'why': 'delete-server', 'a': 'b'}
+        self.assertEqual(
+            get_validated_event(e),
+            {'why': 'Deleting {server_id} server',
+             'isError': True, 'a': 'b',
+             'otter_msg_type': 'delete-server'})
+
+    def test_error_why_not_found(self):
```
cyli: This the same test as `test_error_not_found` - apologies, my previous comment was pretty confusing.  I meant a test for the case where 'why' or 'message' is not in the original event dict (and hence msg_type is None) :D

otter/log/spec.py
```diff
@@ -0,0 +1,84 @@
...
+def SpecificationObserverWrapper(observer):
+    """
+    Return observer that validates messages based on specification
+    and delegates to given observer
+    """
+    def validating_observer(event_dict):
+        try:
+            speced_event = get_validated_event(event_dict)
+        except (ValueError, TypeError):
+            speced_event = error_event(
```
radix: It seems like this might lose more information than we ought to lose, by replacing the original `failure`, `why`, and `message` keys. Perhaps `error_event` could just produce an entirely new dict, with an `original_event` key which contains the entire original `event_dict`.

otter/test/log/test_spec.py
manishtomar: Sheeesh. Yeah, it is the same test :). Removing it.

otter/log/spec.py
manishtomar: Yeah. Good catch.

cyli:
This LGTM @manishtomar once all of travis passes!  Thanks for all the fixes!

PR #1372, opened 2015-04-24
get_scaling_groups option to get null desired groups

manishtomar:
This is needed to do first task of #1363 since currently the migration script using `get_scaling_group` does not insert `false` on groups that has no `desired`.


cyli:
This LGTM - thanks @manishtomar :D

PR #1373, opened 2015-04-24
Convergence delete server fixes

cyli:
Ran into an issue when I was trying to delete servers in convergence - I forgot to wrap stuff up in TenantScope and forgot that ServiceRequest results were (response, body).

Really not sure how to test TenantScope, because I can't use it in the SequenceDispatcher, because the service request calls have callbacks, which makes the effects difficult to test for equivalence.  :|


radix:
LGTM :)

PR #1374, opened 2015-04-24
Use `with SequenceDispatcher.consumed` instead of asserting .sequence == []

radix:
I added a context manager to SequenceDispatcher so we don't need to check the guts of the SequenceDispatcher. It was released in effect==0.1a17 and I've updated our requirements.txt.


cyli:
LGTM, once merge conflicts are resolved!  Yay, this is neater, thank you.  :)

PR #1375, opened 2015-04-24
Better way of unpacking TenantScope and Retry - thanks @radix!

cyli:
A nicer way of performing `<TenantScope(Retry(intent)))>` type intents, thanks to @radix


radix:
@cyli - Awesome!

I think we'll want to generalize this soon -- there's some stuff in test_service.py that will probably want to do the same.  I might do that :) Thanks for the change!

PR #1376, opened 2015-04-25
Delete server with convergence

### LINKED ISSUE #1238: Can't use otter `delete server` REST API on servers that were created with convergence

cyli:
This PR is the final one that fixes #1238.  It:

1.  Enables all the server deletion/disowning cafe tests (with minor modifications to expect the state to take a little bit).
1.  Makes a controller `remove_server_from_group` which dispatches either to `convergence_remove_server_from_group` or `supervisor.remove_server_from_group` based on the tenant.
1.  Modifies the REST api to call `controller.remove_server_from_group` instead of `supervisor.remove_server_from_group`
1.  Moves `convergence_remove_server_from_group`'s params to match the `supervisor.remove_server_from_group`
1.  Fixes a bug with `convergence_remove_server_from_group` where it assumed `GetScalingGroupInfo` returned just the manifest, instead of (group, manifest)
1.  Makes the service dispatcher non-private, since the controller needs to use it.

~~These tests all pass on my dev vm, but one of them will fail on CI until https://github.com/rackerlabs/mimic/pull/221 is merged and we bump our pinned version.~~

Update: mimic PR merged and CI system upgraded

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -481,7 +481,7 @@ def remove_server_from_group(log, trans_id, server_id, replace, purge,
     Remove a specific server from the group, optionally replacing it
     with a new one, and optionally deleting the old one from Nova.
 
-    If the old server is not deleted from Nova, otter-specific metdata
+    If the old server is not deleted from Nova, otter-specific metadata
     is removed: otherwise, a different part of otter may later mistake
     the server as one that *should* still be in the group.
 
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -23,12 +23,9 @@ Storage model for state information:
 import json
 from datetime import datetime
 from decimal import Decimal, ROUND_UP
-from functools import partial
 
 from effect import (
     Effect,
-    ComposedDispatcher,
-    TypeDispatcher,
     parallel_all_errors)
 
 from effect.do import do, do_return
@@ -55,7 +52,6 @@ from otter.models.intents import GetScalingGroupInfo
 from otter.supervisor import (
     CannotDeleteServerBelowMinError,
     ServerNotFoundError,
-    get_supervisor,
     exec_scale_down,
     execute_launch_config)
 from otter.supervisor import (
@@ -68,7 +64,7 @@ from otter.util.retry import (
     retry_effect,
     retry_times)
 from otter.util.timestamp import from_timestamp
-from otter.worker_intents import EvictServerFromScalingGroup, perform_evict_server
+from otter.worker_intents import EvictServerFromScalingGroup
 
 
 class CannotExecutePolicyError(Exception):
@@ -501,7 +497,8 @@ def remove_server_from_group(log, trans_id, server_id, replace, purge,
         return worker_remove_server_from_group(
             log, trans_id, server_id, replace, purge, group, state)
 
-    # convergence case
+    # convergence case - requires that the convergence dispatcher handles
+    # EvictServerFromScalingGroup
     cs = get_convergence_starter()
     eff = convergence_remove_server_from_group(
         log, trans_id, server_id, replace, purge, group, state)
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -23,8 +23,14 @@ Storage model for state information:
 import json
 from datetime import datetime
 from decimal import Decimal, ROUND_UP
+from functools import partial
 
-from effect import Effect, parallel_all_errors
+from effect import (
+    Effect,
+    ComposedDispatcher,
+    TypeDispatcher,
+    parallel_all_errors)
+
 from effect.do import do, do_return
 from effect.twisted import perform
 
@@ -50,6 +56,7 @@ from otter.supervisor import (
     CannotDeleteServerBelowMinError,
     EvictServerFromScalingGroup,
     ServerNotFoundError,
+    get_supervisor,
     exec_scale_down,
     execute_launch_config,
     perform_evict_server)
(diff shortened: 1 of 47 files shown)


cyli:

otter/controller.py
```diff
@@ -467,3 +473,44 @@ def convergence_remove_server_from_group(
...
         yield do_return(state)
+
+
+def remove_server_from_group(log, trans_id, server_id, replace, purge,
+                             group, state, config_value=config_value):
+    """
+    Remove a specific server from the group, optionally replacing it
+    with a new one, and optionally deleting the old one from Nova.
+
+    If the old server is not deleted from Nova, otter-specific metdata
```
radix: typo: `metdata`

otter/controller.py
```diff
@@ -467,3 +473,44 @@ def convergence_remove_server_from_group(
...
+        return worker_remove_server_from_group(
+            log, trans_id, server_id, replace, purge, group, state)
+
+    def kick_off_convergence(new_state):
+        cs.start_convergence(log, group.tenant_id, group.uuid)
+        return new_state
+
+    # convergence case - requires that the convergence dispatcher handles
+    # EvictServerFromScalingGroup
+    cs = get_convergence_starter()
```
radix: There's quite a lot going on in this function. It looks like the test you wrote is pretty good, but I wonder if splitting it up would make it slightly easier. I generally prefer perform() to be in tail-call position of a function, that way no one function is dealing with both Effects and Deferreds, which can be a bit confusing. (I broke this rule in `ConvergenceStarter`, but other things that call perform() in `service.py` do follow the rule. In `ConvergenceStarter`, the only thing done with the result is adding a logging callback).

However, I will leave it up to your discretion, because I'm not strongly bothered by it. This PR LGTM if you want to merge it now. :+1:

cyli: Hmm... `start_convergence` returns a deferred, which is why I thought it wouldn't mix very well in the other effect code, but I guess I am not actually returning or doing anything in the deferred.  Probably seems like `convergence_remove_server_from_group` should not know about starting convergence - can I add `start_convergence` as a callback to the effect returned by `convergence_remove_from_group`?

cyli:

otter/test/test_controller.py
```diff
@@ -1464,3 +1471,96 @@ def test_checks_pass_replace_true_no_purge_failure(self):
...
+            test_dispatcher(),
+            SequenceDispatcher([
+                self._tenant_retry(
+                    get_server_details('server_id').intent,
+                    lambda _: (StubResponse(200, {}), self.server_details)),
+                (GetScalingGroupInfo(tenant_id='tenant_id',
+                                     group_id='group_id'),
+                    lambda _: (self.group, self.group_manifest_info)),
+                self._tenant_retry(
+                    mock.ANY,
```
radix: this mock.ANY is too lax I think, we should specifically be asserting that it's an `EvictServerFromScalingGroup` with all the right parameters.

otter/test/test_controller.py
```diff
@@ -1464,3 +1471,96 @@ def test_checks_pass_replace_true_no_purge_failure(self):
...
+                    lambda i: sync_perform(dispatcher, Effect(i)))
+            ]),
+            get_eviction_dispatcher(fake_super)
```
radix: and I think that this dispatcher should be left out, and the result simulated in the performer on line 1494, instead of calling the real performer in the unit test.  This also allows us to get rid of the scrub_calls mocking in this test.

To explain the philosophy behind this: if the `EvictServerFromScalingGroup` intent is constructed correctly, and the performer for `EvictServerFromScalingGroup` has been separately tested, we should be confident that it will be performed correctly. We shouldn't need to cross this boundary in our test. That is pretty much the primary reason of having abstract, application-level intents like this. Otherwise we could just use `Func` everywhere and have our tests be just as side-effecty as they always were.

otter/test/test_controller.py
```diff
@@ -1464,3 +1471,96 @@ def test_checks_pass_replace_true_no_purge_failure(self):
...
+        """
+        Perform :func:`convergence_remove_server_from_group` with the given
+        dispatcher.
+        """
+        self.state.desired = 2
+        fake_super = FakeSupervisor()
+
+        dispatcher = ComposedDispatcher([
+            test_dispatcher(),
+            SequenceDispatcher([
```
radix: We also probably ought to save this sequence dispatcher in a local variable and use the `with ...consume():` pattern.

otter/test/test_controller.py
cyli: Hmm... fair enough.  I intended the original test as more pinning several units together, but now that I've done the refactor this particular is merely testing that the effect is performed with the given dispatcher...

radix:
I think coveralls is dead. I'll go ahead and merge this, LGTM :)

PR #1377, opened 2015-04-26
otter start buckets error fix

manishtomar:
When starting otter, `divergent_changed` was called before the buckets were allocated causing error when it called `get_current_buckets`. Now, the function does not do anything if the buckets are not acquired.


cyli:
Awesome, this fixes the partitioner traceback :D  Thanks for fixing!  LGTM.

PR #1379, opened 2015-04-27
Added trigger convergence and get group config function

kivattik:


otter/integration/lib/autoscale.py
```diff
@@ -20,6 +21,13 @@
     transient_errors_except,
 )
 
+pp = pprint.PrettyPrinter(indent=4)
+# Get vs dict lookup because it will return 0 if not found, not throw an
```
cyli: Non-blocking - this comment seems superfluous, since the next line of code pretty much says this exactly.

otter/integration/lib/autoscale.py
```diff
@@ -132,6 +140,79 @@ def set_launch_config(self, rcs, launch_config):
...
+                            replacement_config=None):
+        """
+        Update the group configuration of a scaling group. Ifa replacement
+        configuration is provide, it will be used to replace all values.
+        Otherwise, only the specified parameters will be altered.
+        """
+        # Get the old config
+        old_config = self.get_group_config(rcs)
+        new_config = {}
+        if replacement_config is not None:
```
cyli: Non-blocking - probably better to eliminate the replacement_config, since this can be easily synthesized from all the others, and it's less confusing if function parameters didn't have an order of precedence/didn't supercede each other.

otter/integration/lib/autoscale.py
```diff
@@ -132,6 +140,79 @@ def set_launch_config(self, rcs, launch_config):
...
+            for g in rcs.groups:
+                if g['group']['id'] == self.group_id:
+                    g['group']['groupConfiguration'] = new_config
+            if verbosity > 0:
+                print('Updated group_config with {}'.format(new_config))
+            return rcs
+
+        return (
+            treq.put(
+                "%s/groups/%s/config" % (
```
sam-falvo: Non-blocking issue: to be consistent with the rest of the code in the file, you should use `"".format()`` instead of the `%` operator.  No need to change it now, but if you modify this file again later on, you should change it to match.  Thanks!

otter/integration/lib/autoscale.py
```diff
@@ -259,6 +340,9 @@ def start(self, rcs, test):
         def record_results(resp):
             rcs.groups.append(resp)
             self.group_id = str(resp["group"]["id"])
+            if verbosity > 0:
```
sam-falvo: Also non-blocking issue: Trial provides support for logging.  We should probably look into using that going forward.

otter/integration/tests/test_convergence.py
```diff
@@ -83,7 +83,6 @@ def test_scaling_to_clb_max_after_oob_delete_type1(self):
         for a CLB (as of this writing).
         """
         return self._perform_oobd_clb_test(25)
-    test_scaling_to_clb_max_after_oob_delete_type1.timeout = 1800
```
sam-falvo: Just curious -- why remove the timeouts here?

sam-falvo:
+1 LGTM overall.

otter/integration/lib/autoscale.py
```diff
@@ -132,6 +140,79 @@ def set_launch_config(self, rcs, launch_config):
             .addCallback(lambda _: rcs)
         )
 
+    def get_group_config(self, rcs):
```
cyli: +1 - glad this got separated out into its own utility function :D

cyli:
Awesome thanks for factoring this out!  Will be nice to be able to cal this :). LGTM

otter/integration/tests/test_convergence.py
kivattik: Since they were all the same, there was an added class timeout so that the value is only set in one place. (That was done in a different PR and these timeouts were likely merged in parallel.)

otter/integration/lib/autoscale.py
kivattik: Very much agree.

otter/integration/lib/autoscale.py
kivattik: I like the two function option and will make that change with other minor tweaks from comments.

otter/integration/lib/autoscale.py
```diff
@@ -132,6 +138,91 @@ def set_launch_config(self, rcs, launch_config):
...
+                            minEntities=None, maxEntities=None, metadata=None):
+        """
+        Update the group configuration of a scaling group. The provided
+        values will be updated, any others will remain unchanged.
+        """
+        # Get the old config
+        old_config = self.get_group_config(rcs)
+        new_config = {}
+
+        new_config["name"] = name or old_config["name"]
```
cyli: Apologies I didn't catch this before - might want to just loop over all the values here and check for None.  Since if you want to overwrite the name with "", this would not overwrite since "" evaluates to `False`.  Likewise if you wanted to set `minEntities` or `maxEntities` to 0, or if you wanted to set metadata to `{}`.

otter/integration/lib/autoscale.py
cyli: Uh, not sure what happened but github put this in the wrong line

otter/integration/lib/autoscale.py
kivattik: Good point, and I think I need to change it now.

cyli: Bleh, ok, my attempts to move things in github had failed.  this was a reply a comment I made about testing for `None` evaluating to false.  Anyhoo, sorry about the confusion.  :)

cyli:
LGTM!  I think coveralls is hosed. Just going to merge :)

PR #1380, opened 2015-04-27
Utility for getting the servicenet IPs of active servers

cyli:
To be used for the trial integration tests - to test membership of the servers on CLBs


otter/integration/lib/autoscale.py
```diff
@@ -100,11 +101,26 @@ def create_scaling_group_dict(
...
+        created with - a dictionary including 'groupConfiguration',
+        'launhConfiguration', and maybe 'scalingPolicies'
+
+    :ivar pool: a :class:`twisted.web.client.HTTPConnectionPool` to pass to
+        all treq requests
+
+    :ivar reactor: a :class:`twisted.internet.interfaces.IReactorTime`
+        provider, to be used for timeouts and retries
+
+    :ivar _treq: the treq module to use for making requests - if not provided,
```
manishtomar: typo: `treq`?

otter/integration/lib/test_autoscale.py
```diff
@@ -98,3 +100,83 @@ def test_poll_until_timeout(self):
...
+
+        class FakeRCS(object):
+            endpoints = {'nova': 'novaurl'}
+            token = "token"
+
+        @attributes(['code', 'fake_json_response'])
+        class Resp(object):
+            pass
+
+        class FakeTreq(object):
```
manishtomar: You could use `otter.test.utils.StubTreq2` instead.

cyli: I was going to use that and StubResponse, but I think @sam-falvo and @kivattik needed to use as little of the otter-specific test utilities as possible, because they needed to split the library out at some point.  Hence why find_endpoint and the auth stuff is all wrapped.

otter/integration/lib/autoscale.py
cyli: Ah, yes, thanks!  :D

sam-falvo: I had no idea about ivar and other attributes.  Where can I find a definitive source of documentation on class and method pydocs?

otter/integration/lib/autoscale.py
```diff
@@ -181,20 +197,23 @@ def get_scaling_group_state(self, rcs):
             order.  If not found, the result code will be 404, and the state
             will be None.
         """
+        if success_codes is None:
```
sam-falvo: Non-blocking :: this can be expressed more simply as `success_codes = success_codes or [200, 404]`

otter/integration/lib/autoscale.py
```diff
@@ -262,17 +280,58 @@ def record_results(resp):
...
+            if private is not None:
+                return [addr['addr'] for addr in private
+                        if addr['version'] == 4][0]
+            return None
+
+        def _get_the_ips(the_server_ids):
+            the_server_ids = set(the_server_ids)
+            return gatherResults([
+                self.treq.get(
+                    "{0}/servers/{1}".format(rcs.endpoints["nova"], server_id),
```
sam-falvo: Because this method now introduces a hard dependency on having Nova's service catalog endpoint defined, I think it'd be wise to annotate this in the pydoc for the `get_servicenet_ips` method.

At some point in the future, in the interest of keeping everything as modular as possible, I'd like to introduce a .../lib/nova.py component to the library that lets us interact with Nova discretely from any other Openstack API.  I just haven't found a satisfactory way to do that yet.

otter/integration/lib/test_autoscale.py
sam-falvo: @manishtomar It's true; I have pressure from Pranav (and not without merit) to keep as much of this library independent of Otter as possible, because we'd like to take this work and apply it eventually to other projects in the office.  Also, the work being invested in developing this Trial-based library will eventually make its way into OpenCafe.  So, it's very important I distinguish between what is absolutely required to be reused from Otter versus what I can reasonably implement myself externally.

That means that, eventually, I'll need to implement my own authentication mechanism in Identity class.  But, I'll cross that bridge when I get there.

And, since it's so early in the morning for me still, when I say "I", I of course mean the QE group as a whole.  :)

otter/integration/lib/test_autoscale.py
```diff
@@ -98,3 +100,83 @@ def test_poll_until_timeout(self):
...
+    def test_queries_for_provided_server_ids(self):
+        """
+        If server IDs are provided, IPs are queried for those server IDs.
+        And if the same server ID is given multiple times, only one query is
+        made for any given server ID.
+        """
+        server_ids = ['1', '2', '2', '2', '3']
+        d = self.sg.get_servicenet_ips(self.rcs, server_ids)
+        result = self.successResultOf(d)
+        self.assertEqual(['10.0.0.1', '10.0.0.2', '10.0.0.3'],
```
sam-falvo: I don't understand how we arrive at 10.0.0.[1-3] here.  According to line 137 above, we merely return the _length_ of the list of queried server IDs as the final IP octet.  Since the length remains a constant, wouldn't we receive 10.0.0.3 for all three (unique-ified) server IDs?

otter/integration/lib/test_autoscale.py
```diff
@@ -98,3 +100,83 @@ def test_poll_until_timeout(self):
...
+        If server IDs are not provided, IPs are queried for the active servers
+        on the group server IDs.
+        """
+        def get_scaling_group_state(_, success_codes):
+            self.assertEqual(success_codes, [200])
+            return defer.succeed(
+                {'group': {'active': [{'id': '11'}, {'id': '12'}]}}
+            )
+
+        self.sg.get_scaling_group_state = get_scaling_group_state
```
sam-falvo: informational :: The fact that I constantly forget that I can do this is the primary reason why I haven't been writing more tests for my tests.  (Yo dawg...)  Thanks for doing this.  I'll try to make a better effort to remember this going forward.

otter/integration/lib/test_autoscale.py
cyli: Each time that method is called, the length changes.  So the first time it's called, it appends to the list, and the length is 1.  The second time it's called, it appends to the list, and the length is 2.  Third time it's called, it appends to the list, and the length is 3.

At the end, definitely the length is 3, but since `len(self.queried_server_ids)` is called each time the function is called to generate a response, we get 3 different responses :)

otter/integration/lib/autoscale.py
cyli: Here is the sphinx documentation for the Python domain - we usually leave out the :py: stuff since we assume it's understood:  http://sphinx-doc.org/domains.html#the-python-domain.

https://docs.python.org/devguide/documenting.html is also a good guide (it's the documentation standard for CPython I think).

otter/integration/lib/autoscale.py
cyli: How would you feel about this function also checking if the endpoint is there, and if not, then finding it?

otter/integration/lib/autoscale.py
cyli: I generally like to check specifically for None as the sentinel value, in case someone wants to always fail the request by passing `success_codes=[]`, point well taken.  :)  I can definitely make this a one-liner, thank you!

otter/integration/lib/test_autoscale.py
sam-falvo: OH!  DUHH!  I completely missed line 132.  Sorry!

otter/integration/lib/autoscale.py
sam-falvo: Convenient, but not comfortable with it.  I fear too much behavior.  I'd like to avoid god-classes and implicit behavior if possible.

cyli: Ok, will document then.  :)

cyli:
@sam-falvo I think the PR comments have been addressed - would you mind having another look?

sam-falvo:
LGTM +1 :+1: and other related emoji.  ;)

PR #1386, opened 2015-04-29
Use nova's pagination API

cyli:
If we get a list of servers, and it has a `servers_link` object in the body, we use that link instead of constructing our own batch URL.

If that link is not there, either the service doesn't support pagination, or there are no more pages.

Also introduced a check to make sure Nova doesn't return the same link twice, because that would be broken (since Nova depends upon limits and markers to differentiate pages).

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -72,7 +72,7 @@ def get_all_server_details(changes_since=None, batch_size=100):
                     "When gathering server details, got the same 'next' link "
                     "twice from Nova: {0}".format(last_link[-1]))
 
-            last_link.append(continuation[0])
+            last_link[:] = [continuation[0]]
             more_eff = get_server_details(continuation[0])
             return more_eff.on(lambda more_servers: servers + more_servers)
         return servers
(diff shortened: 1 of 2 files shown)


cyli:

otter/convergence/gathering.py
```diff
@@ -41,30 +47,37 @@ def get_all_server_details(changes_since=None, batch_size=100):
...
+        continuation = [link['href'] for link in body.get('servers_links', [])
+                        if link['rel'] == 'next']
+        if continuation:
+            # blow up if we try to fetch the same link twice
+            if last_link and last_link[-1] == continuation[0]:
+                raise UnexpectedBehaviorError(
+                    "When gathering server details, got the same 'next' link "
+                    "twice from Nova: {0}".format(last_link[-1]))
+
+            last_link.append(continuation[0])
```
radix: This could just be `last_link[:] = [continuation[0]]` to avoid keeping old links around, since we don't use them.

radix:
LGTM once integration tests finish. and you can address the one nit I had if you want :)

cyli:
Meh, not going to wait on coveralls.  @radix: thanks for the review!

PR #1387, opened 2015-04-29
Switch some convergence='error' tests to convergence='never', with explanations

radix:
All of these tests fail because they rely on the old worker behavior of deleting servers that couldn't be added to a LB. Now convergence just fails.

The explanations have some, uh, forward-looking statements about the group moving to ERROR state.

Also improved some error reporting in an integration test assertion method.


radix:

cyli:
There seems to be a lint error.  Other than that, if travis and integration tests pass, LGTM :)

radix:

Issue #1225, opened 2015-03-23
Efficiently throttle server creation in convergence


radix:
We need to maintain parity with the existing worker code and make sure to not run multiple CreateServer requests at the same time.

However, the basic convergence throttling isn't really great at this, because if we throttle it to one CreateServer at a time, it'll cause N convergence iterations where N is the number of servers to create, and that's super inefficient.

So we should have an optimizer for CreateServer steps that aggregate them into a CreateServers step, which does its own throttling.

cyli:
We should also make the limit configurable, since this greatly speeds up the otter/integration (as opposed to the cloudcafe integration, although it might help those too) tests.

radix:
well apparently they only ask for 1s between creation requests so it may not matter, but yeah we could configure it to 0 for our mimic integration tests.

manishtomar:
I've never been comfortable with planner doing throttling. Throttling is pure execution strategy and should exist in executor which in our case is `execute_convergence` function. There can be `execute_steps` instead of `steps_to_effect` that takes the steps and executes them using some throttling policy. This method can return `worst_status` depending on how much it managed to execute.

radix:
@manishtomar the "planner doing throttling" is a vague concept, and I don't think a very accurate characterization of what I suggested. My proposal is to have a function of steps -> steps, which can be easily composed into the flow of steps to be executed, and doesn't need to affect either the way steps are originally produced (with `converge`) or performed (with `execute_convergence`). The core logic of `converge` remains as clean as it is today, and we only have to write another function and compose it into `plan`. So yes, I do propose `plan` invoke the work of throttling our steps, but `plan` is only a trivial composition of other things (plan, optimize, throttle), and it will remain so.

Maybe I'm missing something and there's some good reason to put this logic into `execute_convergence`, but right now it seems keeping it separated is beneficial.

Keeping all of this stuff in step-world maintains a nice "functional core, imperative shell" where our internal plan can transparently represent the logic of a convergence iteration without involving opaque things like Effects. This means that more of the logic is easily testable without invoking Effect-y stuff in our tests.

Also, if you're concerned about having throttling be specific to CreateServer, we can easily generalize this concept of throttling steps within a convergence iteration by having a generic step that can throttle any set of other steps, and I recommend we do that (I think we've already got a request from the nova to also throttle deletions, so it will be immediately useful).

cyli:
Isn't our existing throttling already generalized?  It throttles based on step type.

[...]

PR #1388, opened 2015-04-30
Move step-transforming functions to `transforming.py`

radix:
Move the optimizers and step-limiting code to otter.convergence.transforming. 

The TBD throttling code will go in this new module too.

### DIFF
--- a/otter/convergence/transforming.py
+++ b/otter/convergence/transforming.py
@@ -5,7 +5,7 @@ There are a few classes of transformations here:
 
 - optimizing (converting multiple steps into one for the purposes of reducing
   API roundtrips)
-- throttling (by truncating the number of steps we take)
+- limiting (by truncating the number of steps we take)
 - throttling (by wrapping multiple steps with one that gradually executes them)
 """
 
--- a/otter/convergence/transforming.py
+++ b/otter/convergence/transforming.py
@@ -1,12 +1,9 @@
 """
 Functions for transforming collections of steps.
 
-There are a few classes of transformations here:
-
 - optimizing (converting multiple steps into one for the purposes of reducing
   API roundtrips)
 - limiting (by truncating the number of steps we take)
-- throttling (by wrapping multiple steps with one that gradually executes them)
 """
 
 from functools import partial
--- a/otter/convergence/transforming.py
+++ b/otter/convergence/transforming.py
@@ -1,0 +1,117 @@
+"""
+Functions for transforming collections of steps.
+
+There are a few classes of transformations here:
+
+- optimizing (converting multiple steps into one for the purposes of reducing
+  API roundtrips)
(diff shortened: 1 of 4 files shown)


cyli:

cyli:
LGTM - just running jenkins tests out of paranoia.  Please merge once it passes!  :)

manishtomar:
I've some concerns about the structure of code. Will post them shortly. Although not sure if it might help since they are same raised in #1225.

radix:
looking forward to hearing it @manishtomar  :)

otter/convergence/transforming.py
```diff
@@ -0,0 +1,117 @@
+"""
+Functions for transforming collections of steps.
+
+There are a few classes of transformations here:
+
+- optimizing (converting multiple steps into one for the purposes of reducing
+  API roundtrips)
+- limiting (by truncating the number of steps we take)
+- throttling (by wrapping multiple steps with one that gradually executes them)
+"""
```
manishtomar: This is probably just bike-shedding :) but it feels like you are arranging the code based on its _type_ rather than its _role_. Here, this module does more than one thing (role) but is in same module because of its type (functional non-effect code). IMHO, the code should be organized based on its role. So, here the optimizing part can remain in `plan.py` and rest (limiting and throttling) in say `throttle.py`. I do like the idea of representing throttle as a separate step :). In any case, this is non-blocking.

cyli: @manishtomar To jump on the bikeshedding bandwagon, I agree with you that separating based on type may not be a good idea, but I kind of think of optimizing, limiting, and throttling as a role in that shaves things down based on steps produced by the planner.  The planner produces steps based on the current state of the world, and this module refines those steps to be more optimized in terms of networking.

But I can also see the perspective that optimizing is a separate role from limiting and throttling.

cyli: E.g. I'm not disagreeing with that perspective, just pointing out another way to look at it as well.  :)  I don't feel strongly whichever way we go.

radix: yes, there are many different ways to categorize "roles" -- the one I used here is the role of "transformation of steps after the general plan has been calculated".

I do not think any of the code here should stay in plan.py. It's better to err on the side of smaller modules than bigger ones. Would you like me to create "optimizing.py", "throttling.py", and "limiting.py"?

radix:
I updated the docstring to not mention the throttling thing, since we decided against doing that.

radix:
Thanks!

FTR: 
```
radix [12:36 PM] manisht: do you have any more thoughts on https://github.com/rackerlabs/otter/pull/1388 ?
manisht [12:37 PM] radix: sorry no :simple_smile: its all good
```

PR #1392, opened 2015-05-01
Logging effect

manishtomar:
Writing log as effect with context added. `convergence/service.py` has been changed to use new logging.


otter/log/intent.py
```diff
@@ -0,0 +1,76 @@
...
+    return Effect(Log(msg, fields))
+
+
+def err(failure, msg, **fields):
+    return Effect(LogErr(failure, msg, fields))
+
+
+def perform_logging(log, fields, disp, intent, box):
+    all_fields = merge(fields, intent.fields)
+    if type(intent) is Log:
```
radix: The fact that all of the lines in this function (other than the first) involve type-checks on the intent gives a pretty strong indication that this should be three different functions registered as performers for their respective intents. In other words, `TypeDispatcher` already does type dispatching for you.

radix:
This is pretty dang awesome! Thank you, @manishtomar :)

manishtomar:

otter/convergence/service.py
```diff
@@ -415,9 +419,9 @@ def __init__(self, log, dispatcher, buckets, partitioner_factory,
             create an :obj:`Partitioner` to distribute the buckets.
         """
         MultiService.__init__(self)
+        self.log = log.bind(system='converger')
```
radix: Could you add `otter_service` back? It was an intentional decision to use `otter_service` instead of `system` because `system` gets overridden so much that a lot of context gets lost in the logs.

otter/convergence/service.py
```diff
@@ -438,8 +442,11 @@ def buckets_acquired(self, my_buckets):
 
         This is used as the partitioner callback.
         """
-        eff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
+        ceff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
             partial(self._converge_all, my_buckets))
+        eff = Effect(Func(uuid.uuid1)).on(
+            lambda uid: with_log(ceff, system='converger',
+                                 converger_run_id=uid))
```
radix: Cool, thanks for adding this!

otter/convergence/service.py
```diff
@@ -438,8 +442,11 @@ def buckets_acquired(self, my_buckets):
 
         This is used as the partitioner callback.
         """
-        eff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
+        ceff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
             partial(self._converge_all, my_buckets))
+        eff = Effect(Func(uuid.uuid1)).on(
+            lambda uid: with_log(ceff, system='converger',
```
radix: Did you mean to add `system` here again? I'm not sure why it's necessary

radix: oh wait, no, I do understand, since we're not using self.log any more. then I guess my same comment applies: please use `otter_service` instead of (or in addition to) `system`

otter/test/convergence/test_service.py
```diff
@@ -104,24 +108,33 @@ def test_buckets_acquired(self):
...
         """
-        def converge_all_groups(log, currently_converging, _my_buckets,
+        def converge_all_groups(currently_converging, _my_buckets,
                                 all_buckets, divergent_flags):
             return Effect(
-                ('converge-all', log, currently_converging, _my_buckets,
+                ('converge-all', currently_converging, _my_buckets,
                  all_buckets, divergent_flags))
 
+        def perform_ca_eff(bound_fields):
```
radix: I would prefer to move @cyli's `_unwrap_wrapped_effect` from `test_controller.py` to `test.utils` and use it here. I think it allows the code to read a bit better, since you can nest the intents within the main sequence, showing the linear flow of effects mapped directly to linear code

otter/test/convergence/test_service.py
```diff
@@ -203,113 +228,72 @@ class ConvergeOneGroupTests(SynchronousTestCase):
...
-        self.assertEqual(
-            calls,
-            [(self.tenant_id, self.group_id,
-              matches(IsBoundWith(tenant_id=self.tenant_id,
-                                  scaling_group_id=self.group_id)))])
-        self.assertEqual(self.deletions, [True])
-        self.log.msg.assert_any_call(
-            'mark-clean-success',
-            tenant_id=self.tenant_id, scaling_group_id=self.group_id)
+        sequence = SequenceDispatcher([
```
radix: Thank you very much for converting these tests to use SequenceDispatcher :)

otter/test/convergence/test_service.py
```diff
@@ -406,38 +371,36 @@ def test_converge_all_groups(self):
         """
         eff = self._converge_all_groups(['00_g1', '01_g2'])
 
+        def perform_gs_effect(tid, gid, intent):
```
radix: same as the comment above with `perform_ca_effect`

otter/test/convergence/test_service.py
manishtomar: I thought so too but didn't understand that when I last glanced. I'll look at it and try to factor it out.

radix: it's a bit weird, but it leads to nice concise code :) it assumes that the intent passed has an `effect` attribute, which is true for all of our wrapper-intents. It works with any type of wrapper-intent and runs a performer on a SequenceDispatcher for the wrapped intent.

radix: Oh, it looks like the existing function won't work exactly. I suggest changing it slightly to accept a list instead of just one tuple, to allow multiple steps.

otter/test/convergence/test_service.py
```diff
@@ -104,24 +109,31 @@ def test_buckets_acquired(self):
...
             return Effect(
-                ('converge-all', log, currently_converging, _my_buckets,
+                ('converge-all', currently_converging, _my_buckets,
                  all_buckets, divergent_flags))
 
         my_buckets = [0, 5]
-        sequence = SequenceDispatcher([
-            (GetChildren(CONVERGENCE_DIRTY_DIR), lambda i: ['flag1', 'flag2']),
-
+        bound_sequence = [
```
radix: I think this looks better, but one of my original points was specifically to include these intents in-line in the main sequence list, so the order is nicely represented lexically in the code.  Just a stylistic thing, though :)

manishtomar: While I do agree that having single list is nice, unfortunately it leads to deeply intended code. Especially, if its more than one level nested effects. The above would look like
```python
        sequence = SequenceDispatcher([
            (Func(uuid.uuid1), lambda i: 'uid'),
            unwrap_wrapped_effect(
                BoundFields, dict(fields={'system': 'converger',
                                          'converger_run_id': 'uid'}),
                [(GetChildren(CONVERGENCE_DIRTY_DIR),
                    lambda i: ['flag1', 'flag2']),
                (('converge-all',
                    transform_eq(lambda cc: cc is converger.currently_converging,
                                True),
                    my_buckets,
                    self.buckets,
                    ['flag1', 'flag2']),
                    lambda i: 'foo')
               ])
        ])
```
when it is sequential. It will become difficult to manage the indentation with 80 char limit. 

Sometimes, having separate `perform_` function for internal effect makes the distinction more clear. I find it easy to understand the effect code when seeing separate perform functions in test. It has direct mapping to effect code. This seems to be the case with https://github.com/rackerlabs/otter/blob/logging-effect/otter/test/convergence/test_service.py#L366 test. Sometimes they make sense inline. I am actually fine either way as long as there is not too much indentation :)

radix: Okay, no big deal. I'm happy with this branch once the `otter_service` thing is fixed, feel free to merge after that :)

manishtomar: Thanks :) I hope I wasn't difficult in this discussion. I'll change everything to use `unwrap_wrapped_effect` anyway. That function is cool!

radix: Not at all! I think the use of unwrap_wrapped_effect was more important than the stylistic stuff.  Thank you very much :)

otter/convergence/service.py
manishtomar: Oh. I am sorry I just realized that it _was_ otter_service earlier and I changed it to system. I thought it is new change you are suggesting. `otter_service` coming up :)

radix:
LGTM :D

PR #1397, opened 2015-05-02
Refactor: Move things around to categorize catc tests

kivattik:


cyli:
+1 Awesome, thanks for splitting this up.  :)  There are some commented out skips, but as you're about to refactor all of that, we can merge as is and they can be deleted in the next PR.

PR #1402, opened 2015-05-04
Don't format kazoo errors in logging

cyli:
So we stop getting message formatting errors for kazoo.

### DIFF
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -27,6 +27,8 @@ THROTTLED_MESSAGES = [
 
 THROTTLE_COUNT = 50
 
+NON_PEP3101_SYSTEMS = ('kazoo',)
+
 
 class LoggingEncoder(json.JSONEncoder):
     """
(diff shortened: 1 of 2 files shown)


otter/log/formatters.py
```diff
@@ -127,16 +129,17 @@ def PEP3101FormattingWrapper(observer):
...
-        if eventDict.get('why'):
+        if (eventDict.get('why') and
+                eventDict.get('system') not in NON_PEP3101_SYSTEMS):
             eventDict['why'] = eventDict['why'].format(**eventDict)
 
         if 'message' in eventDict:
             message = ' '.join(eventDict['message'])
 
-            if message:
+            if message and eventDict.get('system') not in NON_PEP3101_SYSTEMS:
```
radix: Couldn't this new conditional clause be moved up to the `if 'message' in eventDict` conditional?

radix: Oh, never mind, I see that we still want the ' '.join.

radix:
LGTM!

Issue #1404, opened 2015-05-04
Make cron scheduler tests deterministic


manishtomar:
This can be done by using `wait_for_expected_group_state` with short interval instead of sleeping for `60 + scheduler_interval` seconds and then calling `verify_group_state`. This is already done in few tests like `test_system_cron_style_policy_executes_again`.

PR #1405, opened 2015-05-04
enabling convergence on scheduler tests

manishtomar:
All the scheduler tests pass with convergence. They will become more deterministic after doing #1404.


radix:
Thanks!

PR #1406, opened 2015-05-04
Make integration tests fail faster, other cleanup

cyli:
This PR does the following:

1. Make all errors (except the transient error) terminal while waiting for the state.  This way if there's an error in the code or some issue with mimic or otter, things fail immediately.
1. Fixes the bug where the pool was not being passed to the Nova API when deleting metadata, meaning that the tests hung forever.
1. @glyph suggests that waiting for the FDs is unnecessary, and closing the pool's cached connections will do the right thing.  So remove that.
1. Minor other cleanup with allowing the success code to be passed to execute scaling policy, so we can assert that an execution should have failed.
1. Makes all the waiting time out at 1800.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -666,12 +666,15 @@ class ScalingPolicy(object):
             .addCallback(check_success, [204, 404])
         ).addCallback(lambda _: rcs)
 
-    def execute(self, rcs):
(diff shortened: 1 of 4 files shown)


cyli:

cyli:

cyli:
jenkins please run integration tests.

radix:
Great changes! Thank you @cyli :)

PR #1409, opened 2015-05-05
CATC 4-12 with CLB

### LINKED ISSUE #1306: Test OOB deleted servers LBs
Test for #881. The test should directly delete server in the group from Nova that is attached to CLB and expect the CLB node to be cleaned up in next convergence cycle.

cyli:
This doesn't actually implement the missing 4-12 functions, but refactors the existing ones so that they can also be run with CLBs.

The refactor includes the following:

1.  Introduces a helper object that can be composed into the existing tests to perform useful functions.
1.  Move test resources and authentication to the setUp currently, so each test (with and without CLB) can decide which endpoints they need.  This can probably even just be moved to the helper object and just parse all the endpoints immediately.
1.  Introduces a decorator function that OOB-deletes some servers from a group, executes the decorated function, then asserts that the deleted servers have been removed from the group and cleaned up from any load balancers.
1.  Changes existing tests to make use of the decorator function.
1.  Adds another TestCase class that runs 4-12 with a CLB.  These are currently skipped because autoscale doesn't clean up deleted servers from load balancers yet (see #881).

This PR fixes #1306 

Note that all the tests in `ConvergenceSet1` have been changed, since that seems to be where the 4-12 test cases are going.  But the tests in `TestConvergence` have not been modified more than necessary yet, to try to keep this PR relatively small.

### DIFF
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -4,10 +4,13 @@ Tests covering self-healing should be placed in a separate test file.
 
 from __future__ import print_function
 
+from functools import wraps
 import os
 
 from twisted.internet import reactor
-from twisted.internet.defer import gatherResults
+from twisted.internet.defer import gatherResults, inlineCallbacks, returnValue
+from twisted.internet.task import deferLater
+from twisted.internet.tcp import Client
 from twisted.trial import unittest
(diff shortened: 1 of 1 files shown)


sam-falvo:
LGTM.

PR #1412, opened 2015-05-05
Cloud client throttling

radix:
This is a rather hacky hack to add throttling to server creation and deletion, with configuration.


radix:

radix:

cyli:

otter/cloud_client.py
```diff
@@ -192,11 +200,85 @@ def got_auth((token, catalog)):
...
+    bracket = throttler(service_request.service_type,
+                        service_request.method.lower())
+    if bracket is not None:
+        return Effect(_Throttle(bracket=bracket, effect=eff))
+    else:
+        return eff
+
+
+@attributes(['bracket', 'effect'])
+class _Throttle(object):
```
manishtomar: Is it beginning with underscore because it is dependent on Twisted?

otter/cloud_client.py
```diff
@@ -220,6 +302,21 @@ def scoped_performer(dispatcher, service_request):
     perform(new_disp, tenant_scope.effect.on(box.succeed, box.fail))
 
 
+def get_cloud_client_dispatcher(reactor, authenticator, log, service_configs):
+    """
+    Get a dispatcher suitable for running :obj:`ServiceRequest` and
+    :obj:`TenantScope` intents.
+    """
+    # this throttler could be parameterized but for now it's basically a hack
+    # that we want to keep private to this module
```
manishtomar: Do you want to create a tech-debt ticket that describes the ideal throttler? Or an issue in effect?

otter/cloud_client.py
```diff
@@ -192,11 +200,85 @@ def got_auth((token, catalog)):
...
+
+    A "Deferred bracket" is some function of type
+    ``((f, *args, **kwargs) -> Deferred) -> Deferred``
+    basically, something that "brackets" a call to some Deferred-returning
+    function. This is the case for the ``run`` method of objects like
+    :obj:`.DeferredSemaphore` and :obj:`.DeferredLock`.
+
+    https://wiki.haskell.org/Bracket_pattern
+
+    Ideally Effect would just have built-in concurrency limiters/idioms without
```
manishtomar: Why not do this now? Is it too complicated? It's really unfortunate to see deferred mixed with effect in an intent object. How about `effect -> effect` function that does bracketing/throttling? It's performer could use deferred semaphore. Sorry I am sure you might have thought of this. Maybe it is complicated. I'll try to think some more and let you know if something comes up.

otter/cloud_client.py
radix: No, because I don't want anyone using it :)

otter/cloud_client.py
radix: Mostly because I didn't want to put any more thought into it, since this should just be a temporary thing anyway.

The comment was just about how Effect should have things like EffectLock and EffectSemaphore.

manishtomar: Until those are implemented, what are your thoughts on throttling at `Request` effect's [implementation](https://github.com/rackerlabs/otter/blob/master/otter/util/pure_http.py#L44)? You can parse the url to know the service and throttle using deferred primitives. This is not great but is lesser evil.

radix: Request only knows URLs, not about ServiceType, so it'd be trickier to determine when to do the filtering there.

I think shuffling this stuff around into different forms is not going to help much because we should be able to either have a real, exposed solution with a configurable API, or just delete it instead.

otter/cloud_client.py
```diff
@@ -192,11 +200,85 @@ def got_auth((token, catalog)):
...
+    """
+    Perform :obj:`_Throttle` by performing the effect after acquiring a lock
+    and delaying but some period of time.
+    """
+    lock = throttle.bracket
+    eff = throttle.effect
+    return lock(twisted_perform, dispatcher, eff)
+
+
+def _serialize_and_delay(clock, delay):
```
manishtomar: +1. This is pretty cool. Can be used in worker too!

otter/test/test_cloud_client.py
```diff
@@ -188,7 +201,11 @@ def test_binds_url(self):
         Binds a URL from service config if it has URL instead of binding
         URL from service catalog
         """
-        self.service_configs[ServiceType.CLOUD_SERVERS]['url'] = 'myurl'
+        self.service_configs = update_in(
+            self.service_configs,
+            (ServiceType.CLOUD_SERVERS, 'url'),
+            lambda _: 'myurl'
+        )
```
manishtomar: Non-blocking: Any reasons for doing this?

otter/test/test_cloud_client.py
```diff
@@ -255,6 +272,199 @@ def test_params(self):
...
+        performer.
+        """
+        def bracket(f, *args, **kwargs):
+            return f(*args, **kwargs).addCallback(lambda r: ('bracketed', r))
+        throttle = _Throttle(bracket=bracket, effect=Effect(Constant('foo')))
+        dispatcher = ComposedDispatcher([
+            TypeDispatcher({_Throttle: _perform_throttle}),
+            base_dispatcher])
+        result = sync_perform(dispatcher, Effect(throttle))
+        self.assertEqual(result, ('bracketed', 'foo'))
```
manishtomar: +1. I like the way result reads: "bracketed foo"

otter/test/test_cloud_client.py
radix: Yeah... Since I made it so self.service_configs is a module-level constant (so I could share it between test suites), I had to avoid mutating it. First I tried deepcopying it, but that didn't work because it turns out if you deepcopy a Twisted constant you can no longer use it for identity comparison. :-(

I realized another thing I could do would be to just make a module-level function that returns the dict. I'll make that change.

otter/cloud_client.py
radix: Yeah, thanks. I've added radix/effect#48 -- and it reminded me of radix/effect#8, which it could be used to implement.

otter/test/test_cloud_client.py
radix: Okay, I've simplified this with a function :) thanks for the point

otter/test/test_cloud_client.py
```diff
@@ -255,6 +272,199 @@ def test_params(self):
...
+    def test_serialize_and_delay(self, deferred_lock):
+        """
+        :func:`_serialize_and_delay` returns a function that, when given a
+        function and arguments, calls it inside of a lock and after a specified
+        delay.
+        """
+        class DeferredLock(object):
+            def run(self, f, *args, **kwargs):
+                return f(*args, **kwargs).addCallback(lambda r: ('locked', r))
+        deferred_lock.side_effect = DeferredLock
```
manishtomar: Non-blocking: Do you think that test without mocking `DeferredLock` that calls `serialize_and_delay` twice and notices that they get serialized and delayed would be better? I can understand if you don't want to spend much time on the tests since this implementation should be going away soon :).

otter/test/test_cloud_client.py
```diff
@@ -255,6 +269,199 @@ def test_params(self):
...
+        self.addCleanup(set_config_data, {})
+        clock = Clock()
+        bracket = _default_throttler(clock,
+                                     ServiceType.CLOUD_SERVERS, 'delete')
+        d = bracket(lambda: 'foo')
+        clock.advance(499)
+        self.assertNoResult(d)
+        clock.advance(500)
+        self.assertEqual(self.successResultOf(d), 'foo')
+
```
manishtomar: How about a test for serialization?

manishtomar:
LGTM modulo comments :)

otter/test/test_cloud_client.py
radix: Yeah, I didn't put one in this class, but there are other test cases that test that serialization happens -- the ones that explicitly mock out DeferredLock prove that the effects will be serialized. or is there some edge case I'm not thinking of?

manishtomar: No. Its fine to have test there. I was just wondering if you can have a serialization check that doesn't mock `DeferredLock` but I guess it really doesn't make much difference. Its fine :)

otter/test/test_cloud_client.py
radix: I don't like mocking, but I still like avoiding testing third-party code. I'd rather just rely on the fact that DeferredLock works as documented/intended and not write my own tests for its behavior.

manishtomar: I understand. The question was not about testing `DeferredLock` but testing `serialize_and_delay` as a black box without depending on DeferredLock. Its fine either way :)

radix:
Thanks for the reviews @manishtomar :) sorry for the painful PR.

PR #1415, opened 2015-05-06
delete server throttling in worker

### LINKED ISSUE #1411: Throttle deletion
Throttle deletions to approx 300 delete requests per minute in the existing non-convergence codebase.

manishtomar:
Closes #1411. Allows only 1 delete server per second.

### DIFF
--- a/otter/test/worker/test_launch_server_v1.py
+++ b/otter/test/worker/test_launch_server_v1.py
@@ -2227,7 +2227,7 @@ class DeleteServerTests(RequestBagTestMixin, SynchronousTestCase):
         """
         d = delete_server(self.log, self.request_bag, instance_details,
                           clock=self.clock)
-        self.clock.advance(1)
+        self.clock.advance(0.2)
         return d
 
     def test_delete_server_no_lbs(self):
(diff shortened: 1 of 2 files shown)


otter/test/worker/test_launch_server_v1.py
```diff
@@ -2395,13 +2405,48 @@ def test_delete_and_verify_succeeds_if_get_returns_404(self):
...
+
+        # fire first deferred and notice that next treq.delete is still not
+        # called due to delay
+        delete_ds[0].callback(mock.Mock(code=404))
+        self.assertEqual(self.treq.delete.call_count, 1)
+
+        # advance clock and notice next delete called
+        self.clock.advance(1)
+        self.assertEqual(self.treq.delete.call_count, 2)
+        self.successResultOf(ret_ds[0])
```
cyli: Non-blocking - this may be overkill, wondering if it would make sense to assert that the second deferred has no result yet, even though `treq` was called a second time.

cyli:

cyli:
Thanks for finishing this so fast!  This LGTM as soon as jenkins +1's.

radix:
Could you make this configurable, using the same config entries that I used in my throttling PR?

cyli:
@radix Would that be something we can do in a separate PR?  The create server thing throttling in the worker is also not configurable, it might be nice to do both of them.  And this seems to be a high priority item.

radix:
@cyli yeah, absolutely :)

radix:
also fwiw if you're going to hardcode a number I think 0.2 should be fine, since that will guarantee less than 300 deletions per second (much less when also considering the semaphore serializing the requests)

manishtomar:
jenkins please run unit tests

manishtomar:
jenkins please run unit tests

PR #1418, opened 2015-05-06
Using wait_for_expected_group_state instead of sleeping

### LINKED ISSUE #1404: Make cron scheduler tests deterministic

manishtomar:
Closes #1404. The tests now wait for expected group state for maximum scheduled time. This way the tests will wake up at beginning of minute in cron based tests and continue accordingly.

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
+++ b/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
@@ -57,7 +57,7 @@ class ExecuteNegativeSchedulerPolicy(AutoscaleFixture):
     def check_cooldown_over_trigger(self, group):
         self.wait_for_expected_group_state(
             group.id, self.sp_change,
-            60 + self.scheduler_interval, 1, time_scale=False)
+            self.cron_wait_timeout, 1, time_scale=False)
         # This is sometime between 0 to scheduler_interval of minute.
         # Sleeping for another minute + interval + 30s should be time after
         # next execution which must have same state
@@ -121,7 +121,7 @@ class ExecuteNegativeSchedulerPolicy(AutoscaleFixture):
             group.id, cron_policy['id'])
         self.wait_for_expected_group_state(
             group.id, self.gc_min_entities,
-            60 + self.scheduler_interval, 2, time_scale=False)
+            self.cron_wait_timeout, 2, time_scale=False)
 
     def test_system_scheduler_down(self):
         """
--- a/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
+++ b/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
@@ -2,7 +2,6 @@
 Test negative scenarios for execution of at style and
 cron style scheduler policies
 """
-import unittest
 from time import sleep
 
 from cafe.drivers.unittest.decorators import tags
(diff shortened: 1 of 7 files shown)


autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
```diff
@@ -152,7 +154,9 @@ def test_system_scheduler_batch(self):
             sp_list=at_style_policies_list)
         group = response.entity
         self.resources.add(group, self.empty_scaling_group)
-        sleep(self.scheduler_interval * 2)
+        # An extra 1 second to let scheduler processing take place due to
+        # excess events
+        sleep(10 + self.scheduler_interval + 1)
```
cyli: Non-blocking question:  why the change from twice the scheduler interval to 10 + one interval + 1?  Is the scheduler interval just 10?

autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_scheduler_update_group.py
```diff
@@ -143,23 +145,23 @@ def test_system_upd_launch_config_cron_style_scheduler(self):
...
             schedule_cron='* * * * *')
-        sleep(60 + self.scheduler_interval)
         active_servers = self.sp_change + group.groupConfiguration.minEntities
         active_after_scaleup = self.wait_for_expected_number_of_active_servers(
-            group_id=group.id,
-            expected_servers=active_servers)
-        upd_lc_server = set(
-            active_after_scaleup) - set(active_list_b4_upd)
+            group_id=group.id, expected_servers=active_servers,
+            timeout=(60 + self.scheduler_interval), interval_time=2,
```
cyli: Is this enough time?  The original sleeps for 60+scheduler, and then waits for the number to be reached.  The new version just waits 60+scheduler seconds.  Should a few extra seconds be added for a safety margin?

cyli:
I seem to be having trouble running the `test_system_multi_scheduler_webhook_policies` tests

manishtomar:
@cyli Yeah. That test broke during linting. I am fixing it now.

autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_scheduler_update_group.py
manishtomar: Yeah. You are right. There should be a few seconds margin for safety. I'll change it in all the tests. Thanks :)

cyli:
This otherwise LGTM, since the test fix is in the other PR.  :)

autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
manishtomar: No. The 10 account for the 10s in future when the policy gets executed.

manishtomar:
I'll run all scheduler tests and remove "in progress" after that.

manishtomar:
All the tests pass in dev vm

cyli:
@manishtomar cool, all the tests passed for me previously too except for `test_system_multi_scheduler_webhook_policies`, which has now been fixed.

LGTM!  Thank you for making our tests more dependable!

manishtomar:
> Thank you for making our tests more dependable!

@cyli We'll see about that :)

Issue #1141, opened 2015-03-01
Default logging error in retry


manishtomar:
I just spent 2 days debugging code where the error was occurring in code that was within retry. This is not the first time this has happened to me. I realize that I should be careful and test without retry but it is hard to remember that. I think its best if retry takes a log and *always* logs error until told otherwise. This will really help in future debugging.

PR #1420, opened 2015-05-06
wait_for_state on autoscale group with  matchers.

cyli:
This PR replaces `wait_for_N_servers`, `wait_for_deleted_id_removal`, and `wait_for_expected_state` with a function that takes matchers (`wait_for_state`).

It also replaced usages of those functions with this one.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -10,6 +10,9 @@ import random
 
 from characteristic import Attribute, attributes
 
+from testtools.matchers import (
+    ContainsDict, Equals, MatchesAll, MatchesPredicateWithParams)
+
 import treq
 
 from twisted.internet import reactor
@@ -333,30 +336,8 @@ class ScalingGroup(object):
         :return: If the operation succeeds, the same instance of TestResources.
             Otherwise, an exception is raised.
         """
-        def check(state):
-            code, response = state
-            if code == 404:
-                raise BreakLoopException("Scaling group not found.")
-            servers_active = len(response["group"]["active"])
-            if servers_active == servers_desired:
-                return rcs
-
-            raise TransientRetryError()
-
-        def poll():
-            return self.get_scaling_group_state(rcs).addCallback(check)
-
-        return retry_and_timeout(
-            poll, timeout,
-            can_retry=terminal_errors_except(TransientRetryError),
-            next_interval=repeating_interval(period),
-            clock=clock or reactor,
-            deferred_description=(
-                "Waiting for {} servers to go active.".format(
-                    servers_desired
-                )
-            )
-        )
+        return self.wait_for_state(rcs, HasActive(servers_desired),
+                                   timeout=timeout, period=period, clock=clock)
 
     def start(self, rcs, test):
         """Create a scaling group.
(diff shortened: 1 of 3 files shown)


cyli:
jenkins please run unit tests

manishtomar:
This is nice refactor.

otter/integration/lib/autoscale.py
```diff
@@ -299,7 +301,9 @@ def decide(resp):
             if resp.code == 200:
                 return self.treq.json_content(resp).addCallback(
                     lambda x: (200, x))
-            return (404, None)
+
+            return self.treq.content(resp).addCallback(
```
cyli: Reviewer: the reasons for this change are:

1.  was a previously a bug, since it would return 404 no matter what, and the success code was passable (sorry, I forgot about this one when adding success codes)
2.  twisted can sometimes have a hanging issue if we don't read the response from a request and we use the same connection over and over again.  Not as much of an issue here because the tests kill the pool in between every test, but just in case...

otter/integration/lib/autoscale.py
```diff
@@ -430,119 +388,91 @@ def _get_the_ips(the_server_ids):
...
-            clock=reactor,
-            deferred_description=report,
+    def wait_for_state(self, rcs, matcher, timeout=600, period=10, clock=None):
+        """
+        Wait for the state on the scaling group to match the provided matchers,
+        specified by matcher.
+
+        :param rcs: a :class:`otter.integration.lib.resources.TestResources`
+            instance
+        :param matcher: A :mod:`testtool.matcher`, as specified in
```
manishtomar: Do you think using a simple predicate function `state -> bool` would be better option? Known predicates can be implemented.

cyli: I do like that idea - the reason I'd hesitate though is that I like what comes out of `mismatch.describe()` - I would love to change retrier to take a callable instead of a straight description string, or otherwise report the last error.  The callables would each have to implement their own string for why the test failed, otherwise.

And I think @glyph wanted retry to optionally log every intermediate error that causes a retry, and the log messages might be useful for debugging why a test is failing.

cyli:

otter/integration/lib/autoscale.py
manishtomar: Ok. Makes sense. Its also nice that one can compose matchers.

otter/integration/lib/test_autoscale.py
```diff
@@ -54,14 +55,14 @@ def get_scaling_group_state_happy(self, rcs):
 
         This version emulates completion after a period of time.
         """
-
+        self.assertEquals([200], success_codes)
```
manishtomar: I haven't seen assertions like this. Would you mind flipping the arguments?

otter/integration/lib/test_autoscale.py
```diff
@@ -102,6 +103,56 @@ def test_poll_until_timeout(self):
...
+        """
+        :class:`ExcludesServers` fails when the active list contain any or all
+        of servers with the given IDs.
+        """
+        matcher = ExcludesServers(['id1', 'id2'])
+        self.assertNotEqual(
+            None,
+            matcher.match(
+                {'active': [{'id': "id{0}".format(i)} for i in (1, 2)]}),
+            "Complete match succeeds when none should be present."
```
manishtomar: +1. Nice having description :)

otter/integration/lib/test_autoscale.py
cyli: Oops, sorry, good catch.  Not sure why I did this.

manishtomar:
LGTM on the matchers and all related code :). However, I am not able to understand the `success_codes` related change. Will continue looking at it.

otter/integration/lib/autoscale.py
manishtomar: > And I think @glyph wanted retry to optionally log every intermediate error that causes a retry

https://github.com/rackerlabs/otter/issues/1141

cyli:
@manishtomar Thanks for reviewing!  Do you mean https://github.com/rackerlabs/otter/pull/1420#discussion-diff-29813759?

otter/integration/lib/autoscale.py
cyli: Thanks!!!  Yes, sorry, maybe wrong attribution there.  He probably just told me about your ticket :D

cyli:

manishtomar:
@cyli No. I meant https://github.com/rackerlabs/otter/pull/1420/files#diff-fe2069546d4e727681ee430d20663e72R58. It's probably because I haven't seen overall `get_scaling_group_state` code. I'll look at it again today.

manishtomar:
I've gone through `success_codes`. All LGTM :)

cyli:
@manishtomar thanks!  :D

PR #1424, opened 2015-05-08
Add GroupState.status

radix:
This doesn't do anything interesting, just add a `status` parameter/attribute to GroupState. All code that created GroupState has been updated to pass ACTIVE.

In later PRs we'll populate this value from the database properly.


cyli:
LGTM  :)  Thanks!

PR #1425, opened 2015-05-08
Populate GroupState status in model

radix:
This updates the model code to query the `status` from cassandra and include it in the `GroupState` objects.

view_manifest was already including the status in the `get_deleting` case, but I've moved it from the top-level manifest dict into the `GroupState` object, and updated the code in convergence that relied on this.

Still no changes to the REST code to include the status, this is just model-level work.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -13,7 +13,6 @@ from datetime import datetime
 from characteristic import attributes
 
 from effect import Effect, parallel
-from effect.twisted import deferred_performer
 
 from jsonschema import ValidationError
 
@@ -26,6 +25,8 @@ from silverberg.client import ConsistencyLevel
 from toolz.dicttoolz import keymap
 
 from twisted.internet import defer
+
+from txeffect import deferred_performer
 
 from zope.interface import implementer
 
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -684,15 +684,14 @@ class CassScalingGroup(object):
                 group)
 
         def _generate_manifest_group_part(group):
+            status = self._group_status(group['status'], group['deleting'])
             m = {
                 'groupConfiguration': _jsonloads_data(group['group_config']),
                 'launchConfiguration': _jsonloads_data(group['launch_config']),
                 'id': self.uuid,
-                'state': _unmarshal_state(group)
+                'state': _unmarshal_state(group),
+                'status': status.name,
             }
-            if get_deleting:
-                status = self._group_status(group['status'], group['deleting'])
-                m['status'] = status.name
             return m
 
         def check_deleting(group):
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -92,6 +92,10 @@ def serialize_json_data(data, ver):
 # Otherwise it won't.
 #
 # Thus, selects have a semicolon, everything else doesn't.
+
+# NOTE about deleted groups: Make sure every query involving groups *either*
+# filters out deleting=false, *or* includes the `deleting` query so that
+# _unmarshal_state can parse the status correctly.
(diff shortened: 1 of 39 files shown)


radix:

otter/models/cass.py
```diff
@@ -443,11 +443,27 @@ def _jsonloads_data(raw_data):
...
+            return ScalingGroupStatus.lookupByName(status)
+
+
 def _unmarshal_state(state_dict):
     desired_capacity = state_dict['desired']
     if desired_capacity is None:
         desired_capacity = 0
 
+    status = _group_status(state_dict['status'],
+                           state_dict.get('deleting', False))
```
radix: So, this might be a bit weird. Since some queries don't fetch 'deleting' it won't always be in this state_dict, so in those cases I'm assuming we've got an ACTIVE group. any alternative suggestions?

cyli: You mean assuming we don't have a deleting group?  Since status is a different column from deleting?

radix: I think I'm just being dumb, and the obvious solution is to just make sure all the queries also select deleting so this code can be sure that the group is deleting or not.

cyli: radix:  I think the three places that `_unmarshal_state` is called are in `view_manifest`, `view_state`, and `list_scaling_group_states`.  `view_manifest` does query for `deleting` and include it in the response, but I think `view_state` and `list_scaling_group_states` query the index for groups that are not deleting.  So it doesn't have `deleting` in the response, but `deleting` can be assumed to be `False`.

It is kinda hard to tell that though without digging through all the calls to see what the CQL is though, so yeah, adding `deleting` may be more explicit.

radix: Yeah, exactly. Even if we're sure deleting is False in those cases, there's a chance some new case will be introduced and the query will forget to filter out deleting. Thank you for clearing my head about this :)

cyli: Oh that's true - it'd be pretty hard to debug if a future query searched the column family instead of the index.  Thanks for making the contract and the code much clearer!

radix:

cyli:
I think `_cql_list_states` also needs `status` added to it.

otter/models/cass.py
manishtomar: All the queries except manifest already filter deleting groups. It is not required for queries. If the query result does not contain deleting, then you can be sure that group is NOT deleting and you need to worry only about status which `_group_status` and above code seems to be doing.

radix:
okay, added status to _cql_list_states and added an explicit test. thanks @cyli :)

radix:

radix:
integration passed (I'm logging this here since running the unit tests will clobber it): https://as-jenkins.rax.io/job/otter-pr-integration/169/

radix:
jenkins please run unit tests

radix:
jenkins please run unit tests

otter/models/cass.py
```diff
@@ -444,11 +448,27 @@ def _jsonloads_data(raw_data):
     return data
 
 
+def _group_status(status, deleting):
+    if deleting:
+        return ScalingGroupStatus.DELETING
+    else:
+        # TODO: #1304
```
cyli: Non-blocking nitpick - since this would nice to have, but is already tracked by an issue, can this comment be removed?  Alternately can it instead be moved next to the 'DISABLED' part of these conditionals?

radix: Actually it applies to both the `None` and the `'DISABLED'` since the ticket is about migrating both.

cyli:
This LGTM - up to you whether you want to delete the comment, but this is fine to merge!  Thanks - really looking forward to having the error status!

radix:
I think I'll leave it there until it's fixed. Thanks for the review @cyli :D

PR #1427, opened 2015-05-08
script to run multiple cafe modules parallely

manishtomar:
This script runs each module given in a sub-process reducing overall time taken. This can be very useful to reduce time taken to run scheduler tests. Running all scheduler takes 12m24s. Normally they would take 37m


cyli:
This is awesome!!  Just ran it on some tests, and it works great.  LGTM when +1's.  :)

PR #1428, opened 2015-05-09
Active to error tests

cyli:
Fixes #1320 (CATC11)
Fixes #1321 (CATC12)
Fixes #1322 (CATC13)

These are the tests that assert that autoscale deletes servers that have gone into error from active, and removes them from their relevant CLBs.

PR #1429, opened 2015-05-09
Copy over select test methods from other test case classes

### LINKED ISSUE #1410: Fix CATC-14: For CLB versions of 4-12, check that all new servers created are on the CLBs
This is an addition the work done in #1409, since those tests do not pass yet anyway.

### One possible way to do this:
It'd be great to change the tests so that the CLB tests do not inherit from the other test case, but is an entirely new case that decorates the other test methods and adds additional functionality (checking that all active servers are on the CLB) at the end before the scaling group is cleaned up.

But meta-copying over the test methods from the other test case is slightly problematic, since they depend upon helper methods within the class.

Composing/decorating may require more refactoring.

cyli:
So we don't have to structure our test class methods so they need composing.

Also, this way we can selectively copy and skip methods.

This PR has been split out from #1428.

This also fixes #1410 because the test methods copied over have their CLB info checked.

### DIFF
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -162,18 +162,6 @@ def tag(*tags):
     return decorate
 
 
-def skip(reason):
-    """
-    Decorator that sets the skip property on to a function.
-
-    This should be added upstream to Twisted's trial.
-    """
-    def decorate(function):
-        function.skip = reason
-        return function
-    return decorate
-
-
 class TestConvergence(unittest.TestCase):
     """This class contains test cases aimed at the Otter Converger."""
     timeout = 1800
@@ -783,9 +771,9 @@ class ConvergenceSet1WithCLB(unittest.TestCase):
 
         if any(tag in tags for tag in
                ["CATC-0{0:02d}".format(i) for i in range(4, 9)]):
-            wrapper = skip(
+            wrapper.skip = (
                 "Autoscale does not clean up servers deleted OOB yet. "
-                "See #881.")(wrapper)
+                "See #881.")
         return (name, wrapper)
 
 
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -747,9 +747,9 @@ class ConvergenceSet1(unittest.TestCase):
         )
 
 
-class ConvergenceSet1WithCLB(unittest.TestCase):
+class ConvergenceTestsWith1CLB(unittest.TestCase):
     """
-    Class for CATC 4-12 that run with CLB.
+    Tests for convergence that require a single CLB.
     """
     timeout = 1800
 
@@ -780,16 +780,23 @@ class ConvergenceSet1WithCLB(unittest.TestCase):
         )
 
     @classmethod
-    def _copy_catc_4_to_12(cls, name, method):
+    def _only_oob_del_and_error_tests(cls, name, method):
         """
-        To be used to copy over methods from ConvergenceSet1 using
-        :func:`duplicate_test_methods`.  Note that the methods copied
-        should all return the scaling group, so that the group's active
-        servers can be checked against the CLB.
+        To be used by :func:`copy_test_methods` to filter only certain non-CLB
+        tests (the ones testing out of band deletions, servers going into
+        error, servers timing out from builds), and ensure that active servers
+        on the group, when the test has finished, are all properly on the CLB.
+
+        Note that the methods copied should all return the scaling group, so
+        that the group's active servers can be checked against the CLB.
+
+        Also note that this filters out only the tests tagged CATC-004 through
+        CATC-013, because those where the numbers in the original test plan
+        corresponding to the OOB-delete/error test cases.
         """
         tags = getattr(method, 'tags', ())
         if not any(tag in tags for tag in
-                   ["CATC-0{0:02d}".format(i) for i in range(4, 13)]):
+                   ["CATC-0{0:02d}".format(i) for i in range(4, 14)]):
             return None
 
         @wraps(method)
@@ -815,5 +822,5 @@ class ConvergenceSet1WithCLB(unittest.TestCase):
 
 
 copy_test_methods(
-    ConvergenceSet1, ConvergenceSet1WithCLB,
-    filter_and_change=ConvergenceSet1WithCLB._copy_catc_4_to_12)
+    ConvergenceSet1, ConvergenceTestsWith1CLB,
+    filter_and_change=ConvergenceTestsWith1CLB._only_oob_del_and_error_tests)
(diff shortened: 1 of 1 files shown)


otter/integration/tests/test_convergence.py
```diff
@@ -508,6 +509,7 @@ def test_reaction_to_oob_server_deletion_below_min(self):
                 self.rcs, self.scaling_group, N_SERVERS / 2)(
                 self.scaling_group.trigger_convergence))
         )
+    test_reaction_to_oob_server_deletion_below_min.tags = ["CATC-004"]
```
glyph: One aesthetic suggestion: make this into a decorator, so you can just say `@tagged("CATC-004")` before the test method.

cyli:

cyli:

cyli:

cyli:

otter/integration/tests/test_convergence.py
```diff
@@ -150,6 +150,55 @@ def wrapper(*args, **kwargs):
...
+
+    :param class from_class: The test case to copy from
+    :param class to_class: The test case to copy to
+    :param callable filter_and_change: A function that takes a test name
+        and test method, and returns a tuple of `(name, method)`
+        if the test method should be copied. None else.  This allows the
+        method name to change, the method to be decorated and/or skipped.
+    """
+    for name, attr in from_class.__dict__.items():
+        if name.startswith('test_') and isinstance(attr, type(lambda: None)):
```
manishtomar: How about using [`callable`](https://docs.python.org/2/library/functions.html#callable) function instead of getting type of empty function?

cyli: That's a good idea, thanks!  :)  I was thinking of trying to get functions specifically, but it's very unlikely something named `test_` would be a class, so `callable` should work.

otter/integration/tests/test_convergence.py
```diff
@@ -150,6 +150,55 @@ def wrapper(*args, **kwargs):
...
+    ``to_class`` will first be decorated before being set.
+
+    :param class from_class: The test case to copy from
+    :param class to_class: The test case to copy to
+    :param callable filter_and_change: A function that takes a test name
+        and test method, and returns a tuple of `(name, method)`
+        if the test method should be copied. None else.  This allows the
+        method name to change, the method to be decorated and/or skipped.
+    """
+    for name, attr in from_class.__dict__.items():
```
manishtomar: Non-blocking nitpick: Using `dir` function may be better since dunder methods are supposed to be private?

otter/integration/tests/test_convergence.py
```diff
@@ -150,6 +150,55 @@ def wrapper(*args, **kwargs):
...
+
+    This should be added upstream to Twisted trial.
+    """
+    def decorate(function):
+        function.tags = tags
+        return function
+    return decorate
+
+
+def skip_me(reason):
```
manishtomar: +1. Love the decorators! Thanks to you and @glyph :)

otter/integration/tests/test_convergence.py
```diff
@@ -689,12 +747,11 @@ def test_group_config_update_triggers_convergence(self):
         )
 
 
-class ConvergenceSet1WithCLB(ConvergenceSet1):
+class ConvergenceSet1WithCLB(unittest.TestCase):
```
manishtomar: Non-blocking: Not sure if `Set1` in the name has got any significance but `ConvergenceWithCLBTests` might be a better name?

otter/integration/tests/test_convergence.py
```diff
@@ -721,3 +778,42 @@ def setUp(self):
...
+    def _copy_catc_4_to_12(cls, name, method):
+        """
+        To be used to copy over methods from ConvergenceSet1 using
+        :func:`duplicate_test_methods`.  Note that the methods copied
+        should all return the scaling group, so that the group's active
+        servers can be checked against the CLB.
+        """
+        tags = getattr(method, 'tags', ())
+        if not any(tag in tags for tag in
+                   ["CATC-0{0:02d}".format(i) for i in range(4, 13)]):
```
manishtomar: Non-blocking nitpicky suggestion: A set might make it clearer: `if set(tags).isdisjoint(set("CATC-0{0:02d}".format(i) for i in range(4, 13)))`. Maybe not. Its fine either way :) Also would be nice to have separate function to generate CATC names based on it's range? 
```python
def catc_names(beg, end):
   return ["CATC-0{0:02d}".format(i) for i in range(beg, end)]
```
This can be used below too unless this is not going to be used anywhere :). Again, both these suggestions are non-blocking nitpicks :).

otter/integration/tests/test_convergence.py
cyli: I don't feel strongly about this, but lean towards I `__dict__` because it seems like less work.  :)   `dir` gives all inherited members and all the python magic members, whereas `__dict__` gives only the members defined in the class.  e.g.:

```
>>> class A(object):
...    def a(self):
...       pass
... 
>>> class B(A):
...   def b(self):
...      pass
... 
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'a', 'b']
>>> B.__dict__
dict_proxy({'__module__': '__main__', 'b': <function b at 0x107b61c80>, '__doc__': None})
```

This definitely isn't currently a problem, since this method is written so we can avoid inheritance of test cases anyway, and the function already filters out anything that doesn't start with `test_` and further takes a callable that can customize filtering anyway.  So using `dir` is fine, and not an issue, `__dict__` just seemed simpler (less to think about if there is any test inheritance, and also I don't have to use `getattr`)

I am happy to be overruled if you feel more strongly about it. :D

otter/integration/tests/test_convergence.py
manishtomar: FYI: `callable` doc says it returns `False` for classes in case you were worried.

otter/integration/tests/test_convergence.py
cyli: I was going to rename this one ConvergenceWith1CLB in the next PR - can do this now :D

otter/integration/tests/test_convergence.py
cyli: Oh cool!  good point, thanks!

otter/integration/tests/test_convergence.py
manishtomar: Oh ok. Yeah, I forgot `dir` returns inherited and dunder methods. Using `__dict__` is fine :)

manishtomar:
LGTM modulo nitpicks above :)

cyli:

cyli:

cyli:
Integration tests passed.  Travis is taking a while, so running unit tests on jenkins.

cyli:
jenkins please run unit tests

PR #1430, opened 2015-05-09
Delete a load balancer and scale

### LINKED ISSUE #1328: CATC20: Scale group with CLB deleted partway
Scale group with CLB deleted partway
	- Create a group and scale by x
	- Delete the CLB or otherwise cause it to return 404
	- Attempt to scale up
	- validate that the group enters a TBD error state

In order to make this convergence specific:
- remove a server from each LB out of band and expect them to be readded

Prereq: There MUST be a defined backoff/Invalid group behavior defined.
Prereq: We MUST be able to use Mimic to delete a CLB

cyli:
This creates a group with either one or two load balancers, deletes one load balancer, and attempts to scale.

Fixes #1328.

### DIFF
--- a/otter/integration/lib/cloud_load_balancer.py
+++ b/otter/integration/lib/cloud_load_balancer.py
@@ -155,11 +155,13 @@ class CloudLoadBalancer(object):
                 .addCallback(self.treq.json_content)
                 .addCallback(record_results))
 
-    def delete(self, rcs):
+    def delete(self, rcs, success_codes=None):
         """Stops and deletes the cloud load balancer.
 
         :param TestResources rcs: The resources used to make appropriate API
             calls with.
+        :param list success_codes: A list of HTTP status codes to count as
+            a successful call.
         """
         return (
             self.treq.delete(
@@ -169,7 +171,9 @@ class CloudLoadBalancer(object):
                 ),
                 headers=headers(str(rcs.token)),
                 pool=self.pool
-            ).addCallback(check_success, [202, 404])
+            ).addCallback(
+                check_success,
+                [202, 404] if success_codes is None else success_codes)
         ).addCallback(lambda _: rcs)
 
     def wait_for_nodes(self, rcs, matcher, timeout, period=10, clock=None):
--- a/otter/integration/lib/cloud_load_balancer.py
+++ b/otter/integration/lib/cloud_load_balancer.py
@@ -161,7 +161,8 @@ class CloudLoadBalancer(object):
         :param TestResources rcs: The resources used to make appropriate API
             calls with.
         :param list success_codes: A list of HTTP status codes to count as
-            a successful call.
+            a successful call.  If not provided, defaults to ``[202, 404]``
+            (404 is a successful result because deletes should be idempotent).
         """
         return (
             self.treq.delete(
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -130,7 +130,7 @@ class TestHelper(object):
             :class:`otter.integration.lib.resources.TestResources`
         :param ScalingGroup group: An instance of
             :class:`otter.integration.lib.autoscale.ScalingGroup` to start -
-            this group should not have been started already.
+            this group should have been started already.
         :param int num: The number of servers to delete out of band.
         """
         def decorated(function):
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -834,6 +834,64 @@ class ConvergenceTestsWith1CLB(unittest.TestCase):
                 "See #881.")
         return (name, wrapper)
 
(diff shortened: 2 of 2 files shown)


cyli:

otter/integration/tests/test_convergence.py
```diff
@@ -827,7 +856,134 @@ def wrapper(self, *args, **kwargs):
...
+
+        return (
+            self.helper.start_group_and_wait(group, self.rcs)
+            .addCallback(_delete_a_clb_and_scale, self.helper, group,
+                         scale_by=1, delete_clb=delete_clb)
+            .addCallback(
+                group.wait_for_state,
+                MatchesAll(
+                    ContainsDict({
+                        'pendingCapacity': Equals(1),
```
cyli: @radix: Looks like my assumption about this was wrong - active goes to zero, and pending becomes 2 again, because since the original server is not on any load balancer anymore.  Is that correct?  Is it weird that if the load balancer goes away, the group's active servers go to zero, but we don't delete the servers from nova?

cyli: I mean it also totally makes sense because the servers are broken at that point.  So they're not active.  But it also seems strange that we're telling users they're gone.  I don't know what else we'd do though.

radix: Yeah, I don't know either. We could figure out a way to extend the REST API to provide more information about our knowledge of the world, perhaps? Like, maybe instead of just updating active/pending on every gathering, we could also store all the info about servers + lbs we find?

cyli: :|  Maybe?  I don't know.   Do you think this is reasonable behavior for now?  If so I'll change the test and we can file a tech debt ticket or something?

cyli:

otter/integration/lib/cloud_load_balancer.py
```diff
@@ -155,11 +155,13 @@ def record_results(resp):
...
                 .addCallback(record_results))
 
-    def delete(self, rcs):
+    def delete(self, rcs, success_codes=None):
         """Stops and deletes the cloud load balancer.
 
         :param TestResources rcs: The resources used to make appropriate API
             calls with.
+        :param list success_codes: A list of HTTP status codes to count as
+            a successful call.
```
kivattik: Doc-fix: If omitted, the success codes default to 202 and 404.

otter/integration/tests/test_convergence.py
```diff
@@ -125,6 +125,13 @@ def oob_delete_then(self, rcs, scaling_group, num):
         Return a decorator that wraps a function call with logic to out-of-band
```
kivattik: Documentation: It might be more clear if this is stated to return a function that takes a function, rather than calling it a decorator.

otter/integration/tests/test_convergence.py
```diff
@@ -125,6 +125,13 @@ def oob_delete_then(self, rcs, scaling_group, num):
...
         delete (not disown) some number of servers, and verifies that the
         servers are deleted and cleaned up from CLBs.
+
+        :param TestResources rcs: An instance of
+            :class:`otter.integration.lib.resources.TestResources`
+        :param ScalingGroup group: An instance of
+            :class:`otter.integration.lib.autoscale.ScalingGroup` to start -
+            this group should not have been started already.
```
kivattik: Confirm that group should be un-started?

otter/integration/tests/test_convergence.py
```diff
@@ -881,6 +888,28 @@ def _catc_tags(start_num, end_num):
     return ["CATC-0{0:02d}".format(i) for i in range(start_num, end_num + 1)]
 
 
+def _delete_a_clb_and_scale(rcs, helper, group, scale_by, delete_clb=None):
+    """
+    1. Delete the first load balancer using the provided function (it may set
+       the load balancer to PENDING_DELETE instead, for example)
+    2. Scale.
+
```
kivattik: doc-fix: Would it be helpful to add other parameter descriptions?

otter/integration/tests/test_convergence.py
```diff
@@ -953,7 +982,133 @@ def wrapper(self, *args, **kwargs):
                 "See #881.")
         return (name, wrapper)
 
+    @tag("CATC-020")
+    def test_delete_all_loadbalancers_and_scale_up(self, delete_clb=None):
+        """
+        CATC-020-a. This will also tested with starting with 2 load balancers.
```
kivattik: Minor doc-fix: This will also be tested starting with 2 load balancers.

kivattik:
Wonderful :) LGTM 
(And nice quick work on the doc nit-picks.)

otter/integration/tests/test_convergence.py
cyli: Yes, definitely, thanks!  :D  added.

PR #1432, opened 2015-05-11
Fix LogErr when passed None as failure

radix:
When LogErr was passed None, it wouldn't properly get the exception from the context, since the exception context was lost by the time the intent was performed. This changes the code so that LogErr checks if failure is None and explicitly creates a Failure object.

I also updated transform_eq to be a bit more useful when debugging - it now __repr__s to have a lot more useful information.

### DIFF
--- a/otter/test/log/test_intents.py
+++ b/otter/test/log/test_intents.py
@@ -18,7 +18,7 @@ from otter.log.intents import (
     get_log_dispatcher,
     msg,
     with_log)
-from otter.test.utils import mock_log, transform_eq
+from otter.test.utils import CheckFailureValue, mock_log
 
 
 class LogDispatcherTests(SynchronousTestCase):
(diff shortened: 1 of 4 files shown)


otter/test/log/test_intents.py
```diff
@@ -84,6 +84,25 @@ def test_err(self):
...
+        try:
+            raise RuntimeError('original')
+        except RuntimeError:
+            eff = err(None, "why")
+        try:
+            raise RuntimeError('performing')
+        except RuntimeError:
+            sync_perform(self.disp, eff)
+        self.log.err.assert_called_once_with(
+            transform_eq(lambda f: (f.type, str(f.value)),
```
manishtomar: Non-blocking thought for FYI only: You could use `CheckFailureValue` in this case but even better would be to implement that using `transform_eq` :)

manishtomar:
LGTM once build passes :)

otter/test/log/test_intents.py
radix: d'oh, totally forgot about that!

radix:
jenkins please run unit tests

radix:
jenkins please run unit tests

radix:
ok, gonna merge this since jenkins is cool with it. Thanks for the review @manishtomar :)

PR #1433, opened 2015-05-12
Test for if Nova reports a 500 and creates a server anyway, and for Nova erroring a building server

### LINKED ISSUE #1318: CATC9: Server Transitioned to Error (STE): STE from Building
Server Transitioned to Error (STE): STE from Building
	- Validate that the expected group state is reached when servers error on build

### LINKED ISSUE #1337: CATC29:  Simulate accidental overshoot due to Nova 500 without an actual failure
Simulate accidental overshoot due to Nova 500 without an actual failure
	- Create a scaling group
	- Scale up by X
	- Use mimic to cause Nova to return a 500 but build the server anyway
	- Confirm that any excess builds are stopped and the correct number of servers is produced

Prereq: Mimic MUST be capable of returning a Nova 500 and creating the server anyway.

cyli:
Fixes #1337 
Fixes #1318

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -49,7 +49,7 @@ def extract_active_ids(group_status):
 
 def create_scaling_group_dict(
     image_ref=None, flavor_ref=None, min_entities=0, name=None,
-    max_entities=25, use_lbs=None, server_name=None
+    max_entities=25, use_lbs=None, server_name_prefix=None
 ):
     """This function returns a dictionary containing a scaling group's JSON
     payload.  Note: this function does NOT create a scaling group.
@@ -71,8 +71,9 @@ def create_scaling_group_dict(
         ``otter.lib.CloudLoadBalancer``.  However, you can get the dicts by
         invoking the o.l.CLB.scaling_group_spec() method on such objects.  If
         not given, no load balancers will be used.
-    :param str server_name: Specifies a server name to autoscale - autoscale
-        will use this as the prefix of all server names created by the group.
+    :param str server_name_prefix: Specifies a server name in the server
+        args that get passed to autoscale - autoscale will use this as the
+        prefix of all server names created by the group.
 
     :return: A dictionary containing a scaling group JSON descriptor.  Inside,
         it will contain a default launch config with the provided (or assumed)
@@ -109,8 +110,8 @@ def create_scaling_group_dict(
     if use_lbs:
         launch_config_args["loadBalancers"] = use_lbs
 
-    if server_name is not None:
-        launch_config_args["server"]["name"] = server_name
+    if server_name_prefix is not None:
+        launch_config_args["server"]["name"] = server_name_prefix
 
     return obj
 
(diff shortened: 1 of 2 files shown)


cyli:
I've verified that mimic is doing the right thing by checking otter and mimic logs.  There were only ever 2 creates in otter, and mimic returned 1 `500` on create and 1 `202`.

cyli:

cyli:

cyli:

cyli:

cyli:

otter/integration/lib/autoscale.py
```diff
@@ -49,7 +49,7 @@ def extract_active_ids(group_status):
 
 def create_scaling_group_dict(
     image_ref=None, flavor_ref=None, min_entities=0, name=None,
-    max_entities=25, use_lbs=None
+    max_entities=25, use_lbs=None, server_name=None
```
sam-falvo: Non-blocking :: Recommendation: change to `server_name_prefix` or some similarly suggestive term.

sam-falvo:
I know this is still in-progress, but what I see so far looks good to me.

cyli:

otter/integration/tests/test_convergence.py
```diff
@@ -825,6 +824,35 @@ def test_group_config_update_triggers_convergence(self):
...
+            criteria=[{"server_name": "build-to-error.*"}],
+            behaviors=[
+                {"name": "error", "parameters": {}},
+                {"name": "default"}
+            ])
+        d.addCallback(
+            lambda _: self.helper.start_group_and_wait(group, self.rcs))
+        d.addCallback(wait_for_servers, pool=self.helper.pool, group=group,
+                      matcher=HasLength(2), timeout=600)
+        d.addCallback(lambda _: group)
```
kivattik: Is there a reason to return the group in the last callback that I'm missing? Otherwise, this looks good to me.

radix: likewise. also not sure about why this is necessary.

radix:
LGTM other than that minor nit! :)

otter/integration/tests/test_convergence.py
cyli: @kivattik @radix This is one of the tests that also need to be run with a CLB - https://github.com/rackerlabs/otter/blob/master/otter/integration/tests/test_convergence.py#L962

I return the group so that at the end, the active servers can be verified to be on the CLB.  We could also just set the group to be on the test case instead, and change that requirement that all tests that are copied over should have set `self.group`.  This just seemed slightly more explicit, although I'm not particularly happy about the requirement that THESE tests return a group.  I'm not sure whether setting an attribute on the test case or returning a group is better.

Other suggestions would also be welcome.

radix: Oh right, completely forgot about that stuff.

I think this PR should be merged, though maybe that whole test-copying mechanism could be done differently. I haven't really thought about it much.

radix: (and fwiw I definitely prefer returning values to setting attributes on self)

PR #1434, opened 2015-05-12
Add a mimic API for specifying sequenced behaviors

cyli:
Also, take a server name in the creation json helper so that we can use the server name as the criteria to which we can apply behavior.

This is needed by #1433


radix:
LGTM :)

PR #1435, opened 2015-05-12
Nova utilities to verify the number of servers on a scaling group.

cyli:
So we can verify the number of servers on the scaling group.

This is needed by #1433

### DIFF
--- a/otter/integration/lib/nova.py
+++ b/otter/integration/lib/nova.py
@@ -5,9 +5,16 @@ from characteristic import Attribute, attributes
 
 import treq
 
-from twisted.internet.defer import gatherResults
+from twisted.internet import reactor
+from twisted.internet.defer import gatherResults, inlineCallbacks, returnValue
 
+from otter.util.deferredutils import retry_and_timeout
 from otter.util.http import check_success, headers
+from otter.util.retry import (
+    TransientRetryError,
+    repeating_interval,
+    terminal_errors_except
+)
 
 
 @attributes(["id", "pool",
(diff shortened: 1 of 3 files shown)


otter/integration/lib/nova.py
```diff
@@ -83,3 +90,49 @@ def delete_servers(server_ids, rcs, pool=None, _treq=treq):
...
+        headers=headers(str(rcs.token)),
+        pool=pool
+    ).addCallback(check_success, [200]).addCallback(_treq.json_content)
+
+
+def wait_for_servers(rcs, pool, group, matcher, timeout=600, period=10,
+                     clock=None, _treq=treq):
+    """
+    Wait until Nova reaches a particular state with regards to the servers for
+    the given group.
```
radix: Usually I don't ask for much from docstrings in test code, but I think this could use a bit more information about what "a particular state" means -- i.e. documentation that `matcher` is a testtools matcher.

otter/integration/lib/nova.py
```diff
@@ -83,3 +90,49 @@ def delete_servers(server_ids, rcs, pool=None, _treq=treq):
...
+    """
+    @inlineCallbacks
+    def do_work():
+        servers = yield list_servers(rcs, pool, _treq=_treq)
+        servers_in_group = [
+            server for server in servers['servers']
+            if group.group_id in (
+                server['metadata'].get("rax:autoscale:group:id", None),
+                server['metadata'].get("rax:auto_scaling_group_id", None))
```
radix: Is there a reason you need to check both? I'd rather just check the "new" one, and stop assigning the old one for convergence tenants.

cyli: Good point, thank you!

cyli:
@radix: I think I've addressed all the review comments - might be ready for a re-review :)

radix:
Oh yeah, it LGTM! Thanks @cyli :)

cyli:
@radix: thanks!!

PR #1436, opened 2015-05-12
cron cooldown test fix

manishtomar:
The below test failed sometimes because policy could get triggered after 65 seconds. This can happen if policy executed at beginning of minute and next execution happened at next minute's 10s due to scheduler processing events every 10s. Now, changed the cooldown to be higher than minute + scheduler interval

### DIFF
--- a/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
+++ b/autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
@@ -71,14 +71,15 @@ class ExecuteNegativeSchedulerPolicy(AutoscaleFixture):
     @tags(speed='slow', convergence='yes')
     def test_cron_style_when_policy_cooldown_over_trigger_period(self):
         """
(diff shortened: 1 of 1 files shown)


autoscale_cloudroast/test_repo/autoscale/system/schedule_policies/test_system_execute_scheduler_negative.py
```diff
@@ -71,14 +71,15 @@ def check_cooldown_over_trigger(self, group):
     @tags(speed='slow', convergence='yes')
     def test_cron_style_when_policy_cooldown_over_trigger_period(self):
         """
-        When policy cooldown is set to be greater than a minute by few seconds
+        When policy cooldown is set to be greater than a minute by
```
cyli: "greater than (1 minute + the scheduler interval) by a few seconds?"

manishtomar: Sounds good :)

cyli:
LGTM!  Just a docstring nitpick.  :D  Thanks for fixing this!

PR #1437, opened 2015-05-12
increase timeout as per Cloud DNS request.

wirehead:
Cloud DNS told us this morning that we're not throttling enough.


stretchservers:

cyli:
The tests currently assume a 0.2 second delay, so they probably need to be changed in `test_launch_server_v1.DeleteServerTests._delete_server` and `test_launch_server_v1.DeleteServerTests._delete_and_verify`

radix:
I'll fix this, and the convergence case too.

radix:
ok, updated.

manishtomar:
LGTM once build passes :)

PR #1438, opened 2015-05-12
cron update test fix

manishtomar:
After creating cron policy the test was waiting for policy to get executed *and* servers to become active within cron wait time interval. Now, separated the two. It first waits for policy to get executed and then
waits for servers to become active.


cyli:
That makes a lot of sense - thanks @manishtomar!  LGTM.

Issue #885, opened 2015-01-05
Scaling group status


manishtomar:
The original description is https://github.com/rackerlabs/otter/issues/885#issuecomment-98007717.  There have been discussions about this and following below tasks have been decided: 
- [x] When starting convergence cycle, set group status to ACTIVE  if it is not already active. After executing set it to ERROR if group has errored.
- [x] Provide status in API (either in manifest view or state view)

lvh:
So, FYI, I'm reasonably sure the answer for groups is:

- `ACTIVE`: working as usual.
- `ERROR`: discarding automated input, such as from the observer. You get here because we believe a group is wedged. You get out of this by manually hitting a converge endpoint, which does not check the status before starting, but does *set* the status to `ACTIVE` if it believes it completed without user errors in the group.

... keeping in mind that data gathering, and possibly convergence as a whole, needs to be a tenant-level thing (although that's a different discussion).

manishtomar:
> ERROR: discarding automated input, such as from the observer.

What about scheduled policy execution when group is in ERROR? I am kind of leaning towards making it ACTIVE and trying to execute again since we know group ERRORs can sometimes be transient. Say if launch config's image was getting copied when group went into ERROR and there is scheduled policy in an hour. I don't have strong opinions about it though.

lvh:
Same deal. You start a convergence cycle, and set `ACTIVE` if it worked, and `ERROR` if it failed. You want the state transition on the exit side.

lvh:
We probably can't reasonably answer this question without first having a real answer about converge being tenant- or group-level first.

manishtomar:
W.r.t this I think the answer is converge being group-level since each group has different launch config and only one of them maybe invalid.

lvh:
Convergence is either group-level or tenant-level, it can't be different things with respect to different things. Convergence being group-level entirely has the issues @cyli and I highlighted re: gathering.

manishtomar:
gathering is more of implementation detail rather than issue. Its just more efficient to converge all the groups because of the way gathering _happens_

lvh:
I don't think that's true. It makes a pretty big difference if you're deciding to automatically put groups into `ACTIVE` when you hit the big "converge" button, for example.

lvh:
I also don't think it's an implementation detail, because there's not really an implementation yet. I'm saying we should decide semantics, and that's the exact opposite of an implementation detail.

[...]

PR #1440, opened 2015-05-12
reactivate group when convergence completes

radix:
As described in #885 -- when convergence completes with SUCCESS, put the group back into ACTIVE (if it was in ERROR).


cyli:
LGTM

PR #1445, opened 2015-05-13
Group status change logs to cloud feeds

manishtomar:
Adding logs to cloud feeds when group status is changed after executing convergence. Currently, the code just combines all steps' all reasons as comma separated string for the event message which may not be the best thing to do. Would like other's opinions about it.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -32,7 +32,7 @@ from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
 from otter.convergence.model import ServerState, StepResult
 from otter.convergence.planning import plan
-from otter.log.cloudfeeds import cf_msg
+from otter.log.cloudfeeds import cf_err, cf_msg
 from otter.log.intents import err, msg, with_log
 from otter.models.intents import (
     DeleteGroup, GetScalingGroupInfo, ModifyGroupState, UpdateGroupStatus)
@@ -154,7 +154,7 @@ def execute_convergence(tenant_id, group_id,
     elif worst_status == StepResult.FAILURE:
         yield Effect(UpdateGroupStatus(scaling_group=scaling_group,
                                        status=ScalingGroupStatus.ERROR))
-        yield cf_msg(
+        yield cf_err(
             'group-status-error', status=ScalingGroupStatus.ERROR.name)
 
     yield do_return(worst_status)
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -32,6 +32,7 @@ from otter.convergence.effecting import steps_to_effect
 from otter.convergence.gathering import get_all_convergence_data
 from otter.convergence.model import ServerState, StepResult
 from otter.convergence.planning import plan
+from otter.log.cloudfeeds import cf_msg
 from otter.log.intents import err, msg, with_log
 from otter.models.intents import (
     DeleteGroup, GetScalingGroupInfo, ModifyGroupState, UpdateGroupStatus)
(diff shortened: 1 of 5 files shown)


otter/convergence/service.py
```diff
@@ -148,13 +149,30 @@ def execute_convergence(tenant_id, group_id,
...
 
 
+def worst_reasons_from_sorted_results(results):
+    """
+    Return worst reasons from list of results sorted based on severity of
+    StepResults
+    """
+    for result, reasons in results:
+        if result == StepResult.FAILURE:
+            yield ','.join(reasons)
```
radix: This won't work because we sometimes put non-strings in the reasons list

radix: also, those non-strings can have exception info, which probably isn't something we want to expose to users

radix:
Looks good! I think we can probably just leave out the `reasons` for now until we come up with a better solution for useful stuff to show the user. We probably need some mapping of exception types to user-relevant information

manishtomar:
I kinda wanted the reasons thing to be there. Otherwise, the CF log is not really useful. Can we discuss and decide here or should I create another ticket to discuss that?

cyli:
We should probably have some way to translate exceptions to human-readable cloud feeds formats.  I know a lot of the Nova 400 errors would be really useful to surface to users.

manishtomar:
This also makes me wonder if it makes sense to have list of reasons as return value instead of single reason. I haven't noticed multiple reasons use-case yet

radix:
@manishtomar `reasons`, as it stands, is not suitable to show to end users. I think it's fine to merge this PR without `reasons` and add in some more useful info later -- having the event in CF that says the group was put into ERROR is still useful even without that stuff.

and as for multiple reasons, well, we _definitely_ end up with multiple reasons when multiple steps get executed and e.g. fail. so we need to handle multiple reasons anyway.


Individual steps being able to return multiple reasons is just a flexibility in the API in case we ever do need to do multiple (we almost had a use case for that with a compound throttling step, but we didn't end up implementing that).

radix:
and yeah, I agree with translating exceptions to end-user-suitable messages -- if appropriate. Something similar to the thing @manishtomar for converting exceptions to json-serializable would work here -- having some generic function `present_error(exception)` that converts an exception instance to a end-user  version would be nice, with the default just being to not show anything. But I think that should be done in a separate PR -- this code is very close to mergeable as it stands, no reason to put it off and make a bigger branch.

manishtomar:
Created #1448 to decide this.

otter/log/spec.py
```diff
@@ -20,6 +20,8 @@
     "execute-convergence": "Executing convergence",
     "execute-convergence-results": (
         "Got result of {worst_status} after executing convergence"),
+    "group-status-active": "Group has become ACTIVE",
+    "group-status-error": "Group's status is changed to ERROR",
```
radix: These two messages should probably be made consistent

radix:
Other than my comment this looks good to merge. I have one other question: it doesn't look like the value of otter_msg_type (e.g. 'group-status-active' or 'group-status-error') makes it into the CF event. Is that right? I think it would be good if we included these machine-readable event names in the CF feed so that end-users have something to use as a key in automation (other than the English text in the message).

I guess this would require a change to our CF schema?

manishtomar:
> I guess this would require a change to our CF schema?

@radix Yeah. Also, msg types were originally intended for internal use only. I wouldn't recommend exposing it to customers until we are really sure we want to support this for long time. Since CF schema is versioned, we can start off without msg types and add it in next schema.

manishtomar:
Also, I just realized that group status to ERROR message should probably go as error but that will need `Failure` which we can do as part of #1448 implementation.

radix:
I think you're right - maybe we should pass isError=True? Or perhaps have another utility function called cf_err that adds that?

manishtomar:
I am working on another PR #1450 that will fix parts of CF logging and this can be merged/changed after that.

manishtomar:
@radix About writing the message as error, I was thinking of having below two functions:
```python
def cf_err_no_failure(msg, **fields):
    """
    Log cloud feed error event without failure
    """
    return msg_effect(msg, isError=True, cloud_feed=True, **fields)

def cf_err(failure, msg, **fields):
    """
    Log cloud feed error event with failure
    """
    return err_effect(failure, msg, cloud_feed=True, **fields)
```
The first one is used in this case when we want to log it as error but don't have failure to extract reason from. The second is normal default case. This would assume that we are using `ErrorFormattingWrapper` to extract reason from failure when using `cf_err` based on #1450 implementation. What do you think?

radix:
@manishtomar sounds good. for terseness how about `cf_fail(failure, msg, **fields)` and `cf_err(msg, **fields)`?

manishtomar:
Yes. `cf_fail` sounds nice. Will change it.

radix:
LGTM :)

PR #1449, opened 2015-05-14
removed unnecessary wait_for_exp..building_servers

manishtomar:
since the next line `assert_servers_deleted` is checking for number of servers in the account anyways. Also, above wait_* function was not checking for servers for a long time causing tests to fail in prod sometimes since it takes time to delete servers sometimes.


cyli:
`check_for_expected_number_of_building_servers` also checks the desired capacity, as well as the servers in Nova.  `assert_servers_deleted_successfully` only checks Nova.  Should we also make sure that the number in the scaling group matches the number in Nova?  This would ensure, for instance, that convergence gathers one last time so that the number of servers in Nova is synced with the number of servers autoscale thinks the group has.

manishtomar:
@cyli I've added check for group state also. Kindly check :).

cyli:
LGTM!  Thanks for fixing!

PR #1450, opened 2015-05-14
cloudfeeds observer chain

manishtomar:
cloudfeeds observer is added with spec, pep3103 and error formatter in it. One extra refactor done is replacing 3 and 6 in formatters with `LogLevel.INFO and ERROR`.

### DIFF
--- a/otter/log/cloudfeeds.py
+++ b/otter/log/cloudfeeds.py
@@ -12,13 +12,17 @@ from effect import Effect, Func
 
 from toolz.dicttoolz import keyfilter
 
+from twisted.python.log import addObserver
+
 from txeffect import perform
 
 from otter.cloud_client import TenantScope, service_request
 from otter.constants import ServiceType
 from otter.effect_dispatcher import get_full_dispatcher
 from otter.log import log as otter_log
-from otter.log.formatters import PEP3101FormattingWrapper
+from otter.log.formatters import (
+    ErrorFormattingWrapper, LogLevel, PEP3101FormattingWrapper,
+    SpecificationObserverWrapper)
 from otter.util.http import append_segments
 from otter.util.pure_http import has_code
 from otter.util.retry import (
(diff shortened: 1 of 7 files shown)


radix:
```
manisht: have you considered actually putting it in our normal observer chain instead of keeping it separate?
manisht: like maybe we could just have a SplitObserver that splits it out into multiple child observers, and insert that into the middle of our normal observer chain
so we don’t have to duplicate the beginning parts
I dunno, it probably doesn’t matter, just an idea
```

manishtomar:
> have you considered actually putting it in our normal observer chain instead of keeping it separate?

I would like to put this as part of normal observer chain but its difficult to do that since this observer is added at runtime based on configuration in `tap/api.py`. I _could_ always add the obsever without any config in beginning and put the config when loading in api.py. This would require reference to the observer which may need to be global. That is something we can avoid when adding observer later on. 

> like maybe we could just have a SplitObserver that splits it out into multiple child observers, and insert that into the middle of our normal observer chain

That actually sounds like a lovely idea. In fact this should be part of Twisted logging system. Maybe it is. I haven't seen the new logging system :). However, like I mentioned earlier there is config problem.

radix:
yeah, the whole twisted logging system is a pain when it comes to initialization and configuration. In other projects, I've avoided using the twistd command line options and done everything in my deployment code, by basically hacking into the observers list (which is technically public). Anyway, no worries, we can continue to do it this way. It may be nice to at least try to share the code that sets up the (partial) observer chain so we don't have to remember to change it in two places if we add an transformer to the "front".

otter/log/cloudfeeds.py
```diff
@@ -188,13 +189,24 @@ def __call__(self, event_dict):
...
+
+def add_cf_observer(reactor, authenticator, tenant_id, region,
+                    service_configs):
+    """
+    Add cloud feeds observer after setting up some intial formatting
+    """
+    cf_observer = CloudFeedsObserver(
+        reactor=reactor, authenticator=authenticator, tenant_id=tenant_id,
+        region=region, service_configs=service_configs)
+    addObserver(
```
radix: I would *slightly* prefer not calling addObserver here, but rather returning an observer. Then the deployment code could call addObserver. No big deal though.

radix:
LGTM, just have an optional comment.

otter/log/cloudfeeds.py
manishtomar: Aah. That does sound better. Thanks :) I did this to create the observer chain in this module. But this is much better. This only returns the observer. Will change it.

radix:
Looks nice! I wish our observers had equality so we didn't need to mock so much, but this is good :)

PR #1451, opened 2015-05-15
scheduler stop error fix

manishtomar:
`POST /scheduler/stop` returns string instead of None. This was causing weird traceback inside the reactor when stopping scheduler during deploys.


cyli:
LGTM!  Thanks for noticing and fixing!

PR #1453, opened 2015-05-18
Configurable build timeout

### LINKED ISSUE #1368: server building timeout as config option
For purposes of testing (#1319), otter should provide server building timeout as config option. Testing aside, its still a good idea for this kind of info as configuration.

cyli:
Fixes #1368 

I made this a required argument everywhere mainly because I found it annoying to propagate the exact same default everywhere.  On the other hand, I had to change all the tests.  On the third hand, doing so tests the propagation of the value.

I'm fine with changing it should anyone have strong feelings.

I left it optional in `planning.converge` because that's already well tested, and has a LOT of tests which call `converge` without the timeout.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -62,12 +62,13 @@
     "cloudfeeds": {"service": "cloudFeeds", "tenant_id": "tenant",
                    "url": "http://cfurl.net/"},
     "converger": {
+        "build_timeout": 3600,
         "interval": 30
     },
     "cloud_client": {
-	"throttling": {
-	    "create_server_delay": 1,
-	    "delete_server_delay": 0.2
-	}
+    	"throttling": {
+    	    "create_server_delay": 1,
+    	    "delete_server_delay": 0.2
+    	}
     }
 }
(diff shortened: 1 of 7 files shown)


radix:
this looks good, but I think Nova people have encouraged us to increase the default timeout to something much larger, like 3 or 4 hours, so maybe we should change that?  (though I guess this default doesn't actually matter if we put a value in the production config file that's different).... so... LGTM :)

And thanks for fixing up all those docstrings :)

cyli:

cyli:
@radix Thanks!  I'll make that change to the production config :)

cyli:
@radix: actually, this only changes the convergence build timeout - think that's ok?

radix:
I guess we don't have a configurable timeout for worker builds, do we? I wouldn't worry about it; this PR is fine by itself.

cyli:
@radix ok, thanks

manishtomar:
@cyli Required argument is much better than anticipating defaults! In fact I feel lot of config option shouldn't have defaults and just fail if it is not found :)

cyli:
@manishtomar That's a good argument, yes!  I think we end up having all these defaults because we add in config options later, and the tests end up having huge changes if we require the arg.

PR #1454, opened 2015-05-20
we don't have to worry about this for next 35 years ;)

manishtomar:


cyli:
Yay!  :D. Thanks for fixing!!  LGTM once Travis +1's

PR #1457, opened 2015-05-21
Skip mimic-required tests if mimic is not configured

cyli:
Wasn't sure whether to do this or to check the service catalog.  Checking the service catalog would mean that the decorator would have to reach inside the decorated function, extract self, and get check the service catalog and raise `SkipTest`.

Since we rarely need to determine what kind of system we're testing against on the fly, I figure it can just be part of the config (similarly with RCv3 when we get to doing those tests).

I've tested this locally both with and without setting the environment variables - on Jenkins CI testing I've set the environment variable already, so jenkins would be testing this not-skipping.


cyli:

manishtomar:
LGTM :)

PR #1458, opened 2015-05-23
Servers cache using changes-since filter

manishtomar:
Implements first 3 tasks of #881. Reviewer question:

The cache model interface assumes multiple caches of a scaling group separated by `update_time`. `get_servers` is implemented with that assumption whereas in reality only one cache is always used. Should the model and/or the impl change with only one cache in mind? If so, I'll remove `take_while` and its usage. This is why I don't have its tests yet.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -51,15 +51,16 @@ def get_all_server_details(changes_since=None, batch_size=100):
     url = append_segments('servers', 'detail')
     query = {'limit': batch_size}
     if changes_since is not None:
-        query['changes_since'] = '{0}Z'.format(changes_since.isoformat())
+        query['changes-since'] = '{0}Z'.format(changes_since.isoformat())
 
     last_link = []
 
     def get_server_details(query_params):
+        params = sorted(query_params.items())
         eff = retry_effect(
             service_request(ServiceType.CLOUD_SERVERS, 'GET',
                             "{}?{}".format(url,
-                                           urlencode(query_params, True))),
+                                           urlencode(params, True))),
             retry_times(5), exponential_backoff_interval(2))
         return eff.on(continue_)
 
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -141,6 +141,13 @@ def server_of_group(group_id, server):
     return group_id_from_metadata(server.get('metadata', {})) == group_id
 
 
+def group_servers(group_id, grouped_servers):
+    """
+    Return servers of a group defaulting to empty if not found
+    """
+    return grouped_servers.get(group_id, [])
+
+
 @do
 def get_scaling_group_servers(tenant_id, group_id, now,
                               cache_class=CassScalingGroupServersCache,
@@ -156,6 +163,7 @@ def get_scaling_group_servers(tenant_id, group_id, now,
     :return: Servers as list of dicts
     :rtype: Effect
     """
+
     cache = cache_class(tenant_id, group_id)
     cached_servers, last_update = yield cache.get_servers()
     if last_update is None:
@@ -163,8 +171,9 @@ def get_scaling_group_servers(tenant_id, group_id, now,
         yield cache.insert_servers(now, servers, True)
     elif now - last_update >= timedelta(days=30):
         last_update = now - timedelta(days=30)
-        changes = (yield all_as_servers(last_update)).get(group_id, [])
-        current = (yield all_as_servers()).get(group_id, [])
+        changes, current = yield parallel([
+            all_as_servers(last_update).on(partial(group_servers, group_id)),
+            all_as_servers().on(partial(group_servers, group_id))])
         servers = merge_servers(changes, current)
         yield cache.insert_servers(now, servers, True)
     else:
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -1,12 +1,13 @@
 """Code related to gathering data to inform convergence."""
+from datetime import datetime
 from functools import partial
 from urllib import urlencode
 from urlparse import parse_qs, urlparse
 
-from effect import catch, parallel
+from effect import Func, catch, do, do_return, parallel
 
 from toolz.curried import filter, groupby, keyfilter, map
-from toolz.dicttoolz import get_in
+from toolz.dicttoolz import get_in, merge
 from toolz.functoolz import compose, identity
 from toolz.itertoolz import concat
 
@@ -23,6 +24,7 @@ from otter.convergence.model import (
     RCv3Node,
     group_id_from_metadata)
 from otter.indexer import atom
+from otter.models.cass import CassScalingGroupServersCache
 from otter.util.http import append_segments
 from otter.util.retry import (
     exponential_backoff_interval, retry_effect, retry_times)
@@ -89,7 +91,8 @@ def _discard_response((response, body)):
     return body
 
 
-def get_scaling_group_servers(changes_since=None, server_predicate=identity):
+def get_all_scaling_group_servers(changes_since=None,
+                                  server_predicate=identity):
     """
     Return tenant's servers that belong to a scaling group as
     {group_id: [server1, server2]} ``dict``. No specific ordering is guaranteed
@@ -111,8 +114,43 @@ def get_scaling_group_servers(changes_since=None, server_predicate=identity):
                             filter(server_predicate),
                             filter(has_group_id))
 
-    eff = get_all_server_details(changes_since)
-    return eff.on(servers_apply)
+    return get_all_server_details(changes_since).on(servers_apply)
+
+
+def all_servers(cache, changes):
+    """
+    Return all servers got by merging cache and changes since the cache was
+    last fetch
+    """
+    def sdict(servers):
+        return {s['id']: s for s in servers}
+
+    return merge(sdict(cache), sdict(changes)).itervalues()
+
+
+@do
+def get_scaling_group_servers(tenant_id, group_id, cache_coll=None):
+    """
+    Get a group's servers taken from cache if it exists. Updates cache as
+    part of getting the group
+
+    :return: Servers as iterator of dicts
+    :rtype: ``iterator``
+    """
+    coll = cache_coll or CassScalingGroupServersCache()
+    cache, last_update = yield coll.get_servers(tenant_id, group_id)
+    all_changes = yield get_all_scaling_group_servers(last_update)
+    changes = all_changes[group_id]
+    if last_update is None:
+        # cache is empty. create it from current servers
+        now = yield Func(datetime.now)
+        # TODO: Should the time be "now - (say) 5 mins" to avoid possible
+        # clock drift between this machine and nova servers
+        yield coll.insert_servers(tenant_id, group_id, changes, now)
+        servers = changes
+    else:
+        servers = all_servers(cache, changes)
+    yield do_return(servers)
 
 
 def get_clb_contents():
@@ -235,6 +273,7 @@ def get_rcv3_contents():
 
 
 def get_all_convergence_data(
+        tenant_id,
         group_id,
         get_scaling_group_servers=get_scaling_group_servers,
         get_clb_contents=get_clb_contents,
@@ -246,8 +285,7 @@ def get_all_convergence_data(
     Returns an Effect of ([NovaServer], [LBNode]).
     """
     eff = parallel(
-        [get_scaling_group_servers()
-         .on(lambda servers: servers.get(group_id, []))
+        [get_scaling_group_servers(tenant_id, group_id)
          .on(map(NovaServer.from_server_details_json)).on(list),
          get_clb_contents(),
          get_rcv3_contents()]
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -135,7 +135,8 @@ def merge_servers(first, second):
 
 @do
 def get_scaling_group_servers(tenant_id, group_id, now,
-                              cache_class=CassScalingGroupServersCache):
+                              cache_class=CassScalingGroupServersCache,
+                              all_as_servers=get_all_scaling_group_servers):
     """
     Get a group's servers taken from cache if it exists. Updates cache
     if it is empty from newly fetched servers
@@ -146,17 +147,17 @@ def get_scaling_group_servers(tenant_id, group_id, now,
     :return: Servers as iterator of dicts
     :rtype: Effect
     """
-    coll = cache_class(tenant_id, group_id)
-    cache, last_update = yield coll.get_servers()
-    if last_update is None or last_update - now >= timedelta(days=30):
+    cache = cache_class(tenant_id, group_id)
+    cached_servers, last_update = yield cache.get_servers()
+    if last_update is None or now - last_update >= timedelta(days=30):
         last_update = now - timedelta(days=30)
-        changes = (yield get_all_scaling_group_servers(last_update))[group_id]
-        current = (yield get_all_scaling_group_servers())[group_id]
+        changes = (yield all_as_servers(last_update))[group_id]
+        current = (yield all_as_servers())[group_id]
         servers = merge_servers(changes, current)
-        yield coll.insert_servers(now, servers, True)
+        yield cache.insert_servers(now, servers, True)
     else:
-        changes = (yield get_all_scaling_group_servers(last_update))[group_id]
-        servers = merge_servers(cache, changes)
+        changes = (yield all_as_servers(last_update))[group_id]
+        servers = merge_servers(cached_servers, changes)
     yield do_return(servers)
 
 
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -129,7 +129,8 @@ def all_servers(cache, changes):
 
 
 @do
-def get_scaling_group_servers(tenant_id, group_id, cache_coll=None):
+def get_scaling_group_servers(tenant_id, group_id,
+                              cache_class=CassScalingGroupServersCache):
     """
     Get a group's servers taken from cache if it exists. Updates cache as
     if it is empty from newly fetched servers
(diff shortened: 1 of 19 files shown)


otter/util/fp.py
```diff
@@ -86,3 +86,16 @@ def assoc_obj(o, **k):
     new_o = deepcopy(o)
     new_o.__dict__.update(k)
     return new_o
+
+
+def take_while(pred, items):
+    """
+    Return list until pred is true for item in items. Similar to `takeWhile`
```
cyli: Non-blocking - the `takeWhile` docstring seems pretty great :) Might be worth copying here - something like "Returns the longest prefix (possibly empty) of `items` of elements that satisfy `pred`"?

schema/setup/control_50_servers_cache.cql
```diff
@@ -0,0 +1,14 @@
+USE @@KEYSPACE@@;
+
+CREATE TABLE servers_cache (
+    tenant_id ascii,
+    group_id ascii,
+    last_fetch timestamp,
+    server_id ascii,
+    server_blob ascii,
+    PRIMARY KEY((tenant_id, group_id), last_fetch, server_id)
+) WITH CLUSTERING ORDER BY (last_fetch DESC, server_id ASC) AND
```
cyli: Non-blocking - do we need `server_id` to be sorted?  (Also, does cass do sort it in ascending order?  I tried specifying `last_fetch` only in the clustering order - cass seems to return data with `server_id` sorted in ascending order, but I don't know if it's stored that way or if specifying the `server_id` does something special.  Sorry, I've forgotten since the last time we discussed sort order.  :(  ).

otter/convergence/gathering.py
```diff
@@ -111,8 +114,44 @@ def group_id(s):
...
+
+def all_servers(cache, changes):
+    """
+    Return all servers got by merging cache and changes since the cache was
+    last fetch
+    """
+    def sdict(servers):
+        return {s['id']: s for s in servers}
+
+    return merge(sdict(cache), sdict(changes)).itervalues()
```
cyli: Do we have to do something more than just a dictionary merge?  Otherwise, won't the cache of deleted servers will grow indefinitely?  E.g. - if there is a deleted server in the cache that is no longer in the list of changes, should we then remove it from the next cache, which counts it as cleaned up?

otter/convergence/service.py
```diff
@@ -155,6 +156,9 @@ def execute_convergence(tenant_id, group_id, build_timeout,
                                            status=ScalingGroupStatus.ACTIVE))
             yield cf_msg('group-status-active',
                          status=ScalingGroupStatus.ACTIVE.name)
+        else:
+            # Clear servers cache that removes deleted servers
+            yield cache_class(tenant_id, group_id).delete_servers()
```
cyli: Apologies - I thought `delete_servers` deleted all servers from the cache?  If not, then I guess please ignore my comment above about deleted servers.  (And also, if that's the case, should we also have a delete all servers function that gets called when the group gets deleted for real?)

cyli:
Looks pretty nice so far.  :D  Two additional comments:

1.  Do we need an intent for cleaning up old server caches?  Getting all the servers is going to get slow pretty quickly if we don't clean up regularly.
1.  I don't see the code for writing the new updated cache to the DB yet - just inserting a cache if the cache is empty.

All of my comments can also be handled in other PRs.

schema/setup/control_50_servers_cache.cql
manishtomar: Not really, I could remove `server_id ASC` and CASS will still keep it sorted. I just added it for completeness. The important thing is `last_fetch DESC` since that is how the data is going to be stored and how the query is going to be made.

otter/models/cass.py
```diff
@@ -1682,6 +1693,50 @@ def health_check(self):
...
+    @do
+    def get_servers(self):
+        query = ("SELECT server_blob, last_update FROM {cf} "
+                 "WHERE tenant_id=:tenant_id AND group_id=:group_id "
+                 "ORDER BY last_update DESC;")
+        rows = yield cql_eff(query.format(cf=self.table), self.params)
+        if len(rows) == 0:
+            yield do_return(([], None))
+        last_update = rows[0]['last_update']
+        rows = take_while(lambda r: r['last_update'] == last_update, rows)
```
cyli: Yay!  Awesome use for `take_while`!!

cyli:
After discussion with @manishtomar IRL:  the logic here was just to read the servers into the cache once, at the beginning of convergence.  Then changes since are read over and over during convergence, from the time of that first cache, and the merged data will be kept in memory and not written to the DB.

After convergence is over, the entire cache will be deleted, and the next time convergence is triggered, it will read all of nova again.

Unfortunately that means that if a server was deleted out of band between two convergences, otter will not notice, since we get only the current server details if there is no cache.  So either we have to get both the current server details and all changes since for the last 30 days, or just keep the cache around between convergences.

We are probably just going to keep the cache around between convergences.

1.  When convergence starts up, and there is no cache, it will get the current state of the world as it does in this PR, and write it to the cache.
1.  As is currently in this PR, convergence will take that cache, merge it with updates, and use that merged state during convergence.  But it won't write to the cache, as is true in the current PR.
1.  (difference from current PR) Once the convergence cycle is complete and successful, it will write a the merged state to the cache (similarly to how it writes active servers?)  If the convergence cycle is successful, there are no deleted servers anymore to clean up.  If convergence fails  (goes into error), then eh, the cache doesn't matter if it's not the most up-to-date- it will just help bootstrap the process next time.
1.  (difference from current PR) If the cache is more than 30 days old, we're going to have to get both the current server details and all changes for 30 days, in order to cover the following case:  Convergence finished, cache was written.  Server with CLB is deleted out of band immediately.  25 days later, another server with CLB is deleted out of band.  40 days later, another convergence cycle was triggered.  (We need the state of the world in order to get an accurate count of current servers, because `changes-since` is limited by 30 days.  We need `changes-since` for 30 days because otherwise we can't clean up that second OOB-deleted server).

otter/convergence/gathering.py
```diff
@@ -111,8 +114,49 @@ def group_id(s):
...
+
+    :return: Servers as iterator of dicts
+    :rtype: Effect
+    """
+    coll = cache_class(tenant_id, group_id)
+    cache, last_update = yield coll.get_servers(tenant_id, group_id)
+    if last_update is None or last_update - now >= timedelta(days=30):
+        last_update = now - timedelta(days=30)
+        changes = (yield get_all_scaling_group_servers(last_update))[group_id]
+        current = (yield get_all_server_details())[group_id]
```
cyli: Should this be `(yield get_all_scaling_group_servers())[group_id]`?  `get_all_server_details` doesn't group them by group ID I think.

Non-blocking idle thought - I'm also wondering if we should rename these functions?   (the `get_all_scaling_group_servers` change was good, I'm just wondering if we should name the ones that gather from nova explicitly as gathering from nova, as opposed to this function, which does both cache checking and gathering.

manishtomar: > Should this be (yield get_all_scaling_group_servers())[group_id]? get_all_server_details doesn't group them by group ID I think.

Yes. Thanks :)

> I'm just wondering if we should name the ones that gather from nova explicitly as gathering from nova, as opposed to this function, which does both cache checking and gathering.

From what I understand the current code does that: `get_all_server_details` is the only Nova gathering function. Other functions use that. Can describe in detail what you had in mind?

cyli: Sure, like `get_all_server_details` -> `get_all_tenants_servers_from_nova`, `get_all-scaling_group_servers` -> `get_all_tenants_autoscaled_servers_from_nova` or something.

manishtomar: Hmm.. How about
```python
def get_tenant_servers(changes_since):
   """ Get all servers in the tenant from Nova """

def get_tenant_autoscaled_servers(changes_since):
   """ Get tenant servers belonging to any autoscaling group from Nova """

def get_tenant_autoscaled_group_servers(tenant_id, group_id, now):
   """ Get tenant servers belonging to a particular scaling group optionally using the cache """
```

cyli: Hmm.. what ahout just adding a `_` in front of the other two functions?  Mainly I just wanted to differentiate those two which are only used by the third one from the third one, which is the actual one used by gathering (hence adding "nova" to the other two).

This is just nitpicking though, and completely non-blocking.  We can also not rename it - you're right, it's not too bad currently.  :)

otter/test/convergence/test_gathering.py
```diff
@@ -212,12 +221,84 @@ def test_retry(self):
                                 next_interval=exponential_backoff_interval(2)))
 
 
-class GetScalingGroupServersTests(SynchronousTestCase):
+class GetScalingGroupServerTests(SynchronousTestCase):
```
cyli: Non-blocking nitpick:  Should this still be "Servers" (plural), because the function name is still `get_scaling_group_servers`?

manishtomar: Yes. Thank you :)

otter/models/cass.py
```diff
@@ -1682,6 +1694,54 @@ def health_check(self):
...
+
+    def __init__(self, tenant_id, group_id):
+        self.tenant_id = tenant_id
+        self.group_id = group_id
+        self.table = "servers_cache"
+        self.params = {"tenant_id": tenant_id, "group_id": group_id}
+
+    @do
+    def get_servers(self):
+        query = ("SELECT server_blob, last_update FROM {cf} "
```
cyli: Oops, sorry I didn't catch this before.  :(  I think in the schema this is `last_fetch`, not `last_update`

otter/convergence/gathering.py
```diff
@@ -111,8 +115,50 @@ def group_id(s):
...
+    # they also not take tenant_id and work on the scope?
+
+    :return: Servers as iterator of dicts
+    :rtype: Effect
+    """
+    cache = cache_class(tenant_id, group_id)
+    cached_servers, last_update = yield cache.get_servers()
+    if last_update is None or now - last_update >= timedelta(days=30):
+        last_update = now - timedelta(days=30)
+        changes = (yield all_as_servers(last_update))[group_id]
```
cyli: This probably  needs to be `.get(group_id, {})`, for when there are no servers yet.

manishtomar: Ooh. Great catch! Thank you :)

otter/models/cass.py
```diff
@@ -1682,6 +1694,54 @@ def health_check(self):
...
+            yield do_return(([], None))
+        last_update = rows[0]['last_update']
+        rows = take_while(lambda r: r['last_update'] == last_update, rows)
+        yield do_return(([json.loads(r['server_blob']) for r in rows],
+                         last_update))
+
+    def insert_servers(self, last_update, servers, clear_others):
+        query = ("INSERT INTO {cf} (tenant_id, group_id, last_update, "
+                 "server_id, server_blob "
+                 "VALUES(:tenant_id, :group_id, :last_update, :server_id{i}, "
```
cyli: Do these `i` values need to be enumerated? Like there should be `:server_id1, :server_id2, ..., :server_blob1, :server_blob2,...` ?

cyli: Wait, no I misunderstood, nm.

cyli:
Also, I think the CQL performer needs to be added to the full dispatchers - nothing uses it right now except the `load_cql` script - the other DB interactions right now just use the model performers and intents.

otter/models/cass.py
cyli: I think the error I was getting was that there were no queries (because there are no servers), but the empty query was still being executed with params.  You asked if there was a case where there wouldn't be a cache - and I totally missed this obvious one.  :(  Apologies.  Which means that we get the last 30 days always if there are no servers on the group.  :|  Do we need to provide a sentinel value to stop it?

otter/models/cass.py
```diff
@@ -1682,6 +1694,54 @@ def health_check(self):
...
+        if len(rows) == 0:
+            yield do_return(([], None))
+        last_update = rows[0]['last_update']
+        rows = take_while(lambda r: r['last_update'] == last_update, rows)
+        yield do_return(([json.loads(r['server_blob']) for r in rows],
+                         last_update))
+
+    def insert_servers(self, last_update, servers, clear_others):
+        query = ("INSERT INTO {cf} (tenant_id, group_id, last_update, "
+                 "server_id, server_blob "
```
cyli: Also I think this might be missing a closing parenthesis?

schema/setup/control_50_servers_cache.cql
```diff
@@ -0,0 +1,14 @@
+USE @@KEYSPACE@@;
+
+CREATE TABLE servers_cache (
+    tenant_id ascii,
+    group_id ascii,
+    last_update timestamp,
+    server_id ascii,
+    server_blob ascii,
+    PRIMARY KEY((tenant_id, group_id), last_fetch, server_id)
```
cyli: `last_update` here and in the clustering order too :)

manishtomar: yes. Sorry again :(

cyli: No sorries!  (づ｡◕‿‿◕｡)づ

manishtomar:

manishtomar:

cyli:

manishtomar:
The metadata removal tests fail because next gathering of servers still contain the metadata removed server. This happens because next [fetch](https://github.com/rackerlabs/otter/blob/change-since/otter/convergence/gathering.py#L163) will not contain the server but it will be [added](https://github.com/rackerlabs/otter/blob/change-since/otter/convergence/gathering.py#L164) from the cache. 

I'll be changing the code to get _all_ servers, merged them and then filter out autoscaled servers.

manishtomar:

manishtomar:
There is something with this branch causing otter to consistently seg fault when running all the trial tests in parallel. This doesn't happen with master. I am trying to debug.

manishtomar:

otter/convergence/service.py
```diff
@@ -111,7 +113,8 @@ def execute_convergence(tenant_id, group_id, build_timeout,
     # fetching of the scaling group info from cassandra.
     sg_eff = Effect(GetScalingGroupInfo(tenant_id=tenant_id,
                                         group_id=group_id))
-    gather_eff = get_all_convergence_data(group_id)
+    now_dt = yield Effect(Func(datetime.now))
```
radix: This should be `datetime.utcnow()`

manishtomar: Ah yes. Thank you for noticing this.

otter/convergence/gathering.py
```diff
@@ -111,8 +116,62 @@ def group_id(s):
...
+    :rtype: Effect
+    """
+    cache = cache_class(tenant_id, group_id)
+    cached_servers, last_update = yield cache.get_servers()
+    if last_update is None:
+        servers = (yield all_as_servers()).get(group_id, [])
+        yield cache.insert_servers(now, servers, True)
+    elif now - last_update >= timedelta(days=30):
+        last_update = now - timedelta(days=30)
+        changes = (yield all_as_servers(last_update)).get(group_id, [])
```
cyli: Non-blocking - wondering if this can be done in parallel?  Doesn't need to happen, as this would involve a change in the tests, and it works fine in serial.

manishtomar: Aah yes. Did not think about it. Maybe add a tech-debt ticket or a note here?

cyli: Hmm.... up to you.  :)  I dunno that I'd consider it even important enough to add tech-debt.  Sorry, was just an idle thought :D

otter/test/convergence/test_gathering.py
```diff
@@ -111,7 +122,8 @@ def test_get_all_without_link_to_next_page(self):
         body = {'servers': self.servers}
         eff = get_all_server_details(batch_size=10)
         svcreq = resolve_retry_stubs(eff)
-        result = resolve_svcreq(svcreq, (fake_response, body), *self.req())
```
cyli: Non-blocking - I think `self.req` is no longer called anywhere - can it be removed?

manishtomar: Oh yeah. Sure. Don't know why I didn't remove it. Thanks! :) Will do it.

otter/test/utils.py
```diff
@@ -845,3 +846,39 @@ def as_effect(self):
...
+
+class Cache(object):
+    """ IScalingGroupServersCache impl for testing """
+
+    def __init__(self, tid, gid):
+        self.tid = tid
+        self.gid = gid
+
+    def ids(self, s):
+        return "cache" + s + self.tid + self.gid
```
cyli: Non-blocking - might be easier to read and understand what this is in the tests if this were something like `"cache({},{}).{}".format(self.tid, self.gid, s)`, and `s` were one of (get, insert, delet).

So that the tests would use something like `"cache(tenant,group).get"` in the sequence dispatchers.  This would make the string longer though, and would result in more wrapping.

We can file a tech debt ticket, or just ignore it.  :)

otter/test/convergence/test_gathering.py
```diff
@@ -218,9 +235,101 @@ class GetScalingGroupServersTests(SynchronousTestCase):
...
+
+    def test_from_cache(self):
+        """
+        If cache is < 30 days old then servers returned are merge of
+        changes since the cache time
+        """
+        asmetakey = "rax:autoscale:group:id"
+        cache = [
+            {'id': 'a', 'metadata': {asmetakey: "gid"}},
+            {'id': 'b', 'metadata': {asmetakey: "gid"}}]
```
cyli: To test that cached servers get removed if in the changes, the metadata has been changed, could we also add a server with id `d` to the cache?

manishtomar: Good point. Done :)

manishtomar:
Closing as #1549 completes it

PR #1459, opened 2015-05-26
Initial CATC-19

### LINKED ISSUE #1327: CATC19: Scale group with CLB to over 25
Scale group with CLB to over 25
	- validate that the group enters a TBD error state

Prereq: There must be a defined Backoff/Impossible group state for autoscale. (This does not necessarily need to be implemented, but there must be a requirements definition.)

kivattik:
Scale to over CLB limit and expect group to enter error state

Fixes #1327

Waiting for dependent mimic issue https://github.com/rackerlabs/mimic/issues/291

### DIFF
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -1080,6 +1080,7 @@ class ConvergenceTestsWith1CLB(unittest.TestCase):
                 "See #881.")
         return (name, wrapper)
 
+    @skip_me("Mimic does not support CLB limits, skipped pending Mimic #291")
     @tag("CATC-019")
     def test_scale_over_lb_limit(self):
         """
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -351,6 +351,50 @@ class TestConvergence(unittest.TestCase):
             )
         )
 
+    @tag("CATC-019")
+    def test_scale_over_lb_limit(self):
+        """
+        CATC-019-a: Validate that when a group attempts to scale over the
+        load balancer limit the group enters an error state. It is expected
+        that the active capacity will match the maximum allowed on the load
+        balancer and the pending capacity will become zero once the group is
+        in ERROR state.
+
+        1. Creates a scaling group with a load balancer and max servers
+           greater than the load balancer max. (Assume LB_max == 25)
+        2. Scale up to a desired capacity greater than the LB_max
+        3. Assert that the scaling group goes into error state, that the
+           active count is equal to LB_max, and that no servers are pending.
+
+        """
+        LB_max = 25
+        group_max = 30
+        self.scaling_group = self.helper.create_group(
+            image_ref=image_ref, flavor_ref=flavor_ref, min_entities=1,
+            max_entities=group_max)
+
+        self.scale_up_to_max = ScalingPolicy(
+            scale_by=group_max,
+            scaling_group=self.scaling_group
+        )
+
+        return (
+            self.helper.start_group_and_wait(self.scaling_group, self.rcs)
+            .addCallback(self.scale_up_to_max.start, self)
+            .addCallback(self.scale_up_to_max.execute)
+            .addCallback(
+                self.scaling_group.wait_for_state,
+                MatchesAll(
+                    ContainsDict({
+                        'pendingCapacity': Equals(0),
+                        'desiredCapacity': Equals(group_max),
+                        'status': Equals("ERROR")
+                    }),
+                    HasActive(LB_max)),
+                timeout=1600
+            )
+        )
+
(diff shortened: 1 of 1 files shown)


otter/integration/tests/test_convergence.py
```diff
@@ -351,6 +351,74 @@ def test_scale_over_group_max_after_metadata_removal(self):
...
+           active count is equal to LB_max, and that no servers are pending.
+
+        """
+        LB_max = 25
+        group_max = 30
+
+        rcs = TestResources()
+        self.helper = TestHelper(self, num_clbs=1)
+
+        # def create_clb_first():
```
sam-falvo: Remove dead code.

otter/integration/tests/test_convergence.py
```diff
@@ -351,6 +351,74 @@ def test_scale_over_group_max_after_metadata_removal(self):
...
+                        'pendingCapacity': Equals(0),
+                        'desiredCapacity': Equals(LB_max),
+                        'status': Equals("ACTIVE")
+                    }),
+                    HasActive(LB_max)),
+                timeout=600
+            )
+        )
+
+        # return create_clb_first().addCallback(then_test)
```
sam-falvo: Remove dead code.

otter/integration/tests/test_convergence.py
```diff
@@ -351,6 +351,74 @@ def test_scale_over_group_max_after_metadata_removal(self):
...
+                                             desired=LB_max - 2)
+            .addCallback(self.scale_up_to_group_max.start, self)
+            .addCallback(self.scale_up_to_group_max.execute)
+            .addCallback(
+                self.scaling_group.wait_for_state,
+                MatchesAll(
+                    ContainsDict({
+                        'pendingCapacity': Equals(0),
+                        'desiredCapacity': Equals(LB_max),
+                        'status': Equals("ACTIVE")
```
sam-falvo: Shouldn't this be `ERROR` instead of `ACTIVE`?

otter/integration/tests/test_convergence.py
```diff
@@ -351,6 +351,74 @@ def test_scale_over_group_max_after_metadata_removal(self):
...
+            max_entities=group_max)
+
+        self.scale_up_to_group_max = ScalingPolicy(
+            set_to=LB_max,
+            scaling_group=self.scaling_group
+        )
+
+        return (
+            self.helper.start_group_and_wait(self.scaling_group, rcs,
+                                             desired=LB_max - 2)
```
sam-falvo: Shouldn't this be `LB_max + 2` instead of `LB_max - 2`?

otter/integration/tests/test_convergence.py
```diff
@@ -351,6 +351,74 @@ def test_scale_over_group_max_after_metadata_removal(self):
...
+        #         .addCallback(self.clb.wait_for_state, "ACTIVE", 600)
+        #     )
+
+        # def then_test(_):
+        self.scaling_group = self.helper.create_group(
+            image_ref=image_ref, flavor_ref=flavor_ref, min_entities=1,
+            max_entities=group_max)
+
+        self.scale_up_to_group_max = ScalingPolicy(
+            set_to=LB_max,
```
sam-falvo: Shouldn't this be set to something like `LB_max + 1` if we're to test behavior of it going over CLB maximums?

sam-falvo:
LGTM pending removal of dead code.  I left comments on some sections of code which I think the previous commit had, but you changed in the 2nd commit.  I'm guessing you left it as-is after wanting to just check that your code worked as expected?

cyli:
Probably this test should be moved to `ConvergenceTestsWith1CLB`, which waits for the CLB to be active in the `setUp`.  The `setUp` also creates the 1 clb, and sets up test resources, so the test doesn't have to do it itself.  :)  Once I did that, the test seems to pass.

otter/integration/tests/test_convergence.py
cyli: I thinks this was active currently to make the test pass, because mimic does not yet limit the number of nodes that can go on the CLB (and hence mimic will not actually return an error response if you add more than 25 nodes to the CLB yet).

We can make it "ERROR" and then skip the test for now, pending the feature in mimic though.  :)

sam-falvo: Ah, OK -- I wasn't aware.  Thanks!

kivattik:
I'll definitely move the test over. I was thinking of the 1 LB class as all constructed tests, and had forgotten I could include additional ones as well. That nicely solves being unable to create the LB with the  test helper that is created in the TestConvergence setup.

cyli:
There should probably be a `@skip_me` decorator that actually skips this test :)  Other than that, LGTM.

cyli:
LGTM - thanks @kivattik!

PR #1460, opened 2015-05-28
Add a random string to the error injection server name criteria in trial tests

cyli:
While attempting to debug intermittent timeouts, realized this could also cause intermittent timeouts.  Since the test is just copied over from the test without CLBs to the test with CLBs, both will have the same server name as the criteria.  So if that the same test both with and without CLB are running at the same time (which can happen with parallelization), they'll use the same behavior counter in mimic.

Which may cause some non-deterministic behavior.

I am also adding behavior deletion/cleanup to mimic, so these tests can clean up the behavior they're injecting too, but that will be another PR.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -98,7 +98,7 @@ def create_scaling_group_dict(
             }
         },
         "groupConfiguration": {
-            "name": "{}-{}".format(name, reactor.seconds()),
+            "name": name,
             "cooldown": 0,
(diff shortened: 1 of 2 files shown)


cyli:

cyli:

radix:
@cyli and I discussed a change that would put the random-naming into create_group (or another function) on slack.

cyli:

radix:
there we go. LGTM :)

Issue #1461, opened 2015-05-28
Log retry failures


cyli:
When retrying (in `otter.utils.retry`), transient errors should be optionally logged.  Maybe a callable should be accepted which logs the failures?

manishtomar:
This is duplicate of #1141. Should we close this?

cyli:
Oops, my bad!  I totally forgot about that one.  Yes, closing, sorry!

PR #1462, opened 2015-05-28
Log mismatches when waiting for a desired state in trial tests

cyli:
As in desired autoscale state, or desired nova server state, or desired CLB node state.

These will be logged in `_trial_temp/test.log` if running without concurrency, and in `_trial_temp/<#>/test.log` when running with concurrency.  (the # specifies which thread that test was running on).

This should probably be added to retry instead (see https://github.com/rackerlabs/otter/issues/1461), but this might be useful in the meantime.

Produces messages like:

```
2015-05-28 21:07:08+0000 [HTTP11ClientProtocol,client] Waiting for desired group state.
        Mismatch: Differences: [
        State {u'status': u'ACTIVE', u'desiredCapacity': 2, u'paused': False, u'active': [], u'pendingCapacity': 2, u'activeCapacity': 0, u'name': u'automatically-generated-test-configuration'} does not have 2 active servers.
        Differences: {
          'pendingCapacity': 0 != 2,
        }
        ]
```

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -14,9 +14,9 @@ from testtools.matchers import MatchesPredicateWithParams
 
 import treq
 
-from twisted.python.log import msg
 from twisted.internet import reactor
 from twisted.internet.defer import gatherResults, inlineCallbacks, returnValue
+from twisted.python.log import msg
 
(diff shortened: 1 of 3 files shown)


radix:
Also LGTM. :)

PR #1463, opened 2015-05-29
adding AutoScale DIY as appendix to dev guide

stanzikratel:
Added new appendix file that contains AutoScale DIY content
Added link to appendix to Dev guide

### DIFF
--- a/docbook/src/docbkx/autoscale-devguide.xml
+++ b/docbook/src/docbkx/autoscale-devguide.xml
@@ -1745,14 +1745,14 @@ config at https://github.com/rackerlabs/autoscaling-chef/blob/master/site-cookbo
                 </note>
             </para>
         </section>
-        <section xml:id="audit-log" security="writerOnly">
+        <?section xml:id="audit-log" security="writerOnly">
             <title>Audit logs</title>
             <para>Because of the deferred nature of Auto Scale, most
                 errors occur outside of the user's session. Errors are
                 logged to a user's cloud feed as an audit log. The
                 audit log contains all the events that have occurred
                 within a user's scaling group. </para>
-        </section>
+        </section>?>
         <section xml:id="faults-dle101">
             <title>Faults</title>
             <para>If any &PRODNAME; request results in an error, the
@@ -1817,4 +1817,5 @@ config at https://github.com/rackerlabs/autoscaling-chef/blob/master/site-cookbo
         <xi:include href="chapters/glossary.xml" xpointer="element(Glossary-000)"/>
         <!-- -->
     </chapter>
+    <xi:include href="chapters/appendix-diy.xml"/>
 </book>
--- a/docbook/src/docbkx/autoscale-devguide.xml
+++ b/docbook/src/docbkx/autoscale-devguide.xml
@@ -1745,14 +1745,6 @@ config at https://github.com/rackerlabs/autoscaling-chef/blob/master/site-cookbo
                 </note>
             </para>
         </section>
-        <?section xml:id="audit-log" security="writerOnly">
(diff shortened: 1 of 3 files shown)


stretchservers:

docbook/src/docbkx/autoscale-devguide.xml
```diff
@@ -1745,14 +1745,14 @@ config at https://github.com/rackerlabs/autoscaling-chef/blob/master/site-cookbo
...
-        <section xml:id="audit-log" security="writerOnly">
+        <?section xml:id="audit-log" security="writerOnly">
             <title>Audit logs</title>
             <para>Because of the deferred nature of Auto Scale, most
                 errors occur outside of the user's session. Errors are
                 logged to a user's cloud feed as an audit log. The
                 audit log contains all the events that have occurred
                 within a user's scaling group. </para>
-        </section>
+        </section>?>
```
radix: I'm not sure what this does, but I'm pretty sure it's invalid XML so I doubt it's being parsed the way you expect it. I may be wrong, though!

docbook/src/docbkx/chapters/appendix-diy.xml
```diff
@@ -0,0 +1,1309 @@
...
+                    <para>
+                        <itemizedlist>
+                            <listitem>
+                                <para>Use a text editor, such as Sublime, to form the commands. Copy
+                                    the command from this document into the editor and make the
+                                    required changes there. Then, paste the results into a shell
+                                    window and execute the command (press Enter). Note: Sublime
+                                    isn’t free, but there are several editors out there like
+                                    Notepad++, Vim, and Brackets, that are free. Trying to use the
+                                    commands, even modified, from the Word file directly may not
```
radix: This whole paragraph seems a bit awkward, especially the stuff about Sublime Text being non-free, but the reference to "Word" in particular is wrong, since this documentation is not in the form of a Word file. I would personally leave out the entire paragraph, but if you want to include something like this, here's what I would say:

"Copying and pasting commands directly from this documentation into your terminal program may be somewhat accident-prone, so we recommend you first copy them into your text editor of choice, modify them as required there, and then copy them into your terminal."

radix:
Thanks @stanzikratel ! I haven't had a chance to look through the whole document yet, but it's looking really good so far. I think a proper review would require actually running through the steps and trying to reproduce it, which will take some time. Hopefully one of us will get around to it some time soon!

docbook/src/docbkx/autoscale-devguide.xml
stanzikratel: @radix This was actually a section that the previous writer (Maria I assume) had commented out, so that it doesn't display in the build. I will go ahead and take the entire section out for now since it does not appear to be relevant.

docbook/src/docbkx/chapters/appendix-diy.xml
stanzikratel: @radix Great catch. Thanks for noticing the Word issue. The document was originally in Word format and I transferred it to Docbook/xml and somehow the mention of Word escaped my scrutiny. I went ahead and removed all the references to Sublime, etc and replaced them with the comment you provided.

manishtomar:
@stanzikratel Does this need to be merged since its old docbook style and we've moved to rst?

stanzikratel:
Hi Manish,

It would be good to merge this as we will still continue to have docbook internally for a little bit. Could you please merge it?

Thanks,

Constanze

manishtomar:
@stanzikratel OK cool. Merging :)

stanzikratel:
Great. Thanks! :-)

PR #1465, opened 2015-05-29
Clean up created error behaviors in mimic

cyli:
With the name randomization this should be less of an issue, but also good not to have a ton of error behaviors sitting around in mimic, since mimic iterates through all of them on every server create.

### DIFF
--- a/otter/integration/lib/mimic.py
+++ b/otter/integration/lib/mimic.py
@@ -95,13 +95,11 @@ class MimicNova(object):
 
         if self.test_case is not None:
             self.test_case.addCleanup(
-                self.delete_behavior, rcs, behavior_id, event_description,
-                [204])
+                self.delete_behavior, rcs, behavior_id, event_description)
 
         returnValue(behavior_id)
 
-    def delete_behavior(self, rcs, behavior_id, event_description="creation",
-                        success_codes=None):
+    def delete_behavior(self, rcs, behavior_id, event_description="creation"):
         """
         Given a behavior ID, delete it from mimic.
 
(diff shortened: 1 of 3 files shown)


cyli:

otter/integration/lib/mimic.py
```diff
@@ -71,4 +80,37 @@ def sequenced_behaviors(self, rcs, criteria, behaviors,
...
+
+        if add_cleanup_to is not None:
+            add_cleanup_to.addCleanup(
+                self.delete_behavior, rcs, behavior_id, event_description,
+                [204])
+
+        returnValue(behavior_id)
+
+    def delete_behavior(self, rcs, behavior_id, event_description="creation",
+                        success_codes=None):
```
manishtomar: How does it help to have `success_codes` as argument to this function?

otter/integration/lib/mimic.py
```diff
@@ -41,25 +43,32 @@ def change_server_statuses(self, rcs, ids_to_status):
             pool=self.pool
         ).addCallback(check_success, [201]).addCallback(self.treq.content)
 
+    @inlineCallbacks
     def sequenced_behaviors(self, rcs, criteria, behaviors,
-                            event_description="creation"):
+                            event_description="creation",
+                            add_cleanup_to=None):
```
manishtomar: Non-blocking thought: Do you think it would help if `MimicNova` constructs with test case object? This way it can always cleanup with that object.

manishtomar:
LGTM modulo comment :)

otter/integration/lib/mimic.py
cyli: You're right, it doesn't - I was following a pattern which doesn't really need to apply here unless we have some use for it.  Thanks :)

otter/integration/lib/mimic.py
cyli: Sure, that works :)

cyli:

Issue #1448, opened 2015-05-13
Group status ERROR reasons for user consumption


manishtomar:
Forked from https://github.com/rackerlabs/otter/pull/1445#issuecomment-101774548

When a group's status is changed to ERROR, and event is logged in cloud feeds that should contain reason behind putting group in ERROR. This ticket is to discuss technique used to convert exception reasons returned from step executions to user-readable form *and* to add reasons to the log. The original issue that discussed about this log is second task of #1109

manishtomar:
Closing as completed in #1467

PR #1467, opened 2015-06-04
Better logging for "reasons" for step errors

radix:
This addresses #1448.

There are a couple changes in this branch related to the "reasons" associated with step errors.

- gave a specific type to the elements of the 'reasons' list, `ErrorReason`, a sumtype of Exception, Structured, and Other
- include the traceback along with the exception instance in Exception.ErrorReason
- log the traceback along with the exception in the `execute-convergence-results` message
- also add a mechanism for converting these reasons to end-user-appropriate versions, to be included in the Cloud Feeds `'group-status-error'` event.
- add user-presentable messages for NoSuchCLBError and CLBDeletedError, as a start.

edit: oh yeah, and this adds a dependency on the new `sumtypes` library

### DIFF
--- a/otter/convergence/errors.py
+++ b/otter/convergence/errors.py
@@ -1,0 +1,33 @@
+from singledispatch import singledispatch
+
+from sumtypes import match
+
+from otter.cloud_client import CLBDeletedError, NoSuchCLBError
+from otter.convergence.model import ErrorReason
+
+def present_reasons(reasons):
+    """
+    Get a list of user-presentable messages from a list of :obj:`ErrorReason`.
+    """
+    @match(ErrorReason)
+    class _present_reason(object):
+        def Exception(exc_info): return _present_exception(exc_info[1])
+        def _(_): return None
+    return filter(lambda: None, map(_present_reason, reasons))
+
+
+@singledispatch
+def _present_exception(exception):
+    """Get a user-presentable message or None from an exception instance."""
+    return None
+
+
+@_present_exception.register(NoSuchCLBError)
+def _present_no_such_clb_error(exception):
+    return "Cloud Load Balancer does not exist: %s" % (exception.lb_id,)
+
+
+@_present_exception.register(CLBDeletedError)
+def _present_clb_deleted_error(exception):
+    return ("Cloud Load Balancer is currently being deleted: %s"
+            % (exception.lb_id,))
--- a/otter/convergence/errors.py
+++ b/otter/convergence/errors.py
@@ -1,9 +1,14 @@
+import traceback
+
+from toolz.functoolz import identity
+
 from singledispatch import singledispatch
 
 from sumtypes import match
 
 from otter.cloud_client import CLBDeletedError, NoSuchCLBError
 from otter.convergence.model import ErrorReason
+from otter.log.formatters import serialize_to_jsonable
 
 def present_reasons(reasons):
     """
@@ -31,3 +36,12 @@ def _present_no_such_clb_error(exception):
 def _present_clb_deleted_error(exception):
     return ("Cloud Load Balancer is currently being deleted: %s"
             % (exception.lb_id,))
+
+
+@match(ErrorReason)
+class structure_reason(object):
+    def Exception(exc_info):
+        return  {
+            'exception': serialize_to_jsonable(exc_info[1]),
+            'traceback': ''.join(traceback.format_exception(*exc_info))}
+    _ = identity
--- a/otter/convergence/errors.py
+++ b/otter/convergence/errors.py
@@ -10,14 +10,19 @@ from otter.cloud_client import CLBDeletedError, NoSuchCLBError
 from otter.convergence.model import ErrorReason
 from otter.log.formatters import serialize_to_jsonable
 
+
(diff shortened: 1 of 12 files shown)


radix:
gonna make a change to this...

otter/convergence/steps.py
```diff
@@ -38,7 +38,12 @@ class IStep(Interface):
     """
 
     def as_effect():
-        """Return an Effect which performs this step."""
+        """
+        Return an Effect which performs this step.
+
+        :return: A two-tuple of a :obj:`StepResult` and a list of
```
manishtomar: +1. Thank you :)

otter/test/convergence/test_steps.py
```diff
@@ -573,9 +580,15 @@ def get_result(response, body):
...
+        any_api_error = ErrorReason.Exception(
+            (APIError, matches(IsInstance(APIError)), ANY))
+
         # Fail on 404 or 422 PENDING_DELETE
-        self.assertEqual(get_result(StubResponse(404, {}), ''),
-                         (StepResult.FAILURE, [matches(IsInstance(APIError))]))
+        self.assertEqual(
+            get_result(StubResponse(404, {}), ''),
+            (StepResult.FAILURE,
+             [ErrorReason.Exception(
```
manishtomar: Did you intend to use `[any_api_error]` here?

otter/convergence/errors.py
```diff
@@ -0,0 +1,53 @@
...
+from sumtypes import match
+
+from toolz.functoolz import identity
+
+from otter.cloud_client import CLBDeletedError, NoSuchCLBError
+from otter.convergence.model import ErrorReason
+from otter.log.formatters import serialize_to_jsonable
+
+
+def present_reasons(reasons):
```
manishtomar: Could you add tests for this too?

otter/test/convergence/test_service.py
```diff
@@ -790,7 +842,7 @@ def plan(*args, **kwargs):
                                status=ScalingGroupStatus.ERROR),
              noop),
             (Log('group-status-error', dict(isError=True, cloud_feed=True,
-                                            status='ERROR')),
+                                            status='ERROR', reasons='')),
```
manishtomar: Not super important but will be nice to see actual presentable reason to ensure `present_reasons` is called. I am ok without it too :)

radix: Huh, I thought I had a test with reasons, I must've forgot! Thanks for pointing that out :)

otter/convergence/errors.py
radix: Done, thanks!

otter/test/convergence/test_steps.py
radix: Yes! Thanks for pointing that out. Fixed :)

otter/test/convergence/test_service.py
radix: Done! There was some important code being left untested here, especially the `'; '.join` part. Thanks!

otter/convergence/errors.py
```diff
@@ -0,0 +1,53 @@
...
+            return _present_exception(exc_info[1])
+
+        def _(_):
+            return None
+
+    return filter(None, map(_present_reason, reasons))
+
+
+@singledispatch
+def _present_exception(exception):
```
manishtomar: Non-blocking: What are your thoughts on using a interface `IUserException` that has method to present exception instead of single dispatch?

otter/convergence/model.py
```diff
@@ -112,6 +117,14 @@ class StepResult(Names):
     """
 
 
+@sumtype
+class ErrorReason(object):
+    """A reason for a step to be in a RETRY or FAILURE state."""
+    Exception = constructor('exc_info')
+    Other = constructor(reason=attr.ib(validator=instance_of((unicode, str))))
```
manishtomar: How about `String = constructor...`? That way it will be instantiated by `Exception.String` which I think makes more sense than `Exception.Other`?

radix: Yeah, that makes sense. At first I was thinking it would allow arbitrary objects, but I guess I forgot about that idea when I added the str/unicode constraint :) I'll rename it to String.

otter/convergence/errors.py
radix: I'm kinda torn. It seems nice to put the user-oriented messages near the exceptions, but it also seems nice to put the user-oriented messages all in one place.  (the latter reminds me of our log messages in "spec.py").

manishtomar: Yeah. It seems fine either way. Lets go ahead with this code: user-facing texts in one place.

manishtomar:
One thing I am not sure is about reason being dict like in CLB steps due to which I was not sure if `Exception.Structured` is needed. But we can think about that later. This looks good for now :) 

:+1:

radix:
Yeah, I'm also kinda skeptical about Structured (even though I think I advocated it at one point) -- I just added it to represent the cases we already had.

manishtomar:
I've tested this and found bug in `spec.py` but this PR is good to go :)

PR #1468, opened 2015-06-04
CATC-23

kivattik:


otter/integration/lib/cloud_load_balancer.py
```diff
@@ -264,3 +266,36 @@ def poll():
...
+    Attribute('clb', instance_of=CloudLoadBalancer),
+    Attribute('rcs', instance_of=TestResources)
+])
+class CloudLoadBalancerController(object):
+    """
+    The CloudLoadBalancerController class provides an interface that allows
+    out-of-band control over a cloud load balancer through Mimic's API.
+    """
+
+    def enterPendingDeleteState(self):
```
cyli: Can this be a more general "change_clb_state" function, that can set it to "PENDING_DELETE", "PENDING_UPDATE", or "ERROR" (e.g. it just takes a state to set it to)?

otter/integration/tests/test_convergence.py
```diff
@@ -425,6 +431,87 @@ def _remove_metadata(self, ids, rcs):
...
+    def test_clb_pending_delete_on_scale_up(self):
+        """
+        Simulate getting a 422 on scale up from CLB when in PENDING_DELETE
+        which should cause the group to error.
+        - Create a group with non-zero min servers and a CLB (or more)
+        - Set mimic to return 422 and state PENDING_DELETE on addition to CLB
+        - Attempt to scale up
+        - Group goes into error state since it cannot take action
+
+        Prereq: MUST be able to use Mimic to simulate incorrect CLB behavior
```
cyli: We can add the `@skip_if(not_mimic, "This requires Mimic for error injection.")` decorator instead of the comment, which will skip if the environment variable for mimic isn't set.  :)

otter/integration/tests/test_convergence.py
```diff
@@ -1210,3 +1297,21 @@ def _only_delete_clb_while_scaling(cls, name, method):
 copy_test_methods(
     ConvergenceTestsWith1CLB, ConvergenceTestsWith2CLBs,
     filter_and_change=ConvergenceTestsWith2CLBs._only_delete_clb_while_scaling)
+
+def statusMatches(s, capacities=None):
```
cyli: This would only be applicable if the group is active and desired is 0, or the group is in error somehow (because the pending capacity has to be equal to the desired capacity, as the function is currently defined).  Can we make this more general and just check that the state matches a given status, pending, and desired?

(Also, the group can only be in ACTIVE or ERROR - BUILD status only applies to Nova servers).

otter/integration/tests/test_convergence.py
```diff
@@ -1143,13 +1230,13 @@ def test_delete_all_loadbalancers_and_scale_down(self, delete_clb=None):
...
                 group.wait_for_state,
-                MatchesAll(
-                    ContainsDict({
-                        'pendingCapacity': Equals(0),
-                        'desiredCapacity': Equals(0),
-                        'status': Equals("ACTIVE")
-                    }),
-                    HasActive(0)),
+                MatchesAll(HasActive(0), statusMatches("ACTIVE", capacities=0)),
+#                    ContainsDict({
```
cyli: These commented-out lines can probably be removed.

otter/integration/tests/test_convergence.py
```diff
@@ -1106,13 +1193,13 @@ def test_delete_all_loadbalancers_and_scale_up(self, delete_clb=None):
...
                 group.wait_for_state,
-                MatchesAll(
-                    ContainsDict({
-                        'pendingCapacity': Equals(2),
-                        'desiredCapacity': Equals(2),
-                        'status': Equals("ERROR")
-                    }),
-                    HasActive(0)),
+                MatchesAll(HasActive(0), statusMatches("ERROR", capacities=2)),
+#                    ContainsDict({
```
cyli: These commented-out lines can probably be removed.

kivattik:
These changes have been incorporated into #1473 in order to follow the mimic_nova example that was added after this PR was started.

PR #1469, opened 2015-06-05
cafe-process can discover modules

manishtomar:
Also added subprocess limit that defaults to 10. Examples: `cafe-process.py autoscale dev-worker -p system -l 10` will run all system tests with maximum 10 subprocesses. I ran `python scripts/cafe-process.py autoscale dev-convergence -p system -t convergence=yes` and all the tests passed :). I'll change the jobs to use this once it is merged. 

However, running functional tests fail because `test_scaling_group_pagination` and `test_get_list_scaling_group` rely on çertain number of groups in the tenant which is why `--exclude` option has been added that will exclude those modules from running. So, running `python scripts/cafe-process.py autoscale dev-convergence -p functional -e test_scaling_group_pagination test_get_list_scaling_group` works fine.

One more change is that stdout and stderr of only failed tests are printed on screen.


scripts/cafe-process.py
```diff
@@ -11,28 +11,67 @@
...
+    return autoscale.__path__[0]
+
+
+def get_cafe_args(packages, modules, excludes):
+    if modules:
+        proc_args = []
+        if packages:
+            proc_args.extend(['-p'] + packages)
+        return [proc_args + ['-m', module]
+                for module in modules if module not in excludes]
```
cyli: I think this and the next section assume that excludes is not `None`, which it would be if excludes were not provided, and hence there would be an exception raised.  Should we set a default, like `excludes = () if excludes is None else excludes`?

cyli:
Other than the comment above, LGTM!

scripts/cafe-process.py
manishtomar: Yes. you are right. Thanks :) I'll do that change.

cyli:
If we're adding the env to the cafe-process file, should we also make it executable?  Other than that LGTM and can be merged!

cyli:
Actually I'll just do that and merge, if that's ok with you.  :)

PR #1471, opened 2015-06-05
CATC25: Test that if Nova returns an over-quota error, the group goes into error

### LINKED ISSUE #1333: CATC25: Simulate Nova Quota error (403: Quota), then recovery
Simulate Nova Quota error (403: Quota), then recovery
	- Create a group
	- Cause mimic to return 403: Quota errors on scale up
	- Wait for a time, then fix the quota problem (i.e. Nova now returns normal codes)
	- Validate that scale up completes

Prereq: Mimic MUST be able to have Nova return a predefined error (403) for a time, then revert to correct behavior

cyli:
This is currently skipped because otter doesn't handle that case.

Fixes #1333

### DIFF
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -1047,7 +1047,6 @@ class ConvergenceTestsNoLBs(unittest.TestCase):
             ), timeout=600)
 
 
-
 def _catc_tags(start_num, end_num):
     """
     Return a list of CATC tags corresponding to the start test number and end
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -991,6 +991,62 @@ class ConvergenceTestsNoLBs(unittest.TestCase):
                       matcher=HasLength(2), timeout=600)
         return d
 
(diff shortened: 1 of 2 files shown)


kivattik:
+1 LGTM

otter/integration/tests/test_convergence.py
```diff
@@ -991,6 +991,61 @@ def test_false_negative_on_server_create_from_nova_no_overshoot(self):
...
+            "Quota exceeded for ram: Requested 1024, but already used 131072 "
+            "of 131072 ram")
+
+        behavior_id = yield mimic_nova.sequenced_behaviors(
+            self.rcs,
+            criteria=[{"server_name": server_name_prefix + ".*"}],
+            behaviors=[
+                {"name": "default"},
+                {"name": "fail",
+                 "parameters": {"code": 403, "message": over_quota_message,
```
kivattik: This seems to assume that Otter will error on the first 403, is that correct? Either way I think this is non-blocking until the Otter behavior being tested is implemented.

cyli: Yep, that is the assumption.  Should it not be?

PR #1473, opened 2015-06-08
CATC-23 - Handle 422 PENDING_DELETE

### LINKED ISSUE #1331: CATC23: Simulate “Unable to add Node” from CLB
Simulate getting a 422 on scale up from CLB when in PENDING_DELETE which should cause the group to error

- Create a group with non-min servers and a CLB (or more)
- Set mimic to return 422 and state PENDING_DELETE on the CLB
- Attempt to scale up
- Verify that the group goes into error state since it cannot take action

Prereq: MUST be able to use Mimic to simulate incorrect CLB behavior

kivattik:
Fixes #1331

Validate that receiving a 422, PENDING_DELETE when adding a node to the load balancer causes the group to enter error state.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -307,6 +307,10 @@ class ScalingGroup(object):
         success_codes = [200, 404] if success_codes is None else success_codes
 
         def decide(resp):
+            if verbosity > 0:
+                print('ScalingGroup.get_scaling_group_state content: ')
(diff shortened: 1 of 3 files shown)


kivattik:
Retest this please

kivattik:
Please retest this

sam-falvo:
Closing because it's a duplicate (and inferior) PR to #1474.

PR #1474, opened 2015-06-09
CATC-23 Handle scale down 422 PENDING_DELETE

### LINKED ISSUE #1331: CATC23: Simulate “Unable to add Node” from CLB

kivattik:
Fixes #1331 to handle pending delete

### DIFF
--- a/otter/integration/lib/mimic.py
+++ b/otter/integration/lib/mimic.py
@@ -118,3 +118,44 @@ class MimicNova(object):
         d.addCallback(check_success, [204, 404])
         d.addCallback(self.treq.content)
         return d
+
+
+@attributes(["pool",
+             Attribute("test_case", default_value=None),
+             Attribute("treq", default_value=treq)])
+class MimicCLB(object):
+    """
+    Class that handles HTTP requests to the mimic Cloud Load Blancer
+    control plane.
+
+    Please see the mimic control plane API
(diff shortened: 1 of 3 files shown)


otter/integration/lib/mimic.py
```diff
@@ -118,3 +118,45 @@ def delete_behavior(self, rcs, behavior_id, event_description="creation"):
...
+        all treq requests
+    :ivar test_case: a :class:`twisted.trial.unittest.TestCase`, which if not
+        None, will be used to clean up added behaviors.
+    :ivar treq: the treq module to use for making requests - if not provided,
+        the default library :mod:`treq` will be used.  Mainly to be used for
+        injecting stubs during tests.
+    """
+    def set_clb_attributes(self, rcs, clb_id, kvpairs):
+        """
+
```
sam-falvo: Non-blocking: remove unnecessary whitespace?  Or, replace whitespace with description?

sam-falvo:
LGTM +1

PR #1478, opened 2015-06-11
Skip mimic specific test when not using mimic

manishtomar:


cyli:
Oops!  Sorry I missed that one!  Thanks!  LGTM once tests pass :)

PR #1479, opened 2015-06-11
Only private network when creating servers

manishtomar:
This is necessary when spinning many real servers in prod Nova since public IP address is scarce resource.


cyli:
Oops, yes, thanks for adding that!

PR #1480, opened 2015-06-11
Converger run id during sudden wake up also

manishtomar:
When converger is woken up when any divergent flag is added, then also `converger_run_id` log field is added apart from when it wakes up normally on a regular basis. This is useful when debugging.


cyli:
LGTM!

PR #1492, opened 2015-06-13
Test that otter recovers from impersonation failures

cyli:
It retries when it gets 401s and 500's.

By using a new tenant ID just for this test, we make sure otter doesn't have any of these creds cached (because otter otherwise does not hit identity again).

We can't easily test identity returning 401s and 500s on otter authenticating its service user, because

1. that would affect every single test that gets run concurrently
2. in order to ensure that otter doesn't have its service user creds cached, we'd have to make sure this test gets run before any other test.  or wait for the creds to expire before running this test in isolation.

Note that this is dependent upon the following first before this can be run on jenkins:
- [x] https://github.com/rackerlabs/mimic/pull/316 getting merged
- [x] bumping our CI system's mimic to include the above when it gets merged (see https://github.com/rackerlabs/autoscaling-chef/pull/735/)
- [x] adding a new test convergence tenant to our CI system. (same PR as above)

### DIFF
--- a/otter/integration/tests/test_convergence.py
+++ b/otter/integration/tests/test_convergence.py
@@ -47,13 +47,16 @@ otter_url = os.environ.get('AS_AUTOSCALE_LOCAL_URL')
 nova_key = os.environ.get('AS_NOVA_SC_KEY', 'cloudServersOpenStack')
 clb_key = os.environ.get('AS_CLB_SC_KEY', 'cloudLoadBalancers')
 
+# if this is None, the test will be skipped
+convergence_tenant_auth_errors = os.environ.get(
+    'AS_CONVERGENCE_TENANT_FOR_AUTH_ERRORS')
+
 # these are the service names for mimic control planes
 mimic_nova_key = os.environ.get("MIMICNOVA_SC_KEY", 'cloudServersBehavior')
 mimic_clb_key = os.environ.get("MIMICCLB_SC_KEY", 'cloudLoadBalancerControl')
 
 # otter configuration options for testing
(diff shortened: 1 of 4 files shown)


otter/integration/tests/test_convergence.py
```diff
@@ -868,7 +872,7 @@ def test_servers_that_build_for_too_long_time_out_and_are_replaced(self):
             criteria=[{"server_name": server_name_prefix + ".*"}],
             behaviors=[
                 {"name": "build",
-                 "parameters": {"duration": otter_build_timeout * 2}},
+                 "parameters": {"duration": otter_build_timeout * 3000}},
```
cyli: Note to reviewer - I saw that this test seems to intermittently fail sometimes on CI - am hoping this will be a quick fix (I didn't see any errors from the otter logs, so I figured things just got delayed so that the server became active) - just snuck it in here.

otter/tap/api.py
```diff
@@ -303,7 +303,7 @@ def setup_converger(parent, kz_client, dispatcher, interval, build_timeout):
     Create a Converger service, which has a Partitioner as a child service, so
     that if the Converger is stopped, the partitioner is also stopped.
     """
-    converger_buckets = range(1, 10)
+    converger_buckets = range(10)
```
cyli: Note to reviewer - also snuck in this change - I've filed issue https://github.com/rackerlabs/otter/issues/1493 to deal with it long term but this is a quick fix for now.

kivattik:
+1 Great work as always. :)  Testing that we refresh the token on auth failures through third parties (ex. Nova) is out of scope for this convergence effort, so this is perfect.

cyli:

PR #1494, opened 2015-06-15
cache the pip cache on travis

radix:
According to the docs here http://docs.travis-ci.com/user/caching/

Apparently we can't just use `cache: pip` because we've specified an `install` step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ sudo: false
 python:
     - 2.7
 
+cache:
+  directories:
+    - $HOME/.cache/pip
+
 install:
     - pip install -r dev_requirements.txt
     - pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


radix:
first build took 2:49. I'm rerunning via the travis UI it to see if it picks up the cached artifacts.

radix:
:scream: the rebuild took longer: 4m6s.

radix:
ah. it wasn't caching the *built* dependencies, just the downloads, cause travis is still on pip 6. I threw in a `pip install -U pip wheel`, hopefully that'll speed it up.

radix:
ok, first build after upgrading to 7: 2m57s on first build (where it still had to compile stuff), and 1m7s on the second build (and I've confirmed it's not compiling / building wheels on the second build). Yay!

manishtomar:
LGTM :)

PR #1495, opened 2015-06-15
Pyrsistent version bump and environment variable fix

cyli:
~~Waiting for https://github.com/rackerlabs/autoscaling-chef/pull/736 before this can be tested on jenkins.~~ Done.

Pyrsistent has removed `PMap.set_in`, so I just added it as an FP utility function with as little change to the signature and functionality as possible.

### DIFF
--- a/otter/util/fp.py
+++ b/otter/util/fp.py
@@ -4,7 +4,7 @@
 
 from copy import deepcopy
 
-from pyrsistent import PMap, freeze, thaw
+from pyrsistent import freeze, pmap
 
 from toolz.itertoolz import groupby
 
@@ -90,7 +90,7 @@ def assoc_obj(o, **k):
     return new_o
 
 
-def set_in(old_dict, keys, new_value):
+def set_in(mapping, keys, new_value):
     """
     Take the old dictionary and traverses the dictionary via the list of
     keys.  The returned dictionary will be the same as the old dictionary,
@@ -102,7 +102,7 @@ def set_in(old_dict, keys, new_value):
     Note that the new value does not need to be a pyrsistent data structure -
     this function will freeze everything first.
 
-    :param dict old_dict: The dictionary to change values for.
+    :param dict mapping: The dictionary to change values for.
     :param iterable keys: An ordered collection of keys
     :param new_value: The value to set the keys to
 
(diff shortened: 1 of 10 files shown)


cyli:
Oh oops, looks like the API has changed a bit in 0.10.3.  Fixing...

cyli:

otter/util/fp.py
```diff
@@ -86,3 +88,42 @@ def assoc_obj(o, **k):
...
+    :param iterable keys: An ordered collection of keys
+    :param new_value: The value to set the keys to
+
+    :return: A copy of the old dictionary as PMap, with the new value.
+    """
+    if len(keys) < 1:
+        raise ValueError("Must provide one or more keys")
+
+    if isinstance(old_dict, PMap):
+        new_mutable = thaw(old_dict)
```
radix: This can be formulated a bit nicer (and, I think, more performantly) by using a recursive function on the PMap API, instead of normalizing to mutable dicts and then freezing. Just to make sure I wasn't spouting nonsense, I coded it up to make sure :) how's about this:

```python
    if len(keys) < 1:
        raise ValueError("Must provide one or more keys")

    if isinstance(mapping, dict):
        mapping = freeze(mapping)

    if len(keys) == 1:
        return mapping.set(keys[0], new_value)
    else:
        child = mapping.get(keys[0], pmap())
        return mapping.set(keys[0], set_in(, keys[1:], new_value))
```

I made sure it passed all the existing tests.

cyli: Ah, ok I was hoping to avoid creating _n_ new pmaps given _n_ keys, since I seem to be afraid of pyrsistent data structures now.  But I guess that's the whole point of this PR - to fix the segfaults, and it's not like we're using this deeply nested dictionaries.

Will make the change, thanks.

Also, is it `return mapping.set(keys[0], set_in(child, keys[1:], new_value))`?

radix: oops, yes. I think I accidentally hit undo while I was trying to copy :)

cyli: Ok, minor change to ensure that the new value is also a `pyrsistent` structure, but that is the only deviation.

cyli:

radix:
+1 :)

PR #1496, opened 2015-06-16
Fix convergence buckets to converge tenants that hash down to 0

### LINKED ISSUE #1493: Change ConvergerService to take num_buckets instead of buckets
This way we can ensure that the buckets are calculated correctly (they must be sequential from 0 to num_buckets-1).

This would prevent bugs such as the code that constructs a ConvergerService passing it a random sequence of numbers, or numbers without 0, etc.

cyli:
Fixes #1493 - previously, tenants whose modulo ends up being 0 would not get converged.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -442,12 +442,13 @@ class Converger(MultiService):
         """
         :param log: a bound log
         :param dispatcher: The dispatcher to use to perform effects.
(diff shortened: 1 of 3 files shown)


cyli:

cyli:

radix:
Yay! Thanks for making this change. +1 :)

PR #1499, opened 2015-06-16
get_rcv3_contents return immediately on NoSuchEndpoint

### LINKED ISSUE #1477: RCv3 gathering on `NoSuchEndpoint`
`otter.convergence.gathering.get_rcv3_contents` should fail immediately if it gets `NoSuchEndpoint` error instead of retrying till the end. This is required since RCv3 is opt-in service for tenants and long retries in gathering causes certain tests to fail.

manishtomar:
Closes #1477.

Earlier, it was returning empty list on `NoSuchEndpoint` after it had finished retrying. Now, it is not doing any retry at all and returning empty list immediately. Similarly, I think other gathering functions should also not have any retries to ensure convergence works on most consistent snapshot of reality.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -201,23 +201,17 @@ def get_rcv3_contents():
     """
     Get Rackspace Cloud Load Balancer contents as list of `RCv3Node`.
     """
-    eff = retry_effect(
-        service_request(ServiceType.RACKCONNECT_V3,
-                        'GET', 'load_balancer_pools'),
-        retry_times(5), exponential_backoff_interval(2))
+    eff = service_request(ServiceType.RACKCONNECT_V3, 'GET',
+                          'load_balancer_pools')
 
     def on_listing_pools(lblist_result):
         _, body = lblist_result
         return parallel([
-            retry_effect(
-                service_request(ServiceType.RACKCONNECT_V3, 'GET',
-                                append_segments('load_balancer_pools',
-                                                lb_pool['id'], 'nodes')),
-                retry_times(5), exponential_backoff_interval(2)
-            ).on(
+            service_request(ServiceType.RACKCONNECT_V3, 'GET',
+                            append_segments('load_balancer_pools',
+                                            lb_pool['id'], 'nodes')).on(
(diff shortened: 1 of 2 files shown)


cyli:
Agree, if we're going to remove retries for RCv3, we should probably remove the retries elsewhere.  If we're going to do so we should probably make sure we have comprehensive logs about when gathering starts, and if it fails and what caused it to fail.

Otherwise, maybe we can just reduce the retries and delays for everything?

Do we want to do that here?  Do we want to do that in other PRs?  Other than that, for its stated purpose, this LGTM.  :)

manishtomar:
> If we're going to do so we should probably make sure we have comprehensive logs about when gathering starts, and if it fails and what caused it to fail.

Should I create a ticket for this? 

> Otherwise, maybe we can just reduce the retries and delays for everything?

I am fine with that. Actually, I added retry here initially so that it breaks out immediately on `NoSuchEndpoint` but its tests seemed to be difficult to write due to `compose_retries`. I then decided to remove retry altogether due to consistency. 

> Do we want to do that here? Do we want to do that in other PRs?

We should do it in other PRs. Honestly, I am leaning towards removing retries altogether and having gathering logs mostly because its easier and faster to code that :) </me being impatient>. What do you think?

cyli:
I think you're right, and we can remove at least the top-level (e.g. on listing all the load balancers, and listing the servers) requests, since:

1.  We should have the logging at any rate, whether or not we retry.
1.  We haven't see too many get calls fail in general with our services (except for Nova getting images resulting in spurious 404s :|).

The main reason I want some sort of retrying is that when we do individual gathering of nodes from LBs, that's a lot of requests, each one possibly causing the whole gathering step to fail.  (If it works 99% of the time, and we gather 5 load balancers' nodes, then there's a ~5% change of the gathering step failing).  So maybe we can only do 1 or 2 retries on the individual load balancers themselves with shorter timeouts?

But we have no records or experience either way with how likely this is to fail, so I don't feel strongly about that either, since I don't know that I'm actually being rational about it.  :)

manishtomar:
Created #1515 for removal of gathering retries.

PR #1500, opened 2015-06-16
CLB weight is optional

manishtomar:
According to CLB docs, weight is returned only for load balancers that use WEIGHTED_LEAST_CONNECTIONS or WEIGHTED_ROUND_ROBIN algorithms. This was causing failure of CLB related tests in prod.


cyli:
LGTM!

manishtomar:
I didn't see the note either until I got the error. Even then I noticed only after talking to CLB team on IRC :)

PR #1502, opened 2015-06-17
Log steps to CF

radix:
This causes all steps to be logged to cloud feeds.

It tries to aggregate multiple similar steps into one log statement. So, for example, multiple CreateServers with the same config are aggregated into one log message like so:

```
cf_msg('convergence-create-servers', num_servers=3, server_config={'foo': 'bar'})
```

And if there were multiple CreateServer steps with different server configs, they would be multiple log messages. (that won't happen in practice for a single convergence iteration, AFAIK).

### DIFF
--- a/otter/convergence/logging.py
+++ b/otter/convergence/logging.py
@@ -12,7 +12,7 @@ from toolz.itertoolz import concat
 
 from otter.convergence.steps import (
     AddNodesToCLB, BulkAddToRCv3, BulkRemoveFromRCv3, ChangeCLBNode,
-    CreateServer, DeleteServer, RemoveNodesFromCLB)
+    CreateServer, DeleteServer, RemoveNodesFromCLB, SetMetadataItemOnServer)
 from otter.log.cloudfeeds import cf_msg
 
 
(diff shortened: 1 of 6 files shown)


otter/convergence/logging.py
```diff
@@ -0,0 +1,141 @@
...
+            'convergence-create-servers',
+            num_servers=len(cfg_steps),
+            server_config=dict(cfg))
+        # We sort the items with `thaw` because PMap does not support
+        # comparison
+        for cfg, cfg_steps in sorted(by_cfg.iteritems(), key=thaw)]
+    return parallel(effs)
+
+
+@_logger(SetMetadataItemOnServer)
```
cyli: Non-blocking, because this is what we have so far, but I think we only use this to set DRAINING, and the fact that it's doing so by setting metadata is an implementation detail.  I wonder if we should bother with logging this at all right now, since it's not supported yet - and maybe we want to log something better?  But we can always just change it later, too.  :)

otter/log/spec.py
```diff
@@ -34,6 +27,29 @@
...
+        "Removing nodes from CLB {lb_id}: {nodes}",
+    "convergence-change-clb-nodes":
+        "Changing nodes on CLB {lb_id}: nodes={nodes}, type={type}, "
+        "condition={condition}, weight={weight}",
+    "convergence-add-rcv3-nodes":
+        "Adding servers to RCv3 LB {lb_id}: {servers}",
+    "convergence-remove-rcv3-nodes":
+        "Removing servers from RCv3 LB {lb_id}: {servers}",
+    "group-status-active": "Group's status is changed to ACTIVE",
+    "group-status-error": "Group's status is changed to ERROR",
```
cyli: Non-blocking - do we no longer want to list reasons a group went into error?

radix: Crap! No, this was a mistake I made while resolving conflicts. I'll fix it.

otter/test/utils.py
```diff
@@ -713,11 +713,14 @@ def function(wrapper_intent):
...
 
-def test_dispatcher():
-    return ComposedDispatcher([
+def test_dispatcher(disp=None):
+    disps = [
         base_dispatcher,
         TypeDispatcher({ParallelEffects: perform_parallel_async}),
-    ])
+    ]
+    if disp is not None:
```
cyli: YESSSSSS!!!!!  Thank you, this will be super useful.

otter/convergence/logging.py
radix: I was actually not going to write this logger at first, but then I figured it's probably best to just announce all observable changes to the customer so they have a full audit trail of what's happening to their servers.

radix: I also think having a step like "SetDraining" would be fine.

cyli:
Can we add a test that asserts that nothing will be produced if a non-specified step is returned?  Like `ConvergeLater` or something else? :D

radix:
@cyli done :)

cyli:

cyli:
LGTM!

PR #1503, opened 2015-06-17
msg type is tried from event message also

manishtomar:
For error events, if msg_type is not found in "why" field, it is tried from "message" field since CF error messages will be like that. They will have isError to true but nothing in "why".


cyli:
LGTM!

PR #1506, opened 2015-06-18
Always log convergence results and mark active on success, even if there are no steps

### LINKED ISSUE #1505: Convergence does not always mark group as ACTIVE correctly after it's been in ERROR
This is the bug that CATC-25 is testing, apologies, not #1470.

1.  Group goes into error because there was a fatal issue with Nova.
1.  User fixes the problem, and scales up again.
1.  Convergence produces steps to create the missing servers, resulting in a `StepResults.RETRY` because we have to wait for the servers to be active.  Because the result is a `RETRY`, we don't mark the group as ACTIVE because we only do that in the `StepResults.SUCCESS` case.
1.  Next round of convergence, all the servers are active.  We have achieved our desired state - the planner produces no steps.  As a result, we record the active servers and then just return straightaway from `execute_convergence` with a `StepResults.SUCCESS` without doing any further work (like executing the steps, or marking the group as 'ACTIVE'.

So the group remains in ERROR, even though everything's all good.

cyli:
Fixes #1505

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -91,6 +91,38 @@ def _update_active(scaling_group, active):
 
 
 @do
+def _execute_steps(steps):
+    """
+    Given a set of steps, executes them, logs the result, and returns the worst
+    priority with a list of reasons for that result.
+
+    :return: a tuple of (:class:`StepResult` constant., list of reasons)
+    """
+    if len(steps) > 0:
+        results = yield steps_to_effect(steps)
+
+        severity = [StepResult.FAILURE, StepResult.RETRY, StepResult.SUCCESS]
+        priority = sorted(results,
+                          key=lambda (status, reasons): severity.index(status))
+        worst_status = priority[0][0]
+        results_to_log = zip(
+            steps,
+            [(result, map(structure_reason, reasons))
+             for result, reasons in results])
+
+        reasons = reduce(operator.add,
+                         (x[1] for x in results if x[0] == worst_status))
+    else:
+        worst_status = StepResult.SUCCESS
+        results_to_log = reasons = []
+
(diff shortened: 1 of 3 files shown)


otter/convergence/service.py
```diff
@@ -91,6 +91,38 @@ def update_group_state(group, old_state):
...
+        priority = sorted(results,
+                          key=lambda (status, reasons): severity.index(status))
+        worst_status = priority[0][0]
+        results_to_log = zip(
+            steps,
+            [(result, map(structure_reason, reasons))
+             for result, reasons in results])
+
+        reasons = reduce(operator.add,
+                         (x[1] for x in results if x[0] == worst_status))
```
cyli: To the reviewer:  I wasn't sure why we wanted to pass all possible reasons to `cf_err` so I just went with only returning reasons specifically associated with the worst status.

I know the cloudfeeds logging only logs exceptions it recognizes, but if there happened to be something that sometimes results in a RETRY but is a recognizable exception, we might not want to log that.  I can't think of any right now, though.

cyli:

otter/convergence/service.py
manishtomar: Good point. CF logs should only log reasons for ERROR step result.

otter/convergence/service.py
```diff
@@ -91,6 +91,38 @@ def update_group_state(group, old_state):
 
 
 @do
+def _execute_steps(steps):
```
manishtomar: Thank you for splitting this out :) I was going to suggest a separate refactor PR before even looking at this one.

radix:
LGTM :) Sorry for creating this bug!

PR #1509, opened 2015-06-18
Scale up cloudcafe time convergence

cyli:
Some of the tests are failing because convergence is taking a little long to get to the right state (especially if it's also converging other groups and throttling).

The policy execution tests in particular start a group with some number of servers, and then test that further scaling up or down works.  Some of them have been failing in the setup (where it starts a group with some number of servers).

It times out at 120 seconds or so, and looking at the logs, that's right around when convergence gets around to finishing issuing the create server requests.

So increase the timeouts by 2 for now, if against convergence systems.  Note that the 'convergence' configuration option has already been set on all our configs - we're just going to read it in here.


cyli:
Actually, nm, closing this for now - we'll see if #1499 fixes this, since the reason it's probably taking so long is due to the failed gathering retries.

PR #1511, opened 2015-06-18
CATC-24: Group enters error state on Nova 400

kivattik:


sam-falvo:
+1 LGTM

otter/integration/tests/test_convergence.py
```diff
@@ -993,6 +993,47 @@ def test_false_negative_on_server_create_from_nova_no_overshoot(self):
...
+        message_400 = "Can not find requested image"
+
+        yield mimic_nova.sequenced_behaviors(
+            self.rcs,
+            criteria=[{"server_name": server_name_prefix + ".*"}],
+            behaviors=[
+                {"name": "default"},
+                {"name": "fail",
+                 "parameters": {"code": 400, "message": message_400,
+                                "type": "bad_request"}}
```
cyli: It's actually "badRequest" from Nova, which is why this is failing.  Changing it to "badRequest" makes it pass.

This is still a valid test case though - if we get an error that we can't parse and don't know about, we should probably just fail?

cyli: As per some discussion in slack, probably any unknown 4XX should put it into ERROR.  In which case, do we want to have 2 test cases (one for unknowns, one for knowns)?  

And which (if any) do we want to make this one?

otter/integration/tests/test_convergence.py
```diff
@@ -993,6 +993,47 @@ def test_false_negative_on_server_create_from_nova_no_overshoot(self):
                       matcher=HasLength(2), timeout=600)
         return d
 
+    @skip_me("Skipped until this error transition is fixed in Otter #1510")
+    @skip_if(not_mimic, "This requires Mimic for error injection.")
+    @tag("CATC-024")
+    @inlineCallbacks
+    def test_various_nova_40xs(self):
```
cyli: Since this is only 1 400 case, should the test name be `test_nova_bad_request` or something similar?

otter/integration/tests/test_convergence.py
```diff
@@ -1036,6 +1036,46 @@ def test_recovers_from_nova_intermittent_errors(self):
...
+        message_400 = "Can not find requested image"
+
+        yield mimic_nova.sequenced_behaviors(
+            self.rcs,
+            criteria=[{"server_name": server_name_prefix + ".*"}],
+            behaviors=[
+                {"name": "default"},
+                {"name": "fail",
+                 "parameters": {"code": 400, "message": message_400,
+                                "type": "BadRequest"}}
```
cyli: apologies, it's "badRequest" :|

kivattik: :( Ok, one more try. Didn't notice the typo while the test was skipped.

cyli:

cyli:
LGTM!  Thank you!

PR #1512, opened 2015-06-18
Fail when adding nodes to a CLB results in a node limit error

### LINKED ISSUE #1214: AddNodesToCLB should fail on 25 nodes limit
It should return `StepResult.FAILURE` when 25 nodes limit error occurs since group cannot scale anymore after that. Can be done after #1189.

### LINKED ISSUE #1244: CLBChangeNode should handle over limit errors

### LINKED ISSUE #1260: Remove `get_fake_service_request_performer` from otter/test/utils.py
Hopefully, service request performing will just be done in the client (see #1189), and if so, this may no longer be a necessary general utility.

After which it can be removed.

cyli:
If the are more 25 nodes on a CLB and convergence tries to add more, then the step should fail and convergence should put the group into ERROR.

Since we're touching the add nodes code anyway, I moved it to cloud_client as per #1189.

This makes it so that all 4XX errors, unless we know we can retry, fail.  Other errors are retried.

Fixes #1214.  
Fixes #1244.
Fixes #1260.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -322,17 +322,19 @@ def get_cloud_client_dispatcher(reactor, authenticator, log, service_configs):
 
 _CLB_PENDING_UPDATE_PATTERN = re.compile(
     "^Load Balancer '\d+' has a status of 'PENDING_UPDATE' and is considered "
-    "immutable.$")
+    "immutable\.$")
 _CLB_DELETED_PATTERN = re.compile(
     "^(Load Balancer '\d+' has a status of 'PENDING_DELETE' and is|"
-    "The load balancer is deleted and) considered immutable.$")
+    "The load balancer is deleted and) considered immutable\.$")
 _CLB_NO_SUCH_NODE_PATTERN = re.compile(
     "^Node with id #\d+ not found for loadbalancer #\d+$")
 _CLB_NO_SUCH_LB_PATTERN = re.compile(
     "^Load balancer not found.$")
 _CLB_DUPLICATE_NODES_PATTERN = re.compile(
     "^Duplicate nodes detected. One or more nodes already configured "
-    "on load balancer.$")
+    "on load balancer\.$")
+_CLB_NODE_LIMIT_PATTERN = re.compile(
+    "^Nodes must not exceed \d+ per load balancer\.$")
 
 
 @attributes([Attribute('lb_id', instance_of=six.text_type)])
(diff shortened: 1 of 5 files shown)


cyli:

cyli:

cyli:
Actually, I'm going to try to get #1524 merged first, since I'm going to need those for another PR.

cyli:
Ok, this is probably ready for review again.  #1513 still passes with master merged in and changes.

cyli:

otter/cloud_client.py
```diff
@@ -416,6 +439,52 @@ def try_parsing(api_error_exc_info):
...
+        }
+
+        (weight and type are optional)
+
+    :return: :class:`ServiceRequest` effect
+
+    :raises: :class:`CLBPendingUpdateError`, :class:`CLBDeletedError`,
+        :class:`NoSuchCLBError`, :class:`CLBDuplicateNodesError`,
+        :class:`APIError`
+    """
```
manishtomar: Great docs :)

otter/convergence/steps.py
```diff
@@ -350,18 +341,22 @@ def report_api_failure(result):
...
-                    ErrorReason.Exception(result)]
-            if error.code == 422:
-                message = try_json_with_keys(error.body, ("message",))
-                if message and _CLB_DELETED_PATTERN.search(message):
-                    return StepResult.FAILURE, [
-                        ErrorReason.Exception(result)]
-            return StepResult.RETRY, [ErrorReason.Exception(result)]
+            status = StepResult.FAILURE
+
+            known_retryable = err_type in (CLBDuplicateNodesError,
```
radix: I think maybe with a few small abstractions we could simplify this function a lot, but that's just a side thought. LGTM :)

radix:
Great changes to the unit tests, too. Thanks a lot!

PR #1516, opened 2015-06-18
CQL dispatcher refactor

manishtomar:
Moved `get_cql_dispatcher` to models/cass.py with only `CQLQueryExecute` performer. Changed load_cql.py to use new dispatcher. This is needed as part of #1458 splitup.


cyli:
LGTM!

PR #1517, opened 2015-06-18
take_while function similar to one in Haskell

manishtomar:


cyli:
Filed https://github.com/rackerlabs/otter/issues/1518, but LGTM!

PR #1519, opened 2015-06-18
removing takewhile - thanks @glyph

manishtomar:
since its already there in itertools and this is buggy. Thanks a lot for catching this @glyph !


glyph:
Thanks, always happy to eliminate some code :)

cyli:
+1 when travis +1's :)

PR #1520, opened 2015-06-18
Servers cache model

manishtomar:
interface and implementation. This cache will be used by convergence when gathering servers from Nova using changes-since filter. Part of #1458.

Note to reviewer: 
Cache is removed when group is deleted https://github.com/rackerlabs/otter/pull/1458/files#diff-5894116b577c37059087fcce5ce51693R165. I think its better to remove cache as part of group deletion. But adding that will require new query based on `tenant_id` and `group_id`. It would be easier if I chose existing columns `tenantId` and `groupId` as the cache table can be easily appended [here](https://github.com/rackerlabs/otter/blob/change-since/otter/models/cass.py#L1278). Should I change the column names and cache deletion in group deletion?

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -1702,6 +1702,65 @@ class CassScalingGroupCollection:
         return d
 
 
+@implementer(IScalingGroupServersCache)
+class CassScalingGroupServersCache(object):
+    """
+    Collection of cache of scaling group servers
+    """
+
+    def __init__(self, tenant_id, group_id):
+        self.tenant_id = tenant_id
+        self.group_id = group_id
+        self.table = "servers_cache"
+        self.params = {"tenant_id": tenant_id, "group_id": group_id}
+
+    @do
+    def get_servers(self):
+        """
+        See :method:`IScalingGroupServersCache.get_servers`
+        """
+        query = ("SELECT server_blob, last_update FROM {cf} "
+                 "WHERE tenant_id=:tenant_id AND group_id=:group_id "
+                 "ORDER BY last_update DESC;")
+        rows = yield cql_eff(query.format(cf=self.table), self.params)
+        if len(rows) == 0:
+            yield do_return(([], None))
+        last_update = rows[0]['last_update']
+        rows = take_while(lambda r: r['last_update'] == last_update, rows)
+        yield do_return(([json.loads(r['server_blob']) for r in rows],
+                         last_update))
+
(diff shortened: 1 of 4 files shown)


otter/models/cass.py
```diff
@@ -1702,6 +1704,65 @@ def health_check(self):
...
+        See :method:`IScalingGroupServersCache.get_servers`
+        """
+        query = ("SELECT server_blob, last_update FROM {cf} "
+                 "WHERE tenant_id=:tenant_id AND group_id=:group_id "
+                 "ORDER BY last_update DESC;")
+        rows = yield cql_eff(query.format(cf=self.table), self.params)
+        if len(rows) == 0:
+            yield do_return(([], None))
+        last_update = rows[0]['last_update']
+        rows = takewhile(lambda r: r['last_update'] == last_update, rows)
```
radix: Lovely!

otter/models/interface.py
```diff
@@ -614,92 +614,36 @@ def delete_webhook(policy_id, webhook_id):
...
-        :raises NoSuchScalingGroupError: if this scaling group does not exist
-        :raises NoSuchServerIntentError: if the server intent id does not exist
+    def get_servers():
         """
-
-    def list_server_intents(log, status=None, limit=100, marker=None):
+        Return latest cache of servers in a group along with last time the
+        cache was updated.
+        :return: Effect of (servers, last update time) tuple where servers
+            is list of dict and last update time is datetime object. Will
```
radix: I kind of expected servers to be a dict keyed on server ID instead of just a list, though I guess the ID is in practice inside of the server blob? It looks  like you're converting to that format in the code that uses this cache, https://github.com/rackerlabs/otter/pull/1458/files#diff-153acffdf192c6605ff1a2cf8ad03639R131 so maybe it would be nicer to just return it that way in the first place

radix:
Just one minor design comment, which I'll leave up to you. LGTM otherwise!

otter/models/interface.py
manishtomar: That happens to be done due to the way the cache needs to be merged with current servers. The servers after that are list of dicts. Also, the insertion of servers are as list of dicts. I think its fine to leave it like this :)

radix: then it's worth noting that I don't think we're using the server_id column at all

radix:
LGTM :)

PR #1522, opened 2015-06-19
get-all-details changes_since fix

manishtomar:
`get_all_server_details` was sending `changes_since` filter parameter instead of `changes-since` (note the hypen instead of underscore) due to which mimic was not returning deleted servers. Prod nova might have returned 400 but had tested this only with mimic.

Part of #1458

### DIFF
--- a/otter/test/convergence/test_gathering.py
+++ b/otter/test/convergence/test_gathering.py
@@ -74,7 +74,7 @@ def svc_request_args(**params):
     """
     Return service request args with formatted changes_since argument in it
     """
-    changes_since = params.get('changes-since', None)
+    changes_since = params.pop('changes_since', None)
     if changes_since is not None:
         params['changes-since'] = changes_since.isoformat() + 'Z'
     return (ServiceType.CLOUD_SERVERS, 'GET',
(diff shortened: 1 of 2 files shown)


otter/test/convergence/test_gathering.py
```diff
@@ -69,14 +70,15 @@ def resolve_svcreq(eff, result, service_type,
...
 
 
-def svc_request_args(changes_since, limit):
+def svc_request_args(**params):
     """
     Return service request args with formatted changes_since argument in it
     """
-    params = urlencode([('changes_since', changes_since.isoformat() + 'Z'),
-                        ('limit', limit)])
+    changes_since = params.get('changes-since', None)
```
radix: I'm not sure why you made this change, it just seems to make it harder to pass the parameter (i.e. how you have to use `**{'changes-since': ...}` instead of `change_since=...`.   IMO it'd be better to leave the parameter name the same but then just construct `params` appropriately inside this method.

radix:
other than my one note LGTM :)

otter/test/convergence/test_gathering.py
manishtomar: I realized it after putting the PR :). Will change it

PR #1525, opened 2015-06-19
Store server JSON in NovaServer

manishtomar:
This will be used when storing servers in cache at the succesful convergence cycle. See https://github.com/rackerlabs/otter/pull/1458/files#diff-5894116b577c37059087fcce5ce51693R175

Part of #1458.

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -217,9 +217,10 @@ class NovaServer(object):
         the server is on the ServiceNet network
     :ivar str image_id: The ID of the image the server was launched with
     :ivar str flavor_id: The ID of the flavor the server was launched with
-
     :ivar PSet desired_lbs: An immutable mapping of load balancer IDs to lists
         of :class:`CLBDescription` instances.
+    :var dict json: JSON dict received fro mNova from which this server
+        is created
     """
 
     def __init__(self):
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -10,7 +10,7 @@ from attr.validators import instance_of
 
 from characteristic import Attribute, attributes
 
-from pyrsistent import PSet, freeze, pmap, pset, pvector
+from pyrsistent import PMap, PSet, freeze, pmap, pset, pvector
 
 from sumtypes import constructor, sumtype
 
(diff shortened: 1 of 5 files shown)


radix:
Could you update the NovaServer docstring to add the new :ivar: for json, pointing out it's to maintain the originally received json?

radix:
I think we may also want to update `otter.convergence.service.server_to_json` to return the `json` attribute. That's used to populate the `active` list for convergence tenants, and currently it only returns a rather useless {'id': server_id} dict.

It should be a pretty simple change, but if you think it should be done in a separate PR that's fine. I can even do it real quick.

radix:
LGTM after the docstring  fix :)

manishtomar:
> I think we may also want to update otter.convergence.service.server_to_json to return the json attribute. That's used to populate the active list for convergence tenants, and currently it only returns a rather useless {'id': server_id, 'links': ...} dict.

@radix I think that should remain the same to be consistent with worker. Also, note that that dict is returned as part of `/state` call which should not change

radix:
Hm, I guess I was confused about what the legacy `active` list contains. I thought it contained the server data. I just tracked it down and I *think* it contains these keys:

- `id`: server id
- `links`: links (we already support this in convergence)
- `name`: we don't have this yet
- `lb_info`: some LB stuff
- `created`: time created

We don't have `name`, `lb_info`, or `created` in new convergence stuff. I guess that's not important (especially since we don't document these keys anywhere :smile:)

anyway forget about my comment :)

otter/convergence/model.py
```diff
@@ -216,9 +217,10 @@ class NovaServer(object):
         the server is on the ServiceNet network
     :ivar str image_id: The ID of the image the server was launched with
     :ivar str flavor_id: The ID of the flavor the server was launched with
-
     :ivar PSet desired_lbs: An immutable mapping of load balancer IDs to lists
         of :class:`CLBDescription` instances.
+    :var dict json: JSON dict received fro mNova from which this server
```
radix: sorry, you have a typo here :)

radix:
LGTM

otter/convergence/model.py
manishtomar: Oh sorry. Thanks :)

PR #1526, opened 2015-06-19
LBSH tests OOB deletion of a CLB node

### LINKED ISSUE #1481: Servers managed by autoscale SHALL be...1 group, 1 LB in config
Servers managed by autoscale SHALL be only on the nodes on the load-balancers in their launch config after convergence is triggered.

1 group, 1 LB in config:

Server node removed from LB
Assert: Server is put back

cyli:
Otter should put it back on the next convergence cycle.

Note that if we are running these against production and the LB is in PENDING_UDPATE, we want to retry the delete operation, since we don't want the test to fail just on that alone.

Fixes #1481 

This depends upon a bugfix in mimic ~~(forthcoming)~~ (merged, but version bump not in ci yet), so please don't merge yet.

### DIFF
--- a/otter/integration/lib/cloud_load_balancer.py
+++ b/otter/integration/lib/cloud_load_balancer.py
@@ -299,6 +299,28 @@ class CloudLoadBalancer(object):
         d.addCallback(self.treq.content)
         return d
 
+    def delete_nodes(self, rcs, node_ids):
+        """
+        Delete one or more nodes from a load balancer.
+
+        :param rcs: a :class:`otter.integration.lib.resources.TestResources`
+            instance
+
+        :param list node_ids: A list of `int` node ids to delete.
+
+        :return: An empty string if successful.
+        """
+        d = self.treq.delete(
+            "{0}/loadbalancers/{1}/nodes".format(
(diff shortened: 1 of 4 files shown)


otter/integration/lib/cloud_load_balancer.py
```diff
@@ -22,6 +22,20 @@
...
 
+def _pending_update_to_transient(f):
+    """
+    A cloud load balancer locks on every update, so to ensure that the test
+    doesn't fail because of that, we want to retry POST/PUT/DELETE commands
+    issued by the test.  This is a utility function that checks if a treq
+    API failure is a 422 PENDING_UDPATE failure, and if so, re-raises a
+    TransientRetryError instead.
+    """
+    if f.check(APIError):
```
manishtomar: Non-blocking: You could do `f.trap(APIError)` and need not have `return f` in the end.

cyli: Oh yes, thanks!  I forgot about `trap` for some reason

kivattik:
:+1: LGTM

image

cyli:

PR #1529, opened 2015-06-22
servers cache table schema fix

manishtomar:
Had changed column names to tenantId/groupId to be consistent with other tables but forgot to change it in primary key definition.


radix:
LGTM!

PR #1531, opened 2015-06-22
deleting servers are considered DELETED

### LINKED ISSUE #1523: Consider servers with task_state as "deleting" to be DELETED
Nova takes time to delete servers. But in most cases it changes the task_state of the server it is deleting to "deleting" much earlier. This is how worker decides to stop delete retry. Convergence gathering should consider these servers as DELETED so that their CLB ids are removed and new servers are created to replace them (if needed). This should increase the time taken to run OOB deletion convergence tests against prod. They also have more chances of succeeding.

manishtomar:
Closes #1523


radix:

radix:
LGTM :)

PR #1532, opened 2015-06-23
don't blow up when GetStat returns None

radix:
The `GetStat` will return None (indicating the node has been deleted) sometimes, and currently causes an exception about NoneType not having a `version` attribute.

It's safe to ignore this error because it indicates that we were in the process of finishing up convergence of this group. This is because the start of convergence looks like this:

- list divergent flags
- filter out ones that are in `currently_converging`
- for each divergent flag:
  - stat individual znode of flag to get the `version`
  - add group associated with flag to `currently_converging`
  - ...

and the end of convergence (for an individual group) looks like this:

- ...
- delete from `currently_converging`
- delete divergent flag

So, if we list & filter out divergent flags in between those last two steps, and then the deletion happens, by the time we get to "stat individual znode of flag to  get the `version`", it will return None and current code throws an exception when accessing the `version`.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -397,12 +397,19 @@ def converge_all_groups(currently_converging, my_buckets, all_buckets,
               currently_converging=list(cc))
 
     def converge(tenant_id, group_id, dirty_flag):
-        return Effect(GetStat(dirty_flag)).on(
-            lambda stat: Effect(TenantScope(
-                converge_one_group(currently_converging,
-                                   tenant_id, group_id, stat.version,
-                                   build_timeout),
-                tenant_id)))
+        def got_stat(stat):
+            # If the node disappeared, ignore it. `stat` will be None here only
+            # after the group is removed from currently_converging, but before
+            # the divergent flag is deleted. This happens when we're "noticing"
+            # a divergent flag just we're also finishing up a convergence
+            # process for the same group.
+            if stat is not None:
+                return Effect(TenantScope(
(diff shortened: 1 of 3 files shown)


radix:

radix:

radix:
hm. trial integration tests failed. that was unexpected. https://as-jenkins.rax.io/job/_ci-integration-convergence-trial/183/console

I'm going to blindly rerun them and see if it was a spurious failure.

radix:

otter/convergence/service.py
```diff
@@ -397,12 +397,20 @@ def converge_all_groups(currently_converging, my_buckets, all_buckets,
...
-                                   build_timeout),
-                tenant_id)))
+        def got_stat(stat):
+            # If the node disappeared, ignore it. `stat` will be None here if
+            # the divergent flag was discovered only after the group is removed
+            # from currently_converging, but before the divergent flag is
+            # deleted, and then the deletion happens, and then our GetStat
+            # happens. This basically means it happens when one convergence is
+            # starting as another one for the same group is ending.
+            if stat is not None:
```
manishtomar: Non-blocking thought: I wonder if logging the fact that `stat` is None can be useful?

manishtomar:
LGTM modulo non-blocking comment above :)

radix:
I generally prefer if/else to if/return/else, depending on the situation, so I'll leave it that way. Thanks for the review, merging!

Issue #1304, opened 2015-04-06
Migrate existing groups status


manishtomar:
As discussed https://github.com/rackerlabs/otter/pull/1268#discussion_r27781016, all the existing groups should have status to be either "ACTIVE" or "ERROR" or "DELETING". Hence, have a script that sets ACTIVE for groups with status as `None` and "ERROR" for groups having "DISABLED" status.

PR #1533, opened 2015-06-23
metrics: If convergence-tenants is specified, only fetch metrics for  those tenants

### LINKED ISSUE #931: Metrics should run on convergence tenants
Currently metrics collects server data from all tenants. It should have an option to run on certain tenants which mostly will be convergence enabled tenants. It should probably also inject separate metrics for each tenant apart from the total.

radix:
Fixes #931. If `convergence-tenants` is in the config file, only those tenants will be used.

I decided to just write a new function called `get_specific_scaling_groups`, instead of modifying `get_scaling_groups`, because that function is huge and scary. The new one doesn't limit by batch size so if we add TONS of convergence tenants to the config file we may run into performance issues down the line.

Also:

- fixes a bug in metrics where I never updated it to use `get_legacy_dispatcher` instead of `get_full_dispatcher` (and mocks in tests prevented the tests from failing ;_;)
- added a `__str__` to `NoSuchEndpoint` in `otter.auth` just 'cause I noticed it didn't have one while I was running the script to test it

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -32,13 +32,29 @@ from otter.auth import generate_authenticator
 from otter.cloud_client import TenantScope, service_request
 from otter.constants import ServiceType, get_service_configs
 from otter.convergence.gathering import get_scaling_group_servers
-from otter.effect_dispatcher import get_full_dispatcher
+from otter.effect_dispatcher import get_legacy_dispatcher
 from otter.log import log as otter_log
 from otter.util.fp import predicate_all
 
 
 # TODO: Remove this and pass it from service to other functions
 metrics_log = otter_log.bind(system='otter.metrics')
+
+
+@defer.inlineCallbacks
+def get_specific_scaling_groups(client, tenant_ids):
+    tids = ', '.join("'{}'".format(tid) for tid in tenant_ids)
+    query = (
+        'SELECT '
+        '"tenantId", "groupId", desired, active, pending, created_at, status '
+        'FROM scaling_group WHERE "tenantId" IN ({tids})')
+    query = query.format(tids=tids)
+    print("query", query)
+    results = yield client.execute(query, {}, ConsistencyLevel.ONE)
+    defer.returnValue(r for r in results
+                      if r['created_at'] is not None and
+                      r['desired'] is not None and
+                      r['status'] != 'DISABLED')
 
 
 @defer.inlineCallbacks
@@ -243,7 +259,7 @@ def connect_cass_servers(reactor, config):
 def collect_metrics(reactor, config, client=None, authenticator=None,
                     _print=False,
                     perform=perform,
-                    get_full_dispatcher=get_full_dispatcher):
+                    get_legacy_dispatcher=get_legacy_dispatcher):
     """
     Start collecting the metrics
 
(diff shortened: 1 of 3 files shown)


otter/metrics.py
```diff
@@ -41,6 +41,23 @@
...
+
+@defer.inlineCallbacks
+def get_specific_scaling_groups(client, tenant_ids):
+    tids = ', '.join("'{}'".format(tid) for tid in tenant_ids)
+    query = QUERY_GROUPS_OF_TENANTS.format(tids=tids)
+    results = yield client.execute(query, {}, ConsistencyLevel.ONE)
+    defer.returnValue(r for r in results
+                      if r.get('created_at') is not None and
+                      r.get('desired') is not None and
+                      r.get('status') != 'DISABLED')
```
manishtomar: You might want add a check for "ERROR" too unless https://github.com/rackerlabs/otter/issues/1304 is implemented

otter/metrics.py
```diff
@@ -260,18 +277,25 @@ def collect_metrics(reactor, config, client=None, authenticator=None,
 
     :return: :class:`Deferred` with None
     """
+    convergence_tids = None
+    if config.get('convergence-tenants', []):
```
manishtomar: Non-blocking: I think `if "convergence-tenants" in config` is more clear

otter/metrics.py
```diff
@@ -260,18 +277,25 @@ def collect_metrics(reactor, config, client=None, authenticator=None,
...
-    dispatcher = get_full_dispatcher(reactor, authenticator, metrics_log,
-                                     service_configs)
+    dispatcher = get_legacy_dispatcher(reactor, authenticator, metrics_log,
+                                       service_configs)
 
     # calculate metrics
-    cass_groups = yield get_scaling_groups(
-        _client, props=['status'],
-        group_pred=lambda g: g['status'] != 'DISABLED')
+    if convergence_tids:
```
manishtomar: If you decide to use above if statement, this can be changed to `if convergence_tids is not None` to be more explicit. Again, all non-blocking :)

otter/test/test_metrics.py
```diff
@@ -51,6 +53,29 @@
...
+    def test_query(self):
+        def _exec(query, params, c):
+            return succeed(exec_args[(query, freeze(params))])
+        client = mock.Mock(spec=CQLClient)
+        rows = [
+            {'created_at': '0', 'desired': 'some', 'status': 'ACTIVE'},
+            {'desired': 'some', 'status': 'ACTIVE'},  # no created_at
+            {'created_at': '0', 'status': 'ACTIVE'},  # no desired
+            {'created_at': '0', 'desired': 'some'},   # no status
+            {'created_at': '0', 'desired': 'some', 'status': 'DISABLED'}]
```
manishtomar: Similarly, could you add test for "ERROR" too?

otter/metrics.py
radix: Actually, the way I did it was an intentional choice, since I didn't want to run this codepath in the case of an empty list of convergence tenants.

Mostly this was because I was assuming that "SELECT * WHERE FOO IN ()" would be a syntax error, as it in is Postgres. But apparently in CQL it's not a syntax error, so I guess I can change it :)

manishtomar:
LGTM modulo comments :). 

Could you please run this once and try it out? I can also do it if needed :)

radix:
@manishtomar I don't have an environment set up that lets me talk to prod nova, so it'd be cool if you could :)

radix:
but, fwiw, I have actually run the script on my dev environment and tested that the changed code works as expected -- I verified that it fetches the explicitly requested tenants with the right queries.

radix:
ok, merging this! Thanks @manishtomar

PR #1534, opened 2015-06-23
updating converted RST content after check

stanzikratel:


stretchservers:

cyli:
This seems to contain the html result of building all the docs, including the cloud servers docs?  Should this have gone to another repo?

PR #1535, opened 2015-06-23
Convergence test clb log

manishtomar:
`test_delete_all_loadbalancers_and_scale_up` test deletes CLB as part of the test. These tests delete the created CLB as part of cleanup irrelevant of the fact that test deleted it or not. So, it is expected that cleanup might try to delete an already deleted CLB. Prod CLB seems to be returning 400 instead of 404 in this case due to which error gets propagated to trial and the test is reported as failure. This change logs the failure and allows the test to pass.


cyli:
LGTM, thanks for addressing this!

PR #1536, opened 2015-06-23
Add cleanup for servers on trial tests

cyli:
So the ones with their metadata removed will be cleaned up after the test


manishtomar:
LGTM pending build :)

PR #1537, opened 2015-06-24
Test delete server waits until server is deleted

manishtomar:
This is required because prod Nova takes time to delete servers and many OOB tests expect otter to react based on server being deleted. With this merged, we can run all prod convergence tests in https://as-jenkins.rax.io/job/otter-trial-ord-convergence-tests

### DIFF
--- a/otter/integration/lib/test_nova.py
+++ b/otter/integration/lib/test_nova.py
@@ -105,9 +105,6 @@ class NovaServerTestCase(SynchronousTestCase):
                                  Response(204), "delete response")
         d = server.delete(self.rcs)
         self.assertNoResult(d)
-        server.treq = get_fake_treq(
-            self, 'delete', 'novaurl/servers/server_id',
-            ((), self.expected_kwargs), (Response(204), 'del resp'))
         self.clock.pump([5] * 24)
         self.failureResultOf(d)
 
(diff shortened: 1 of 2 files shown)


otter/integration/lib/test_nova.py
```diff
@@ -68,25 +69,47 @@ def setUp(self):
...
         """
         server = self.get_server('delete', 'novaurl/servers/server_id',
                                  ((), self.expected_kwargs),
                                  Response(204), "delete response")
         d = server.delete(self.rcs)
-        self.assertEqual('delete response', self.successResultOf(d))
+        self.assertNoResult(d)
+        server.treq = get_fake_treq(
+            self, 'delete', 'novaurl/servers/server_id',
+            ((), self.expected_kwargs), (Response(204), 'del resp'))
```
kivattik: Is the only reason to replace the fake_treq to change the response message?

kivattik:
+1 LGTM :)

otter/integration/lib/test_nova.py
manishtomar: Oh yeah. Good point. This particular test doesn't need the changed `treq`. Thanks for catching this! Will remove it.

manishtomar:
@kivattik Aww... that is such a lovely pic! Thanks a lot :D

PR #1540, opened 2015-06-24
quick hack to log some info when the lock is held too long

radix:
does a couple things
- `otter.util.deferredutils.with_lock` now logs extra fields:
  - lock_status
  - lock
  - locked_func
- it also runs a delayed call for 120 seconds after acquiring a lock that will log a message *if* the lock is still acquired.
- updated all callers of `with_lock` to bind a `lock_reason` key into the log, so we know who's acquiring locks. Since with_log uses this bound log for the new 120-second-timeout, it will give us an idea of which lock-user is holding the lock too long.
- `with_lock` now defaults the `log` argument to the default otter log instead of just not logging. This was mostly just so I could get rid of the millions of `if` statements we had to placate the linter which was complaining about function complexity.
- modify_state now takes a `modify_state_reason` to bind into the log as `modify_state_reason`. Since `modify_state` is by far the most common acquirer of locks, this will help us further determine which `modify_state` caller might be keeping a lock for too long.

Also, caveat:
- I deleted some really horrible assertions about `log.bind().bind()` in `test_cass_models.py` because I can't even

### DIFF
--- a/otter/test/test_util.py
+++ b/otter/test/test_util.py
@@ -502,6 +502,11 @@ class WithLockTests(SynchronousTestCase):
         self.log = mock_log()
         self.log_fields = {'lock': self.lock, 'locked_func': self.method}
 
+        # This is shared between multiple tests, and used in negative
+        # assertions. Centralizing the definition means that if the message
+        # changes the negative assertions will also be updated.
+        self.too_long_message = "Lock held for more than 120 seconds!"
+
     def test_acquire_release(self):
         """
         Acquires, calls method, releases and returns method's result. Logs time taken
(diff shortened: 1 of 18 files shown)


radix:

radix:

radix:

otter/test/test_util.py
```diff
@@ -564,20 +574,27 @@ def test_method_failure(self):
...
+            [mock.call('Lock acquisition in 10.0 seconds',
+                       acquire_time=10.0, lock_status='Acquired',
+                       **self.log_fields),
+             mock.call('Starting lock release', lock_status='Releasing',
+                       **self.log_fields)])
         self.method.assert_called_once_with()
 
         self.reactor.advance(3)
         self.release_d.callback(None)
         self.log.msg.assert_called_with('Lock release in 3.0 seconds',
```
cyli: Non-blocking - just to test that that message is not logged when the lock has not been held for that long (for the sake of test coverage, since the logic seems like it wouldn't do that), would it make sense to add a call to `self.reactor.advance(120)` here and maybe in a few places?

cyli:
LGTM to me aside from my one testing nitpick!

cyli:
+1

PR #1541, opened 2015-06-24
Create server moved to cloud client

### LINKED ISSUE #1167: CreateServer should return a real exception on failure, not a message
Currently it does: 

```python
        def report_failure(result):
            """
            If the failure is an APIError with a 400 or 403, attempt to parse
            the results and return a :obj:`StepResult.FAILURE` - if the errors
            are unrecognized, return a :obj:`StepResult.RETRY` for now.
            """
            err_type, error, traceback = result
            if err_type == APIError:
                message = _parse_nova_user_error(error)
                if message is not None:
                    return StepResult.FAILURE, [message]
```

### LINKED ISSUE #1510: Post convergence groups should error on any 400 from Nova
As discussed in the comments on #1332, when a group receives a 400: BadRequest response from Nova, the group should transition to ERROR state.

cyli:
Fixes #1167.
Fixes #1510.

Addresses part of #1189.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -5,6 +5,8 @@ import json
 import re
 from functools import partial, wraps
 
+import attr
+
 from characteristic import Attribute, attributes
 
 from effect import (
@@ -341,6 +343,33 @@ _CLB_NODE_LIMIT_PATTERN = re.compile(
 _CLB_OVER_LIMIT_PATTERN = re.compile("^OverLimit Retry\.{3}$")
 
 
+@attr.s(these={"message": attr.ib()}, init=False)
+class ExceptionWithMessage(Exception):
+    """
+    The builtin `Exception` doesn't have equality (it tests for identity).
+    ``attr`` provides equality based on the attributes.
+
+    But letting ``attr`` generate the `__init__` function for you means that
+    means extra kwargs do not get passed to the superclass's
+    ``__init__``.
+
+    So this gives us a base class that does both.  By using providing
+    our own ``__init__`` function, we can make this compatible with the builtin
+    Exception and also get the Python27 Exception's ``__str__`` function
+    automatically.
+
+    Also useful to note that Python 3 does not store the message on the
+    Exception, it just stores the args passed to it, so this lets us have a
+    message attribute in Python 3 as well.
+    """
+    def __init__(self, message):
+        """
+        Set ``self.message`` and also call the base class's ``__init__``.
+        """
+        super(ExceptionWithMessage, self).__init__(message)
+        self.message = message
+
+
 @attributes([Attribute('lb_id', instance_of=six.text_type)])
 class CLBPendingUpdateError(Exception):
     """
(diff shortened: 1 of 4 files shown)


cyli:

otter/cloud_client.py
```diff
@@ -507,6 +524,25 @@ class NovaComputeFaultError(Exception):
     """
 
 
+@attributes([])
```
manishtomar: Is this needed? Is it to ensure that no argument is passed?

cyli: Hmm, you're right, it shouldn't be needed.  Not sure why I put that there - will delete.  Thanks!  :)

cyli: Oh, no, right, it's so that characteristic can set up the equality test for me.  Otherwise passing a unicode message vs a bytes message would mean that that the errors were different.

This actually brings up a question I had about characteristic vs attr btw - I was thinking about converting these to use `@attr.s`, but attr doesn't pass extra arguments through to the `__init__`, and characteristic does.

What that means is that we can't extend `Exceptions` and be able to use `Exception("msg")`.  We *can* define our own `message` attribute to our exceptions.  Wondering if that would be better?

cyli:
Another question - what happens if we get an error we're not expecting?  Not an `APIError` we're not expecting, but say something is wrong with our code like `TypeError` or something.  It should retry, right?

radix:
@cyli yes, that's already what happens since steps_to_effect treats any arbitrary exception as a RETRY.

cyli:
@radix ok cool - in this case since we catch errors and report them along with a status, I wanted to know what status to report in case performing the ServiceRequest get resulted in a `TypeError` or something.

otter/convergence/steps.py
```diff
@@ -147,17 +98,24 @@ def report_success(result):
...
+            Otherwise, all other errors are considered non-terminal and a
+            :obj:`StepResult.RETRY` is returned.
             """
             err_type, error, traceback = result
-            if err_type == APIError:
-                message = _parse_nova_user_error(error)
-                if message is not None:
-                    return StepResult.FAILURE, [ErrorReason.String(message)]
 
+            terminal_error = (
```
cyli: Note to reviewer: I plan on refactoring this (and in the `AddNodesToCLB` step) in the next PR.

radix: yay :D

cyli:

otter/cloud_client.py
manishtomar: Is it possible to always pass unicode message to avoid this?

cyli: Oh nope, I'm totally wrong, sorry :(  it's testing identity, so:'

```
>>> Exception('12345') == Exception('12345')
False
```

Whereas using characteristic will compare only the values passed specified in the `@attributes` decorator, ignoring others.  So:

```
>>> @attributes([])
... class MyException(Exception):
...    pass
... 
>>> MyException('message') == MyException('message')
True
>>> MyException('message') == MyException()
True
```

:|  Guess we will need `attr` after all?  Or to specify 'message' as a required attribute?

cyli:
```python
@attr.s
class _BaseExceptionWithMessage(Exception):
     message = attr.ib()
     def __str__(self):
          return self.message

class NovaSomethingSomethingException(_BaseExceptionWithMessage):
     pass
```

Since they all need messages?

manishtomar: Its frustrating that default exception doesn't have equality check. The above `_BaseExceptionWithMessage` is a good idea except that it can be renamed to something like `ExceptionWithEquality`?

cyli: @manishtomar How do you feel about `ExceptionWithMessage`?  So long as `@attr` will used, all of them have equality, the one with message is just so, like all the other builtin exceptions, our exceptions will also take a message in the `__init__`.

manishtomar: The reason i was thinking putting `equality` in name is because anyone reading `exceptionwithmessage` name will wonder why this is so since base exception has message. What do you think? This is probably not that important as a good documentation should do the job :).

cyli:

radix:
Nice change with the exceptions. LGTM! :)

PR #1542, opened 2015-06-25
Don't log BadVersionError when marking clean as an error

radix:
BadVersionError is expected any time multiple convergences are requested, and we don't want to be alerted of this.


cyli:
LGTM, thanks!  This should reduce a lot of the error spam in the logs :D

PR #1543, opened 2015-06-25
Replaced unwrap_wrapped_effect with a new nested_sequence function

radix:
Cons:
- more verbose

Pros:
- a little bit clearer: you don't have to pass an intent *class* and kwargs any more; you're in control of creating the expected intent (meaning you pretty much always have to use mock.ANY yourself)
- more general: doesn't require the wrapper-intent to have an "effect" attribute -- in fact, it can work even work with things like ParallelEffects which have "effects". This is done by accepting a function to of wrapper_intent -> Effect. I'll add a `nested_parallel` function that takes advantage of this in another PR.

I also noticed a few spots in our tests where we weren't really ensuring an effect was being run within the context of another effect, and made them use nested_sequence.

### DIFF
--- a/otter/test/utils.py
+++ b/otter/test/utils.py
@@ -697,7 +697,7 @@ def resolve_stubs(eff):
     return eff_resolve_stubs(base_dispatcher, eff)
 
 
-def perform_sequence(seq, eff, fallback_dispatcher=None):
+def perform_sequence(seq, eff, fallback_dispatcher=base_dispatcher):
     """
     Create a :obj:`SequenceDispatcher` with the given ``seq``, and perform
     ``eff`` with it.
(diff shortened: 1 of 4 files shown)


otter/test/utils.py
```diff
@@ -693,25 +696,56 @@ def resolve_stubs(eff):
...
+        SequenceDispatcher([
+            (BoundFields(effect=mock.ANY, fields={...}),
+             nested_sequence([(SomeIntent(), perform_some_intent)]))
+        ])
+
+    The point is that sometimes you have an intent that wraps another effect,
+    and you want to ensure that the nested effects follow some sequence in the
+    context of that wrapper intent.
+
+    `get_effect` defaults to attrgetter('effect'), so you can override it if
```
cyli: Nice!  I like that this handles ParallelEffects too!

cyli:
LGTM!  Thanks!  This is much cleaner than it was before!

PR #1544, opened 2015-06-25
Also retry updating a node from the trial CLB tests

cyli:
Since we don't want the tests to fail in production if the CLB happened to be in PENDING_UPDATE.

### DIFF
--- a/otter/integration/lib/test_cloud_load_balancer.py
+++ b/otter/integration/lib/test_cloud_load_balancer.py
@@ -22,6 +22,16 @@ class _FakeRCS(object):
     token = "token"
 
 
+pending_update_response = [
+    Response(422),
+    json.dumps({
+        "message": ("Load Balancer '12345' has a status of "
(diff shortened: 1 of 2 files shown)


otter/integration/lib/test_cloud_load_balancer.py
```diff
@@ -59,94 +69,136 @@ def test_list_nodes(self):
...
+        def update(clb, clock):
+            return clb.update_node(self.rcs, 54321, weight=5, clock=clock)
+
+        self.assert_mutate_function_retries_until_success(
+            update, main_treq_args, (Response(202), ""), "")
+
+        self.assert_mutate_function_retries_until_timeout(
+            update, main_treq_args, 60)
+
+        self.assert_mutate_function_does_not_retry_if_not_pending_update(
```
kivattik: Nice use of PENDING_UPDATE vs PENDING_DELETE. :)

kivattik:
:+1: This looks like a great refactor of some things (endpoint) and very useful.

PR #1545, opened 2015-06-25
Read raw passwords from cloudcafe config.

cyli:
Otherwise it tries to interpolate special characters, which might be in passwords.


radix:
LGTM :) I wonder if there are other things we should use get_raw on?

PR #1546, opened 2015-06-26
Nested parallel

radix:
This introduces a `nested_parallel` function which returns a SequenceDispatcher entry (both the intent and the handler), and accepts an intent/handler sequence representing the parallel effects to be performed.

This means our tests can actually ensure that parallel things are run in parallel, whereas right now our tests would pass if we got rid of all of our `parallel()` calls in service.py.

I didn't update all the tests to use it. ExecuteConvergenceTests are going to be a bit tricky and they have some weird cases, and I think they'd be better off refactored more holistically.

### DIFF
--- a/otter/test/utils.py
+++ b/otter/test/utils.py
@@ -748,6 +748,23 @@ def nested_sequence(seq, get_effect=attrgetter('effect'),
         get_effect)
 
 
+def nested_parallel(parallel, fallback_dispatcher=base_dispatcher):
+    """
+    Return a two-tuple for use in a :obj:`SequenceDispatcher` which ensures
+    that all the intents in ``parallel`` are performed in parallel. Note that
+    the items in ``parallel`` must match the order that they're given to the
+    :func:`effect.parallel` function, since the order of inputs affects the
+    order of results.
+
+    :param parallel: sequence of (intent, (intent -> result) function), like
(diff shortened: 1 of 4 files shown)


otter/test/utils.py
```diff
@@ -748,6 +749,24 @@ def nested_sequence(seq, get_effect=attrgetter('effect'),
         get_effect)
 
 
+def nested_parallel(parallel, fallback_dispatcher=base_dispatcher):
```
manishtomar: `fallback_dispatcher` is not used in the function. Am I missing something?

radix: nope, I forgot to pass it through :) I guess that means this deserves some unit tests... :P

radix: Ok, this is used now, and I've added tests. thanks for the comment :)

cyli:
Non-blocking questions:

1.  Is the plan for moving `nested_sequence` and `nested_parallel` to effect at some point?  (I'm just curious)
1.  Would it make sense to have a negative test that `nested_parallel` would fail if you provided a sequence that was not in parallel?

Other than that, LGTM!

radix:
@cyli 
1. yes, very likely :)
2. yeah, I'll add one, thanks for the suggestion :)

cyli:
LGTM!

PR #1547, opened 2015-06-26
Fix NoSuchScalingGroup when deleting group

cyli:
Because the group has a status of DELETING, `view_state` filters the groups out.  However, `delete_group` calls `view_state` to determine if the group really exists before deleting it, and to determine whether the group has servers in it.

Fixed so that if the group is in DELETING state, `view_state` will not filter it out.  And if there are still servers in the DELETING group (because convergence won't bother to update the active cache before deleting the group), do not refuse to delete.

Have verified on the dev vm that there are no more deleting groups remaining in the DB, and that after running the trial tests the servers are still cleaned up from mimic.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -153,11 +153,6 @@ _cql_insert_group_state = (
     '"policyTouched", paused, desired) VALUES(:tenantId, :groupId, :active, '
     ':pending, :groupTouched, :policyTouched, :paused, :desired) '
     'USING TIMESTAMP :ts')
-_cql_view_group_state = (
-    'SELECT "tenantId", "groupId", group_config, active, pending, '
-    '"groupTouched", "policyTouched", paused, desired, created_at, status '
-    'FROM {cf} '
-    'WHERE "tenantId"=:tenantId AND "groupId"=:groupId AND deleting=false;')
 
 # --- Event related queries
 _cql_insert_group_event = (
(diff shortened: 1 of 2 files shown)


cyli:

otter/test/models/test_cass_models.py
```diff
@@ -528,28 +546,20 @@ def test_view_state(self):
...
-             'active': '{"A":"R"}',
-             'pending': '{"P":"R"}',
-             'groupTouched': '2014-01-01T00:00:05Z.1234',
-             'policyTouched': '{"PT":"R"}',
-             'paused': False,
-             'created_at': 23,
-             'desired': 10,
-             'status': 'ACTIVE'}]
-        self.returns = [cass_response]
+        self.returns = [[
```
manishtomar: Nice refactor. Thanks!

otter/models/cass.py
```diff
@@ -805,7 +800,12 @@ def view_state(self, consistency=None):
                           NoSuchScalingGroupError(self.tenant_id, self.uuid),
                           self.log)
 
-        return d.addCallback(_unmarshal_state)
+        def check_deleting(group):
```
manishtomar: This function is also in `view_manifest`. Do you think you can extract it out and use it in both places?

cyli: Good point, thanks!  Don't know why I didn't do that.

otter/models/cass.py
```diff
@@ -1292,7 +1292,8 @@ def _delete_everything(ts, webhooks):
             return b.execute(self.connection)
 
         def _maybe_delete(state):
-            if len(state.active) + len(state.pending) > 0:
+            if (state.status != ScalingGroupStatus.DELETING and
```
manishtomar: Nice check

radix:
Nice changes! LGTM. But I'll wait until @manishtomar finished too :)

manishtomar:
LGTM :)

cyli:

PR #1548, opened 2015-06-26
Load balancer type validation fix

cyli:
We weren't really ensuring that random types couldn't be passed in as a load balancer type, because our unit tests were passing since the example given in the unit test would always fail because the server args were invalid, not because the lb args were invalid.

Fix that, and the schema validation.  Also add a functional test for this.

### DIFF
--- a/otter/json_schema/group_schemas.py
+++ b/otter/json_schema/group_schemas.py
@@ -117,7 +117,7 @@ _rcv3_lb = {
             "description": (
                 "What type of a load balancer is in use"),
             "required": True,
-            "pattern": "^\s*RackConnectV3\s*$"
+            "pattern": "^RackConnectV3$"
         }
     },
     "additionalProperties": False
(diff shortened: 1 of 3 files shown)


cyli:

otter/json_schema/group_schemas.py
```diff
@@ -151,7 +151,7 @@
             "description": (
                 "What type of a load balancer is in use"),
             "required": False,
-            "oneOf": ["CloudLoadBalancer"]
+            "pattern": "^CloudLoadBalancer$"
```
manishtomar: Just curious (non-blocking): What was wrong with `oneOf`?

manishtomar:
LGTM!

otter/json_schema/group_schemas.py
cyli: It didn't work :|  Seems like the values themselves have to be schemas?  http://json-schema.org/latest/json-schema-validation.html#anchor88

cyli:
Thanks!  (Just going to wait for the response from reach first before merging)

PR #1549, opened 2015-06-26
Gathering cached servers

manishtomar:
Final work of #1458 and partial work of #1558. This is dependent on #1567. Please review and merge that first.

Here, group state and servers cache both are updated with Autoscale active servers. The next PR will change get group state REST implementation to get it from new servers cache instead of group state cache at which time updating group state cache will be removed. 

Cleaned up `exec_convergence` and servers cache tests by using `nested_parallel` and some minor factoring.


manishtomar:

otter/test/utils.py
```diff
@@ -846,6 +846,34 @@ def noop(_):
...
 
+def intent_func(fname):
+    """
+    Return function that returns Effect of tuple of fname and its args. Useful
+    in writing tests that expect intent based on args
+    """
+    return lambda *a: Effect((fname,) + a)
+
+
+class Cache(object):
```
radix: this class should probably have a more specific name than Cache. and maybe an `@implements`

otter/convergence/gathering.py
```diff
@@ -112,8 +116,71 @@ def group_id(s):
...
+def get_scaling_group_servers(tenant_id, group_id, now,
+                              cache_class=CassScalingGroupServersCache,
+                              all_as_servers=get_all_scaling_group_servers,
+                              all_servers=get_all_server_details):
+    """
+    Get a group's servers taken from cache if it exists. Updates cache
+    if it is empty from newly fetched servers
+    # NOTE: This function takes tenant_id even though the whole effect is
+    # scoped on the tenant because cache calls require tenant_id. Should
+    # they also not take tenant_id and work on the scope?
```
radix: We could perhaps treat TenantScope more like a reader monad, and create a "GetTenantID" effect.

Which then really begs for a general DynamicScope effect with a GetVariable effect. Oh man, I feel dirty. :) And it means that Effect could implement the vision of twisted.python.context in a coherent way! I would feel a lot less dirty about this idea if we had a type system that could ensure all dynamic accesses were valid. 

We could still do this without completely generalizing to dynamic scoping by making TenantScope provide a performer for an `UpdateServersCache` intent, just  like it does for `ServiceRequest`.

Anyway, this is just some bikeshedding. I don't think you need to change this code.

otter/convergence/gathering.py
```diff
@@ -112,8 +116,71 @@ def group_id(s):
...
+        last_update = now - timedelta(days=30)
+        changes, current = yield parallel([
+            all_as_servers(last_update).on(partial(group_servers, group_id)),
+            all_as_servers().on(partial(group_servers, group_id))])
+        servers = merge_servers(changes, current)
+        yield cache.insert_servers(now, servers, True)
+    else:
+        changes = yield all_servers(last_update)
+        servers = merge_servers(cached_servers, changes)
+        servers = list(filter(partial(server_of_group, group_id), servers))
```
radix: I can't really remember what the conclusion was about why we're not updating the servers cache in this case either, so perhaps you could add a code comment here explaining why?

otter/convergence/service.py
```diff
@@ -183,6 +185,13 @@ def execute_convergence(tenant_id, group_id, build_timeout,
                                            status=ScalingGroupStatus.ACTIVE))
             yield cf_msg('group-status-active',
                          status=ScalingGroupStatus.ACTIVE.name)
+        else:
+            # Clear servers cache and update it with latest servers
```
radix: Related to my previous comment in `get_scaling_group_servers`, I'm not sure why we're waiting until here to update the servers cache, instead of doing it in the gathering process, since these servers we're saving here are exactly the ones we gathered.

otter/convergence/service.py
```diff
@@ -174,6 +175,7 @@ def execute_convergence(tenant_id, group_id, build_timeout,
 
     worst_status, reasons = yield _execute_steps(steps)
 
+    cache = cache_class(tenant_id, group_id)
```
radix: This seemed weird to me, and for a while I couldn't figure out why. I think I figured it out: I think we should have an UpdateServersCache intent, just like the other intents in otter.models.intents like ModifyGroupState.

- we'd be able to get rid of this parameterized `cache_class`
- we'd be able to get rid of the fake `Cache` class entirely

What do you think?

otter/test/convergence/test_gathering.py
```diff
@@ -213,85 +221,96 @@ class GetScalingGroupServersTests(SynchronousTestCase):
...
+
+    def test_no_cache(self):
+        """
+        If cache is empty then current list of servers are returned and added
+        to the cache
+        """
+        self._test_no_cache(False)
+        self._test_no_cache(True)
+
+    def _test_old_case(self, last_update, as_srvs=True, cur_srvs=True):
```
radix: Is this meant to be called _test_old_cache?

radix:
Thank you so much for improving so many unit tests! you rock! :)

manishtomar:

otter/test/convergence/test_gathering.py
manishtomar: I intended it to be `_test_old_case` but I guess `_test_old_cache` would make more sense? I can change it if needed :)

otter/test/convergence/test_gathering.py
```diff
@@ -213,85 +221,87 @@ class GetScalingGroupServersTests(SynchronousTestCase):
...
+    def test_no_cache(self):
+        """
+        If cache is empty then current list of servers are returned
+        """
+        self._test_no_cache(False)
+        self._test_no_cache(True)
+
+    def _test_old_case(self, last_update, as_srvs=True, cur_srvs=True):
+        exp_last_update = datetime(2010, 5, 1)
+        as_servers = self.servers1
```
cyli: There is no overlap between `as-servers` and `current` - would it make sense to add another server to `current` that has the same ID as `as-servers`, so that we can test that current servers take precedence over the changes-since servers?  If the order of the servers passed to `merge_servers` for the older than 30-day case was reversed, this test wouldn't catch it.

otter/test/convergence/test_gathering.py
cyli: I don't particularly have an opinion, but I think they may have asked because `test_no_cache` uses the private function `_test_no_cache`, and this is being used by the test `test_old_cache`, but it's named `_test_old_case`

otter/test/convergence/test_gathering.py
```diff
@@ -213,85 +221,87 @@ class GetScalingGroupServersTests(SynchronousTestCase):
     """
 
     def setUp(self):
-        """Save basic reused data."""
```
cyli: I'm not sure we should throw away the tests for `get_all_scaling_group_servers` - it's invoked by `get_scaling_group_servers`, which you now have tests for, but `get_all_scaling_group_servers` is stubbed out in those tests.

If the functionality for actually filtering stuff out by group, etc. in `get_all_scaling_group_servers` is broken, I'm not sure the unit tests would catch them now, and then `get_scaling_group_servers` would break because it assumes that `get_all_scaling_group_servers` would do it correctly.

otter/test/convergence/test_gathering.py
```diff
@@ -576,8 +586,8 @@ def no_endpoint(intent):
             sync_perform(dispatcher, get_rcv3_contents()), [])
 
 
-def _constant_as_eff(constant):
-    return lambda: Effect(Stub(Constant(constant)))
+def _constant_as_eff(args, retval):
```
cyli: Nice!  Glad we verify args now!

otter/models/intents.py
```diff
@@ -1,29 +1,38 @@
...
 from txeffect import deferred_performer
 
+from otter.models.cass import CassScalingGroupServersCache
+from otter.util.fp import assoc_obj
+
 
-@attributes(['scaling_group', 'modifier'])
-class ModifyGroupState(object):
+@attr.s
+class ModifyGroupStateActive(object):
```
cyli: I was wondering what the reason was for this change to `ModifyGroupState`?  I thought part of the point of putting the active boolean in the server cache was to eliminate writing to the state (and hence eliminate one more get a lock)?  Just modifying the active column also grabs a lock, so this doesn't seem to reduce lock contention.

radix: Well, it would make tests a lot easier, since they no longer need to deal with a callable, instead just doing an equality check on `active`, so that's nice!

cyli: True, it would make existing tests easier, but I thought we were going to try to eliminate the need for this intent entirely?

manishtomar: That will be done in next PR which will get group state active from servers cache in rest layer. Here, I put it both places to have working master

cyli: Ok - I just wanted to verify that that was still happening, and I didn't miss something.

manishtomar: Sure. Glad you asked. I had updated the PR description above with this information. I am sorry I should've mentioned that on slack.

otter/test/convergence/test_gathering.py
manishtomar: Oh. I am sorry I removed it. I don't know why I did that. Thanks so much for catching this. I'll add the tests. I guess the function name change caused the confusion.

otter/test/convergence/test_gathering.py
manishtomar: Good idea. Will add it.

otter/test/convergence/test_service.py
```diff
@@ -790,161 +772,149 @@ def plan(*args, **kwargs):
...
 
-        eff = execute_convergence(self.tenant_id, self.group_id,
-                                  build_timeout=3600,
-                                  get_all_convergence_data=gacd, plan=fplan)
-        disp = self._get_dispatcher([(self.gsgi, (self.group, self.manifest))])
-        # This succeeded without DeleteGroup performer being there ensuring
-        # that it was not called
-        self.assertEqual(sync_perform(disp, eff),
-                         (StepResult.RETRY, ScalingGroupStatus.DELETING))
+    def test_deleting_group_retry(self):
```
cyli: +1 splitting this into two tests

cyli:
The unit tests for service.py are so much cleaner, thank you!

cyli:

manishtomar:

cyli:

PR #1552, opened 2015-06-29
reduce fruitless gathering and spurious NoSuchGroupErrors after group deletion

radix:
Pretty simple change: just always delete divergent flag (by passing version=-1) a group is deleted.

I had to change execute_convergence's return value to include the group status so converge_one_group could know to do this.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -140,8 +140,8 @@ def execute_convergence(tenant_id, group_id, build_timeout,
         :func`get_all_convergence_data`, used for testing.
     :param callable plan: like :func:`plan`, to be used for test injection only
 
-    :return: Effect of two-tuple of (most severe StepResult, group status). If
-        group status is DELETING it's safe to assume the group has been
+    :return: Effect of two-tuple of (most severe StepResult, group status).
+        When group status is None it means the group has been successfully
         deleted.
     :raise: :obj:`NoSuchScalingGroupError` if the group doesn't exist.
     """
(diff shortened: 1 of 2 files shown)


radix:

otter/convergence/service.py
```diff
@@ -389,7 +394,14 @@ def converge_one_group(currently_converging, tenant_id, group_id, version,
...
         yield err(None, 'converge-non-fatal-error')
     else:
-        if result in (StepResult.FAILURE, StepResult.SUCCESS):
+        if result[0] in (StepResult.FAILURE, StepResult.SUCCESS):
+            # In order to avoid doing extra work and reporting spurious errors,
+            # if the group status is DELETING it means the group has
+            # successfully been deleted by execute_convergence. And so we will
+            # unconditionally delete the divergent flag to avoid any further
+            # queued-up convergences that will imminently fail.
+            if result[1] == ScalingGroupStatus.DELETING:
```
cyli: Non-blocking:  Is there a case where the group is in deleting but StepResult ends in FAILURE?  If so, how can we recover from this?  Nothing is going to trigger convergence on the group again.  Should we just delete the group anyway?

radix: To answer your first question, it is indeed possible

and thinking about this, I think this whole change is invalid. I realized that we may require multiple convergence iterations to actually delete a group, and this change I made just causes it to delete the divergent flag after the first iteration unconditionally.

We're going to need to distinguish DELETING from "actually deleted".

radix:

cyli:
This LGTM - it doesn't eliminate the NoSuchScalingGroup failures, since as @manishtomar points out if there is an error while converging a deleting group, it tries to write the FAILURE state to the group, and can't because the group is deleting.  But this does eliminate the ones that occur due to re-converging after success.

manishtomar:
LGTM :)

radix:
ok, I'll merge this up then. Thanks for filing #1553 Ying :)

PR #1554, opened 2015-06-29
Intermittent cafe failure when scaling up over 25

cyli:
Against mimic.

The test `...system.policies.test_system_execute_updated_policies.ExecuteUpdatedPoliciesTest.test_update_policy_desired_capacity_over_25` sometimes fails after 30 seconds with the error that there are 26 servers as the desired capacity, but only 21 servers on the mimic account.  Likely this is due to throttling - we should probably increase the timeout on this test specifically.


radix:
good to see throttling is working I guess :)

autoscale_cloudroast/test_repo/autoscale/system/policies/test_system_execute_updated_policies.py
```diff
@@ -122,7 +122,7 @@ def test_update_policy_desired_capacity_over_25(self):
             .format(upd_policy_to_desired_capacity_execute, self.group.id))
         self.check_for_expected_number_of_building_servers(
             group_id=self.group.id,
-            expected_servers=26)
+            expected_servers=26, time_scale=False)
```
radix: I guess a comment here would be nice, explaining why we're disabling time scale on this mimic-using test

PR #1556, opened 2015-06-29
skip deleting groups when collecting metrics

manishtomar:


cyli:
LGTM...  although non-blocking question - do we want to gather any metrics on how long it takes to delete a group?

manishtomar:
Maybe. We should be able to extract it from logs right? Like you said that's outside scope of this PR.

cyli:
Oh that's true.  :)

PR #1560, opened 2015-06-30
First take at LBSH-002

kivattik:
There are quite a few code optimizations I would like to do here eventually, but this should be ok for now.

### DIFF
--- a/otter/integration/tests/test_lbsh.py
+++ b/otter/integration/tests/test_lbsh.py
@@ -13,6 +13,7 @@ from twisted.internet.defer import gatherResults, inlineCallbacks
 from twisted.trial import unittest
 
 from otter.integration.lib.cloud_load_balancer import (
+    CloudLoadBalancer,
     ContainsAllIPs,
     HasLength
 )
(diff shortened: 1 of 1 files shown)


otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
...
+
+        Server node moved from LB1 to LB2
+        Assert: Server put back on LB1
+        Assert: Server removed from LB2
+        """
+        # Create another loadbalancer not to be used in autoscale
+        # The CLB will not be added to the helper, since when the helper
+        # creates a group, it automatically adds the clb
+        clb_other = CloudLoadBalancer(pool=self.helper.pool)
+        clb_other.start(self.rcs, self)
```
cyli: This seems to be called twice - once with yield, and here without.  Probably this line can be removed?

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
...
+            "There should be 1 node on the autoscale CLB.")
+        self.assertEqual(
+            len(nodes_other['nodes']), 0,
+            "There should still be 0 nodes on the out of band CLB")
+
+        the_node = nodes_as["nodes"][0]
+        node_info = {
+            "address": the_node["address"],
+            "port": the_node["port"],
+            "weight": 2
```
cyli: I think this might not work in production?  I think condition is required in addition to port and address (although it doesn't seem to be documented.  :| But I can't add a node without it)

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
...
+            ),
+            timeout=timeout_default
+        )
+
+        yield clb_other.wait_for_nodes(
+            self.rcs,
+            MatchesAll(
+                HasLength(0)
+            ),
+            timeout=timeout_default
```
cyli: Ooh very nice :D

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
             ),
             timeout=timeout_default
         )
+
+    @tag("LBSH")
```
cyli: Did we want to tag this LBSH-002 (and the other one LBSH-001)?

otter/integration/tests/test_lbsh.py
kivattik: Good to know, I'll add that.

otter/integration/tests/test_lbsh.py
kivattik: Yes, yes we did.

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
...
+            MatchesAll(
+                HasLength(1),
+                ContainsAllIPs([the_node["address"]])
+            ),
+            timeout=timeout_default
+        )
+
+        yield clb_other.wait_for_nodes(
+            self.rcs,
+            MatchesAll(
```
radix: Is there a reason to use `MatchesAll` with only one argument? I think this can just be replaced with `HasLength(0)` without the `MatchesAll`

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,82 @@ def test_oob_deleted_clb_node(self):
...
+                         "There should be no nodes on the CLB yet.")
+        self.assertEqual(len(nodes_other['nodes']), 0,
+                         "There should be no nodes on the CLB yet.")
+
+        group, _ = self.helper.create_group(min_entities=1)
+        yield self.helper.start_group_and_wait(group, self.rcs)
+
+        nodes_as = yield clb_as.list_nodes(self.rcs)
+        nodes_other = yield clb_other.list_nodes(self.rcs)
+        self.assertEqual(
```
cyli: Non-blocking - I was running my tests against prod yesterday and my tests seemed to be failing on these assertions.  I'm guessing that it takes a sec for CLB to update the listing?  So I changed it to wait_for_nodes to reach a certain number, and that seemed to fix it.

I was planning on going through and updating the other test anyway, so if you don't want to do that in this PR, I can just update them both later.  :)

kivattik: I agree with radix's above comment, so I'll just fix this now.

otter/integration/tests/test_lbsh.py
kivattik: Good point. The reason would be copy pasta.

otter/integration/tests/test_lbsh.py
cyli: I think this one also needs fixing in addition to the ones there (probably the initial assertion that there are zero to start with is fine).

otter/integration/tests/test_lbsh.py
```diff
@@ -97,3 +98,80 @@ def test_oob_deleted_clb_node(self):
...
+            "condition": the_node["condition"],
+            "weight": 2
+        }
+
+        yield clb_as.delete_nodes(self.rcs, [the_node['id']])
+        yield clb_other.add_nodes(self.rcs, [node_info])
+
+        nodes_as = yield clb_as.list_nodes(self.rcs)
+        nodes_other = yield clb_other.list_nodes(self.rcs)
+        self.assertEqual(clb_as.wait_for_nodes(
```
cyli: This should probably just be `clb_as.wait_for_nodes(self.rcs, HasLength(0), timeout=timeout_default)`, since that would error anyway if there aren't the requisite number of nodes.

cyli:

cyli:
LGTM, aside from that one change, which we can finish in another PR :)

PR #1563, opened 2015-07-01
Create server in trial tests

cyli:
Add an API for creating servers directly with Nova for the LBSH tests.

Also change the `wait_for_servers` function to be able to match against all servers on the account, not just the autoscaled servers.  That way it can be used to wait for particular non-autoscale servers to become active.

### DIFF
--- a/otter/integration/lib/nova.py
+++ b/otter/integration/lib/nova.py
@@ -136,10 +136,10 @@ def create_server(rcs, pool, server_args, _treq=treq):
     """
     d = _treq.post(
         "{0}/servers".format(rcs.endpoints['nova']),
-        server_args,
+        json.dumps(server_args),
         headers=headers(str(rcs.token)),
         pool=pool
-    ).addCallback(check_success, [200]).addCallback(_treq.json_content)
+    ).addCallback(check_success, [202]).addCallback(_treq.json_content)
     return d.addCallback(itemgetter('server')).addCallback(itemgetter('id'))
 
 
--- a/otter/integration/lib/nova.py
+++ b/otter/integration/lib/nova.py
@@ -1,6 +1,8 @@
 """Contains reusable classes relating to nova."""
 import json
 
+from operator import itemgetter
+
 from characteristic import Attribute, attributes
 
 import treq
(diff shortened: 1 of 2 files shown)


cyli:

otter/integration/lib/nova.py
```diff
@@ -120,35 +122,62 @@ def list_servers(rcs, pool, _treq=treq):
...
+
+    :return: the server ID of the created server.
+    """
+    d = _treq.post(
+        "{0}/servers".format(rcs.endpoints['nova']),
+        json.dumps(server_args),
+        headers=headers(str(rcs.token)),
+        pool=pool
+    ).addCallback(check_success, [202]).addCallback(_treq.json_content)
+    return d.addCallback(itemgetter('server')).addCallback(itemgetter('id'))
```
kivattik: I like this means of getting a value from the response. :)

otter/integration/lib/test_nova.py
```diff
@@ -150,20 +150,33 @@ def setUp(self):
...
         d = nova.list_servers(self.rcs, pool=self.pool,
                               _treq=treq)
         self.assertEqual({'servers': {}}, self.successResultOf(d))
 
+    def test_create_server(self):
+        """
+        Create a server with particular args and succeeds on 202.
+        """
+        treq = get_fake_treq(self, 'post', 'novaurl/servers',
+                             (('{"server": {}}',), self.expected_kwargs),
```
kivattik: The args passed in is empty, but the created server has an id of 12345, which doesn't seem right.

otter/integration/lib/test_nova.py
```diff
@@ -202,8 +215,8 @@ def test_wait_for_servers_retries_until_matcher_matches(self):
         """
         If the matcher does not matches the nova state, retries until it does.
```
kivattik: Already in existence and non-blocking, s/matches/match

cyli: Good catch, thanks!

kivattik:
:+1:  With ninja edits :)

cyli:
Thanks @kivattik!

PR #1564, opened 2015-07-01
LBSH-004 test to ensure Autoscale doesn't touch non-autoscaled servers

cyli:
~~This depends on #1560.~~

~~Tested the merged PRs against prod and mimic.~~

Dependent PRs have been merged.


cyli:

otter/integration/tests/test_lbsh.py
```diff
@@ -34,6 +38,7 @@ class TestLoadBalancerSelfHealing(unittest.TestCase):
     This class contains test cases to test the load balancer healing
     function of the Otter Converger.
     """
+    timeout = 1800
```
kivattik: Is this used anywhere? It looks to me like everything is using timeout_default, though I could be missing something.

cyli: This is used to time out the a test specifically.  If any of the test methods in this class runs for more than 1800 seconds, trial will kill it.  Otherwise the default is something much shorter (which works against mimic, but not against production).

Everywhere is to time out polling for stuff, like a specific `wait_for_X` operation that we wrote ourselves.  I figured since a test includes multiple `wait_for_X` with the default timeouts, then the test timeout should probably be a little longer.

otter/integration/tests/test_lbsh.py
```diff
@@ -173,3 +178,80 @@ def test_move_node_to_oob_lb(self):
...
+
+        # create the other two non-autoscaled servers - just wait until they
+        # have servicenet addresses - don't bother waiting for them to be
+        # active, which will take too long
+        other_servers = yield self.helper.create_servers(
+            self.rcs, 2, wait_for=ContainsDict({
+                "addresses": ContainsDict({
+                    'private': MatchesSetwise(
+                        ContainsDict({
+                            "addr": MatchesRegex("(\d+\.){3}\d+")
```
kivattik: Nice. I'm learning cool things about matchers. :)

kivattik:
:+1: LGTM. I'd forgotten about the configureable test case timeout.

manishtomar:
Sorry for chiming in late, but I am really not seeing the point of 004 labeling system mainly because I am not sure if there is an authoritative place where these numbers are defined. AFAIK, they exist only as github issue title name which can be replaced by the issue number and linked in the code. At least that way it is easily linked. I think that is better than separate numbering system even if the issue gets closed. Moreover, the test name and docstrings in the test are really nice and the number just confuses me. I don't know what to understand of the PR when I see "Fixed LBSH 005" in the title. Sorry for the rant.

cyli:
@manishtomar They are linked in the test plan, and I think they're mainly to ensure accurate QE reporting of test case completion based on a test plan document.  This PR had both the number and a descriptive title and a link to the issue describing it - I can copy the issue description here if you like.

manishtomar:
@cyli Thanks :) And sorry for the passive aggressive comment above. I think have asked this question earlier too. I promise not to ask again. Great test though :) Again, sorry for the comment above :(

cyli:
@manishtomar No, not passive aggressive!  You make a good point in that the issues can also be linked to the test document instead of the name.  I'm not entirely sure what the requirements are actually, just that it was used for tracking against that document - and in the previous case (all the other integration tests), to determine which tests to copy over for the >1 CLB case.

@kivattik - would it be sufficient to just link the created issues to the test plan document in the future?

PR #1565, opened 2015-07-01
Steps reporting refactor

cyli:
This does 2 things:

1. Refactor some common error reporting patterns in the steps code
1. Fix a bug where if a TypeError or some other error occurs in our cloud client code, it should be a RETRY rather than a FAILURE.

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -17,11 +17,11 @@ from twisted.python.constants import NamedConstant
 from zope.interface import Interface, implementer
 
 from otter.cloud_client import (
-    CLBDuplicateNodesError,
-    CLBPendingUpdateError,
-    CLBRateLimitError,
+    CLBDeletedError,
+    CLBNodeLimitError,
     CreateServerConfigurationError,
     CreateServerOverQuoteError,
+    NoSuchCLBError,
     add_clb_nodes,
     create_server,
     has_code,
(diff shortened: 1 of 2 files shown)


cyli:

otter/convergence/steps.py
```diff
@@ -70,6 +70,46 @@ def set_server_name(server_config_args, name_suffix):
...
+        )
+
+        if terminal_error:
+            return StepResult.FAILURE, [ErrorReason.Exception(exc_tuple)]
+        return StepResult.RETRY, [ErrorReason.Exception(exc_tuple)]
+
+    return reporter
+
+
+def _success_reporter(success_reason):
```
radix: I know this is because of a rather poor design we currently have in steps (and I take the blame for that), but it seems weird have something called `_success_reporter` be a thing that returns RETRY. I think it'd be preferable to call it `_retry_reporter`, though at some point in the future it'd be nice if we could get rid of the necessity of having most of the steps return RETRY.

radix: but this is just a really minor nitpick, (and perhaps there's benefit to calling it `_success_reporter` as an *abstraction*) so feel free to ignore.

radix:
beautiful refactor! LGTM :)

otter/convergence/steps.py
cyli: Ah ok, I was interpreting it as the cloud client succeeded, and how to report that success, same with `_failure_reporter`, since it also can return RETRY as well as FAILURE.

cyli: Alternately, name them `_handle_success` vs `_handle_failure`?

radix: I don't think you need to clarify anything. I've changed my mind :) hopefully we'll just be able to update this function at some point to return SUCCESS.

cyli: Ok, well please change it in your upcoming PR if you find something better!  :D

PR #1566, opened 2015-07-01
metrics service is optional

manishtomar:
metrics service uses code from otter but otter does not need to depend on it. This way metrics can be separate role in https://github.com/rackerlabs/autoscaling-chef/pull/745


cyli:
LGTM once travis +1's :)

Issue #1558, opened 2015-06-30
Using the server cache for the active servers returned by state


cyli:
The Autoscale active server cache (AASC) != the Nova server cache (NSC), because the autoscale active servers are on the CLBs they should be on, whereas the server cache would only give us servers that Nova considers active.

We'd like to be able to synthesize the AASC from the NSC, partially because it's the same information in two places, partially to reduce lock contention.

But the NSC doesn't have any information about CLB, so this issue is to serve as a sort of place to discuss ways we can get around this, whether we should do this, etc.

Possibilities:

1.  Just re-gather from Nova and CLB every time the user queries state.  This seems expensive for Nova and CLB though - people poll Autoscale, and we in turn would be polling Nova + CLB every time we were polled, as well as every time we converge.  This might overload Nova, not to mention that it may make our GET request responses very slow.

2.  Gather active servers from nova from the NSC, and just query CLB every time the state is is requested - we don't actually care about the entire state of the CLB world, so we could only get the load balancers that we cared about, making it less expensive than a full gather.  Not sure how CLB will feel about this.

3.  Instead of writing to the state when we want to update the AASC, we write a boolean to the NSC stating whether autoscale considers the server ID fully active.  That way, we can query the NSC to synthesize the AASC by filtering out all servers that don't have the boolean set to True.  We don't need a lock, because one api node can write to the NSC at a time due to partitioning, and if that fails, eh... it's a cache.  ~~The problem is we have to figure out a way to ship the last updated timestamp around.~~ 

    If we write the cache at every gathering, the we have the last updated time in `execute_convergence`, which passes that time to the gathering and can also pass it to `_update_active`.

radix:
I'm happy with #3, and I think we should move the cache-updating out of gathering.py and into execute_convergence, unifying all of our cache-updating to one place.

cyli:
Can this be closed now?

manishtomar:
Oh yes. Closing as fixed in #1567.

PR #1567, opened 2015-07-01
Server cache model change

manishtomar:
Change needed for #1558

- Changed model and added API to store `server_as_active` boolean along with server which will be
set to true when the server is known to have become AS active, i.e. been added in all LBs in the group
- Defaulted consistency level to `DEFAULT_CONSISTENCY` in `cql_eff`

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -1763,6 +1763,18 @@ class CassScalingGroupServersCache(object):
         else:
             return cql_eff(batch(queries), params)
 
+    def update_server_in_lbs(self, last_update, server_id, server_in_lbs):
+        """
+        See :method:`IScalingGroupServersCache.update_server_in_lbs
+        """
+        query = ('UPDATE {cf} SET server_in_lbs=:server_in_lbs '
+                 'WHERE "tenantId"=:tenantId AND "groupId"=:groupId '
+                 'AND last_update=:last_update AND server_id=:server_id')
+        params = merge(
+            self.params, {"last_update": last_update, "server_id": server_id,
+                          "server_in_lbs": server_in_lbs})
+        return cql_eff(query.format(cf=self.table), params)
+
     def delete_servers(self):
         """
         See :method:`IScalingGroupServersCache.delete_servers`
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -59,6 +59,10 @@ from otter.util.retry import repeating_interval, retry, retry_times
 
 LOCK_PATH = '/locks'
 
+DEFAULT_CONSISTENCY = ConsistencyLevel.QUORUM
+
+QUERY_LIMIT = 10000
+
 
 @attributes(['query', 'params', 'consistency_level'])
 class CQLQueryExecute(object):
@@ -87,7 +91,7 @@ def get_cql_dispatcher(connection):
     })
 
 
-def cql_eff(query, params={}, consistency_level=ConsistencyLevel.ONE):
+def cql_eff(query, params={}, consistency_level=DEFAULT_CONSISTENCY):
     """
     Return Effect of CQLQueryExecute intent
     """
@@ -305,11 +309,6 @@ def _paginated_list(tenant_id, group_id=None, policy_id=None, limit=100,
 
     cql_parts.append(" LIMIT :limit;")
     return (''.join(cql_parts), params)
-
-
-DEFAULT_CONSISTENCY = ConsistencyLevel.QUORUM
-
-QUERY_LIMIT = 10000
 
 
 def _build_policies(policies, policies_table, event_table, queries, data,
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -1763,17 +1763,21 @@ class CassScalingGroupServersCache(object):
         else:
             return cql_eff(batch(queries), params)
 
-    def update_server_in_lbs(self, last_update, server_id, server_in_lbs):
+    def set_servers_as_active(self, last_update, server_ids):
         """
-        See :method:`IScalingGroupServersCache.update_server_in_lbs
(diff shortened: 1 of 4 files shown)


cyli:
This LGTM for the stated purpose of the PR, but I thought @radix suggested that rather than update the active column, we write the cache with an active boolean at the same time, rather than in two passes?  In which case the update function may not be necessary...

Also do we need an alter table schema to add that column?

manishtomar:
Hmm.. I remember him asking about column mutation but didn't realize this is what he had suggested. Now that I think carefully, I remember him talking about moving cache insertion to gathering. Cha.. waste of effort :( I'll make that change. 

> Also do we need an alter table schema to add that column?

We probably do but this is not yet used in prod and we are not running migrations in CI anyways. So, left it out. I'll be happy to do it though :)

manishtomar:
I am not sure if @radix suggestion is feasible. We need deleted servers in the cache until convergence succeeds and `exec_convergence` will insert only active servers. I could change to insert all servers (including deleted servers) but have `server_as_active` field set for certain servers. I'll need to change `insert_servers` API for that to take `server_as_active` field either within the server dict or as a separate list. What do you think?

cyli:
@manishtomar:  probably the `insert_servers` API change would make sense - and as for within the server dict or as a separate list, I guess whichever is easier.  :)  Conceptually I think it makes sense to have it in the server dict, but maybe that would be harder?

otter/test/models/test_cass_models.py
```diff
@@ -3708,36 +3710,53 @@ def setUp(self):
...
+        `get_servers` fetches only AS active servers that have highest
+        last_fetch time
+        """
+        self._test_get_servers(
+            True,
+            [{"server_blob": '{"a": "b"}', "last_update": self.dt,
+              "server_as_active": True},
+             {"server_blob": '{"d": "e"}', "last_update": self.dt,
+              "server_as_active": False}],
+            ([{"a": "b"}], self.dt))
```
cyli: Non-blocking - do we want to add an extra cache blob with an earlier update time here?  The test case `test_get_servers_diff_last_udpate` tests it for where `only_as_active` is False, and it might be nice to include a test for when it's True.

otter/test/models/test_cass_models.py
```diff
@@ -3708,36 +3710,53 @@ def setUp(self):
...
         """
         `get_servers` fetches all servers that have highest last_fetch
         time
         """
         self._test_get_servers(
-            [{"server_blob": '{"a": "b"}', "last_update": self.dt},
-             {"server_blob": '{"d": "e"}', "last_update": self.dt}],
+            False,
+            [{"server_blob": '{"a": "b"}', "last_update": self.dt,
+              "server_as_active": False},
```
cyli: Non-blocking - do we want to make one of these True (same with `test_get_servers_diff_last_update`) as a sanity check/regression test to make sure we're not filtering out all True's also?

otter/models/cass.py
```diff
@@ -1739,8 +1740,13 @@ def get_servers(self):
             yield do_return(([], None))
         last_update = rows[0]['last_update']
         rows = takewhile(lambda r: r['last_update'] == last_update, rows)
-        yield do_return(([json.loads(r['server_blob']) for r in rows],
-                         last_update))
+
+        def _dict(r): return json.loads(r['server_blob'])
+        rfunc = (
+            compose(map(_dict), filter(lambda r: r['server_as_active']))
```
cyli: Yay!  Nice composition :D

cyli:

cyli:
LGTM aside from some test nitpicks, which I leave up to you whether we want to do them here.

otter/test/models/test_cass_models.py
```diff
@@ -3745,30 +3766,43 @@ def test_get_servers_diff_last_update(self):
...
+        # Test with only_as_active as True
+        self._test_get_servers(
+            True,
+            [{"server_blob": '{"a": "b"}', "last_update": dt_earlier,
+              "server_as_active": False},
+             {"server_blob": '{"d": "e"}', "last_update": dt_earlier,
+              "server_as_active": True},
+             {"server_blob": '{"c": "f"}', "last_update": self.dt,
+              "server_as_active": True}],
+            ([{"d": "e"}], dt_earlier))
```
cyli: Hmm... apologies, actually, should this return []?  If there is a later cache, and there are no as-active servers in the later cache, then shouldn't active servers be []?  (e.g. the old cache should be ignored?)

otter/test/models/test_cass_models.py
```diff
@@ -3745,30 +3766,43 @@ def test_get_servers_diff_last_update(self):
...
-            [{"server_blob": '{"a": "b"}', "last_update": dt_earlier},
-             {"server_blob": '{"d": "e"}', "last_update": dt_earlier},
-             {"server_blob": '{"c": "f"}', "last_update": self.dt}],
+            False,
+            [{"server_blob": '{"a": "b"}', "last_update": dt_earlier,
+              "server_as_active": False},
+             {"server_blob": '{"d": "e"}', "last_update": dt_earlier,
+              "server_as_active": False},
+             {"server_blob": '{"c": "f"}', "last_update": self.dt,
+              "server_as_active": False}],
```
cyli: Actually sorry I didn't notice before either - although maybe I am just misunderstanding this now.  Should it fetch the latest cache, not the earliest cache?

otter/test/models/test_cass_models.py
manishtomar: Yes. you are right! The test is incorrectly written which unfortunately doesnt show that the code is incorrect. The code will still work since it will return list ordered differently, i.e. the list returned will have later date first which is what the test should do. I'll change it. Thanks for catching this!

cyli:
LGTM now, thanks!  :D

PR #1568, opened 2015-07-02
When deleting a CLB in the trial tests, retry

### LINKED ISSUE #1562: Trial tests should figure out if a CLB has actually been deleted to avoid cleanup errors.
We can't tell if a CLB has been deleted with the status code, necessarily.  No matter if it's in PENDING_UPDATE, PENDING_DELETE, or DELETED, the error message is:

```
{
    "message": "Must provide valid load balancers: 381529  could not be found.",
    "code": 400
}
```

Once it has been deleted, getting the load balancer results in:

```
{
    "loadBalancer": {
        "name": "a-load-balancer",
        "id": 381529,
        "status": "DELETED",
        "created": {
            "time": "2015-07-01T01:11:15Z"
        },
        "updated": {
            "time": "2015-07-01T01:19:28Z"
        }
    }
}
```

cyli:
Fixes #1562.

Because it might be in pending update, and we don't want the tests to fail.  Also, if it's deleted, a 404 is NOT returned for real.  It goes into PENDING_DELETE state for a while, then into DELETED state.

All of these states mean that the CLB is immutable, and a 400 is returned when trying to delete (the error says that the CLB is immutable, but not which state it's in).  Which means we need to GET the CLB to figure out if we succeeded or need to retry.

### DIFF
--- a/otter/integration/lib/cloud_load_balancer.py
+++ b/otter/integration/lib/cloud_load_balancer.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 
 import json
 
+from functools import wraps, partial
+
 from characteristic import Attribute, attributes
 
 from testtools.matchers import MatchesPredicateWithParams
@@ -11,6 +13,7 @@ from testtools.matchers import MatchesPredicateWithParams
 import treq
 
 from twisted.internet import reactor
+from twisted.internet.defer import inlineCallbacks
 from twisted.python.log import msg
 
 from otter.util.deferredutils import retry_and_timeout
@@ -34,6 +37,26 @@ def _pending_update_to_transient(f):
     if f.value.code == 422 and 'PENDING_UPDATE' in f.value.body:
         raise TransientRetryError()
     return f
+
+
+def _retry(reason, timeout=60, period=3, clock=reactor):
+    """
+    Helper that decorates a function to retry it until success it succeeds or
+    times out.  Assumes the function will raise :class:`TransientRetryError`
+    if it can be retried.
+    """
+    def decorator(f):
+        @wraps(f)
+        def retrier(*args, **kwargs):
+            return retry_and_timeout(
+                partial(f, *args, **kwargs), timeout,
+                can_retry=terminal_errors_except(TransientRetryError),
+                next_interval=repeating_interval(period),
+                clock=clock,
+                deferred_description=reason
+            )
+        return retrier
+    return decorator
 
 
 @attributes([
@@ -99,9 +122,10 @@ class CloudLoadBalancer(object):
         )
 
     def wait_for_state(
-        self, rcs, state_desired, timeout, period=10, clock=None
+        self, rcs, state_desired, timeout, period=10, clock=reactor
     ):
-        """Waits for the cloud load balancer to reach a certain state.  After
+        """
+        Wait for the cloud load balancer to reach a certain state.  After
         a timeout, a `TimeoutError` exception will occur.
 
         :param TestResources rcs: The resources used to make appropriate API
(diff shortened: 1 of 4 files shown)


cyli:

cyli:

otter/integration/lib/cloud_load_balancer.py
```diff
@@ -36,6 +39,26 @@ def _pending_update_to_transient(f):
     return f
 
 
+def _retry(reason, timeout=60, period=3, clock=reactor):
+    """
+    Helper that decorates a function to retry it until success it succeeds or
+    times out.  Assumes the function will raise :class:`TransientRetryError`
+    if it can be retried.
+    """
```
manishtomar: Nice factoring :)

otter/integration/lib/test_cloud_load_balancer.py
```diff
@@ -235,6 +238,120 @@ def add(clb, clock):
...
+
+            def get(cls, _url, *args, **kwargs):
+                cls.delete(_url, *args, **kwargs)
+                return succeed(get_response)
+
+            def content(cls, resp):
+                if get_response == Response(200):
+                    self.assertEqual(resp, del_response)
+                else:
+                    self.assertIn(resp, (del_response, get_response))
```
manishtomar: Non-blocking: I couldn't understand this `if`. Is it possible to document the intent of `get_response` and `del_response` parameters?

cyli: Good point, thanks!  I'll add that.

otter/integration/lib/test_cloud_load_balancer.py
```diff
@@ -235,6 +238,123 @@ def add(clb, clock):
...
+
+            def get(cls, _url, *args, **kwargs):
+                cls.delete(_url, *args, **kwargs)
+                return succeed(get_response)
+
+            def content(cls, resp):
+                # If the get_response is not a 200, then, it would be
+                # considered an error and hence treq.content would be called.
+                # Otherwise, treq.content is only called for the del response.
+                if get_response == Response(200):
```
manishtomar: I am really sorry but this `if` still confuses me. Is it possible to add `content` to `Response` defaulting to empty string? 
```python
class Response(object):
    """Fake response object"""
    def __init__(self, code, headers={}, content=''):
        self.code = code
        self.headers = headers
        self.content = content
```
Then `treq.content` and `.json_content` can have default implementation that returns the content from here. I can understand if you don't want to do this since you are already implementing proper Stub in treq itself.

manishtomar:
LGTM! Wonderful factoring :)

otter/integration/lib/test_cloud_load_balancer.py
cyli: Will do, thanks!

radix:
It looks like this can be merged? I'm not sure if there's a reason it hasn't, so I'll leave it to @cyli :)

cyli:
@radix:  @manishtomar suggested I run it against production first - and this was a great idea, because everything failed, because we hit CLB over limit errors.  The retry period I had was too short for this to run in parallel, so I increased the timeout and increased the polling period.  It worked at timeout=120 and period=15, but I upped it some more since we are running even more tests in prod (cafe tests too), just in case.

cyli:
Ok, no ran it a couple more times, and it's still sometimes hitting over limit errors.  Possibly on get requests.  I should handle that in this PR too, since we're introducing so much more retrying.  :|  Sorry

cyli:
Hmm... the 413 and timeout scaling is turning out to be a bigger change than I thought - I'm going to back out the last change I made (to the commit that @manishtomar reviewed), merge this, and submit a new PR for the 413 and polling changes.

Issue #1559, opened 2015-06-30
Handle CLB in BUILD state by RETRY-ing


cyli:
Some of our cafe tests is failing in production because the load balancers are building and nodes can't be added to them.  We should retry on BUILD.

status code: 422
body:
`{"message":"Load Balancer '42473' has a status of 'BUILD' and is considered immutable.","code":422}`

manishtomar:
As in convergence should be retried? or tests should retry?

cyli:
Er sorry, convergence should retry.  The step should return RETRY.

Issue #1551, opened 2015-06-26
Handle 410 load balancer deleted from CLBs


cyli:
Saw `APIError(code=410, body='{"message":"The loadbalancer is marked as deleted.","code":410}'` earlier.  We don't handle this currently.

radix:
this happens during node deletion.

radix:
and adding :)

PR #1569, opened 2015-07-02
Refactor CLB node removal to cloud_client

radix:
- moves the core logic (including retry when some nodes are invalid) into cloud_client.py
- refactored it all to use the same error-handling that the rest of cloud_client.py uses, simplifying the code in the process
- the RemoveNodesFromCLB step now interprets exceptions instead of HTTP codes

This is a preliminary step before implementing #1551 and #1559, which are *not* implemented in this branch, since this PR is getting big enough.

I'm going to disappear for several days after this so if there are any issues with the PR then please feel free to make the change directly in this branch. I'm also unassigning #1551 and #1559 from myself. sorry :-(

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -568,11 +568,12 @@ def remove_clb_nodes(lb_id, node_ids):
     This function will handle the case where *some* of the nodes are valid and
     some aren't, by retrying deleting only the valid ones.
     """
+    node_ids = map(str, node_ids)
     eff = service_request(
         ServiceType.CLOUD_LOAD_BALANCERS,
         'DELETE',
         append_segments('loadbalancers', lb_id, 'nodes'),
-        params={'id': [str(node_id) for node_id in node_ids]},
+        params={'id': node_ids},
         success_pred=has_code(202))
 
     def check_invalid_nodes(exc_info):
@@ -583,8 +584,8 @@ def remove_clb_nodes(lb_id, node_ids):
                 body, ["validationErrors", "messages", 0])
             match = _CLB_NODE_REMOVED_PATTERN.match(message)
             if match:
-                removed = map(int, concat([group.split(',')
-                                           for group in match.groups()]))
+                removed = concat([group.split(',')
+                                  for group in match.groups()])
                 return remove_clb_nodes(lb_id, set(node_ids) - set(removed))
         six.reraise(*exc_info)
 
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -30,7 +30,7 @@ from txeffect import deferred_performer, perform as twisted_perform
 from otter.auth import Authenticate, InvalidateToken, public_endpoint_url
 from otter.constants import ServiceType
 from otter.util.config import config_value
-from otter.util.http import APIError, append_segments
+from otter.util.http import APIError, append_segments, try_json_with_keys
 from otter.util.http import headers as otter_headers
 from otter.util.pure_http import (
     add_bind_root,
@@ -575,17 +575,25 @@ def remove_clb_nodes(lb_id, node_ids):
         params={'id': [str(node_id) for node_id in node_ids]},
         success_pred=has_code(202))
 
-    @_only_json_api_errors
-    def parse_known_errors(code, json_body):
+    def check_invalid_nodes(exc_info):
+        code = exc_info[1].code
+        body = exc_info[1].body
         if code == 400:
-            message = get_in(["validationErrors", "messages", 0], json_body)
+            message = try_json_with_keys(
+                body, ["validationErrors", "messages", 0])
             match = _CLB_NODE_REMOVED_PATTERN.match(message)
             if match:
-                removed = concat([group.split(',')
-                                  for group in match.groups()])
+                removed = map(int, concat([group.split(',')
+                                           for group in match.groups()]))
                 return remove_clb_nodes(lb_id, set(node_ids) - set(removed))
-        _process_clb_api_error(code, json_body, lb_id)
-    return eff.on(error=parse_known_errors)
+        six.reraise(*exc_info)
+
+    return eff.on(
+        error=catch(APIError, check_invalid_nodes)
+    ).on(
+        error=_only_json_api_errors(
+            lambda c, b: _process_clb_api_error(c, b, lb_id))
+    ).on(success=lambda _: None)
 
 
 def _expand_clb_matches(matches_tuples, lb_id, node_id=None):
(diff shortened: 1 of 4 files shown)


otter/cloud_client.py
```diff
@@ -552,6 +555,48 @@ def _parse_known_errors(code, json_body):
...
+        code = exc_info[1].code
+        body = exc_info[1].body
+        if code == 400:
+            message = try_json_with_keys(
+                body, ["validationErrors", "messages", 0])
+            match = _CLB_NODE_REMOVED_PATTERN.match(message)
+            if match:
+                removed = concat([group.split(',')
+                                  for group in match.groups()])
+                return remove_clb_nodes(lb_id, set(node_ids) - set(removed))
```
cyli: Non-blocking, as this can happen in a future PR, but I wonder if this should be done in the step phase rather than in cloud_client?

otter/test/convergence/test_steps.py
```diff
@@ -504,59 +506,16 @@ def test_remove_nodes_from_clb(self):
...
-        # This one is just malformed but similar to a good message.
-        self.assertFalse(predicate(
-            StubResponse(422, {}),
-            json.dumps({
-                "message": "The load balancer is considered immutable.",
-                "code": 422
-            })))
+        self.assertFalse(predicate(StubResponse(200, {}), None))
 
     def test_remove_nodes_from_clb_retry(self):
```
cyli: This seems like a duplicate test to the cloud_client test `test_remove_clb_nodes_retry_on_some_invalid_nodes` - One of them can probably be removed, but it depends on where we want the retry to happen.

otter/cloud_client.py
cyli: Actually apologies, maybe the decision is blocking, though, because there is a duplicate test in both steps and cloud_client, and we should probably decide where it should go :)

radix: I think it's reasonable to put it here. I just left the test in steps so we know there's no regression.

otter/test/test_cloud_client.py
```diff
@@ -719,6 +721,66 @@ def test_add_clb_nodes(self):
...
+    def test_remove_clb_nodes_random_400(self):
+        """Random 400s are bubbled up as an APIError."""
+        eff = remove_clb_nodes(self.lb_id, [1, 2])
+        response = stub_pure_response(
+            {'validationErrors': {'messages': ['bar']}}, 400)
+        seq = [
+            (self.expected_node_removal_req().intent,
+             service_request_eqf(response)),
+        ]
+        self.assertRaises(APIError, perform_sequence, seq, eff)
```
cyli: We should probably call `self.assert_parses_common_clb_errors(expected.intent, eff)` here or in the previous test.

otter/cloud_client.py
cyli: Oh ok - in that case I'll leave a comment to that effect in the steps tests so we know it can be removed in the next PR.

otter/cloud_client.py
```diff
@@ -552,6 +555,48 @@ def _parse_known_errors(code, json_body):
...
+        'DELETE',
+        append_segments('loadbalancers', lb_id, 'nodes'),
+        params={'id': node_ids},
+        success_pred=has_code(202))
+
+    def check_invalid_nodes(exc_info):
+        code = exc_info[1].code
+        body = exc_info[1].body
+        if code == 400:
+            message = try_json_with_keys(
```
cyli: This seems to be the only usage left of `try_json_with_keys` - we can probably remove it and use `_only_json_api_errors` again here?

cyli: Oh nm, sorry, that re-raises no matter what and this wants to return.

radix: Yeah, I actually tried to do this at one point but then I realized _only_json_api_errors doesn't allow recovery, after being confused for a while :)

cyli: This might return None though, if the body is malformed, which might break when attempting to match it to a pattern.

cyli: @radix: wondering if I should refactor that a bit to allow recovery...

cyli:
We should probably also add tests for steps to ensure that PENDING_UPDATE/any other non-4xx error gets a RETRY, and only 4XX errors get a FAILURE.

otter/test/convergence/test_steps.py
cyli: Added comment to the steps test that it should be able to be removed at a later point:  https://github.com/rackerlabs/otter/commit/4122e9a1e1913859846f6e0f7fbe36ec0856ad13

otter/cloud_client.py
cyli: Added check if `message=None` in https://github.com/rackerlabs/otter/commit/7948aa9f2bfc519dd768715cea15e247ed757035

radix:
@cyli Is there anything else to do on this PR? the changes you made to the branch all LGTM :)

cyli:
@radix:  nope I think that was it for me :). Just needed someone to review the changes :)

cyli:

cyli:

cyli:
This PR seems cursed with jenkins network failures.  Will try again in a bit.

cyli:

cyli:
This is failing due to a bug in mimic that was being ignored before.

The bug is that if the server is deleted, mimic should return a 422 with the message "The load balancer is deleted and considered immutable."

Instead, it returns "Load balancer not ACTIVE" (https://github.com/rackerlabs/mimic/blob/master/mimic/model/clb_objects.py#L370).

We check for a regex to make sure that it's deleted before counting it as a success.  Previously, this check failed, and the resultant API error containing this message was not handled, resulting in it being bubbled up to `execute_convergence`, which I guess resulted in convergence retrying - so it would succeed, because the next convergence run, there would be no load balancer and hence no nodes, and the group is all good.

Now, any 4XX APIError results in a FAILURE, which means that this group goes into an error state.  I'll file the bug with mimic.

cyli:
Ok, I just double-checked responses from CLB.  As it turns out, it is not exactly a mimic bug.  With PENDING_DELETE, the error message is indeed "Load balancer is not ACTIVE".  If the CLB is in DELETED state, it's "The load balancer is deleted and considered immutable."

Until a CLB is deleted, we can't tell why it's in an immutable state without a GET.  See https://github.com/rackerlabs/mimic/issues/270#issuecomment-120297327 for a full rundown of the error messages.

It looks like though we can just return RETRY if we get "Load balancer is not ACTIVE"?  (which ONLY gets returned when trying to delete nodes, not adding nodes or changing nodes or deleting the CLB, and which ONLY gets returned on BUILD, PENDING_UPDATE, and PENDING_DELETE  -__-)

radix:

cyli:
Awesome, thanks for fixing that so fast.  LGTM

PR #1570, opened 2015-07-02
LBSH: fix yields to work in production

kivattik:


otter/integration/tests/test_lbsh.py
```diff
@@ -70,24 +70,23 @@ def test_oob_deleted_clb_node(self):
...
+
         nodes = yield clb.list_nodes(self.rcs)
-        self.assertEqual(
-            len(nodes['nodes']), 1,
-            "There should be 1 node on the CLB now that the group is active.")
         the_node = nodes["nodes"][0]
 
         yield clb.delete_nodes(self.rcs, [the_node['id']])
 
         nodes = yield clb.list_nodes(self.rcs)
```
cyli: This line can probably removed since we are just waiting for the nodes to be 0  :)

otter/integration/tests/test_lbsh.py
```diff
@@ -120,24 +119,20 @@ def test_move_node_to_oob_lb(self):
...
-                         "There should be no nodes on the CLB yet.")
+        yield clb_as.wait_for_nodes(
+            self.rcs, HasLength(0), timeout=timeout_default)
+        yield clb_other.wait_for_nodes(
+            self.rcs, HasLength(0), timeout=timeout_default)
 
         group, _ = self.helper.create_group(min_entities=1)
         yield self.helper.start_group_and_wait(group, self.rcs)
 
         nodes_as = yield clb_as.list_nodes(self.rcs)
```
cyli: This should probably be moved down below the `wait_for_nodes` calls, since listing might result in 0 nodes if CLB hasn't updated yet.  If it's done after the waits, then we can be sure that there is 1.

cyli:
Actually in order to make it less confusing as to when things should be listed vs waiting, maybe we could just incorporate the change to `wait_for_nodes` here to return the nodes?  Should just be the changes to the `cloud_load_balancer.py` file and `test_cloud_load_balancer.py` file from this PR:  https://github.com/rackerlabs/otter/pull/1564/files

Apologies, it just seems like it's useful for other PRs too :)

kivattik:
Agree. I'll review that PR then incorporate.

cyli:
Oh ok, I just meant pick those two changes out of that PR and put it in this one.  :)  And then when this one gets merged, I can merge forward the other one.  But your suggestion would also work.

otter/integration/tests/test_lbsh.py
```diff
@@ -74,25 +74,21 @@ def test_oob_deleted_clb_node(self):
         """
         clb = self.helper.clbs[0]
 
-        nodes = yield clb.list_nodes(self.rcs)
-        self.assertEqual(len(nodes['nodes']), 0,
-                         "There should be no nodes on the CLB yet.")
+        nodes = yield clb.wait_for_nodes(
```
cyli: Non-blocking - if the nodes aren't needed, it's probably less confusing to just not assign it to a variable, to make it more clear that we are just asserting something about the length, as opposed to doing something with the nodes here.

otter/integration/tests/test_lbsh.py
```diff
@@ -155,8 +145,6 @@ def test_move_node_to_oob_lb(self):
         yield clb_as.delete_nodes(self.rcs, [the_node['id']])
         yield clb_other.add_nodes(self.rcs, [node_info])
 
-        nodes_as = yield clb_as.list_nodes(self.rcs)
-        nodes_other = yield clb_other.list_nodes(self.rcs)
         clb_as.wait_for_nodes(
```
cyli: These need to be yielded, or the tests won't actually wait until this is the case.  :|

cyli:

cyli:
This LGTM pending +1 from jenkins :)

PR #1579, opened 2015-07-09
REST group state active from servers cache table

manishtomar:
For convergence tenants, getting group state active servers in rest layer from servers cache table instead of from scaling_group table. Removed updating old group state cache during convergence gathering.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -224,13 +224,12 @@ def convergence_succeeded(scaling_group, group_state, servers, now):
                                        status=ScalingGroupStatus.ACTIVE))
         yield cf_msg('group-status-active',
                      status=ScalingGroupStatus.ACTIVE.name)
-    else:
-        # update servers cache with latest servers
-        yield Effect(
-            UpdateServersCache(
-                scaling_group.tenant_id, scaling_group.uuid, now,
-                [merge(thaw(s.json), {"_is_as_active": True})
-                 for s in servers if s.state != ServerState.DELETED]))
+    # update servers cache with latest servers
+    yield Effect(
+        UpdateServersCache(
+            scaling_group.tenant_id, scaling_group.uuid, now,
+            [merge(thaw(s.json), {"_is_as_active": True})
+                for s in servers if s.state != ServerState.DELETED]))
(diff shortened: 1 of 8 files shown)


manishtomar:

otter/rest/groups.py
```diff
@@ -6,15 +6,22 @@
 
 from functools import partial
 
+from twisted.internet import reactor
```
cyli: I think both `CassScalingGroup` and `CassScalingGroupCollection` have a reactor on the objects, which can be passed to `get_active_cache` rather than importing the global reactor here.

otter/rest/groups.py
```diff
@@ -442,6 +473,18 @@ def __init__(self, store, tenant_id, group_id):
         self.tenant_id = tenant_id
         self.group_id = group_id
 
+    def with_active_cache(self, get_func, *args, **kwargs):
```
cyli: Nice!  This makes all the other changes much shorter than they could have been!

otter/rest/groups.py
manishtomar: Oh yeah. Didn't realize. Thanks! Will change it.

otter/convergence/service.py
```diff
@@ -237,8 +229,8 @@ def convergence_succeeded(scaling_group, group_state, servers, now):
         yield Effect(
```
cyli: Non-blocking question that has nothing to with this PR - I think I'm just blanking - how is the group's active servers updated if the server bounces back from ERROR to ACTIVE?  This seems to only update the cache if the group was ACTIVE in the first place.

We specifically have a test for recovering from ERROR, and it passes, I just can't figure out what that mechanism is right now.  I'm just missing something obvious probably.

manishtomar: This is great catch @cyli. The test passes because it does not check about servers being updated. I'll add it.

cyli:

cyli:
LGTM pending +1 from jenkins!  :)  Thanks for fixing this - so nice to not use the state for active!

manishtomar:
Thanks @cyli :) Merging.

PR #1581, opened 2015-07-09
Small fix to CF dispatcher

cyli:
Cloudfeeds was calling `get_full_dispatcher` with the wrong number of arguments, but it really only needs to make `cloud_client` calls, so switch dispatchers to `get_legacy_dispatcher`.

Wondering if we should rename `get_legacy_dispatcher`, since it's just a dispatcher that makes cloud client calls in addition to the simple dispatcher?

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -62,7 +62,6 @@
     "cloudfeeds": {
         "service": "cloudFeeds",
         "tenant_id": "identity_admin_tenant",
-        "region": "CF region to log to",
         "url": "http://cfurl.net/not/in/service/catalog"
     },
     "converger": {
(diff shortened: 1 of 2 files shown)


manishtomar:
> Wondering if we should rename get_legacy_dispatcher, since it's just a dispatcher that makes cloud client calls in addition to the simple dispatcher?

How about `get_cloud_client_dispatcher`? But that should be separate PR if you decide to do it :).

cyli:
> How about get_cloud_client_dispatcher? But that should be separate PR if you decide to do it :).

There already is one.  :|  `get_legacy_dispatcher` is the composition of that and `get_simple_dispatcher`, otherwise I would have just gone and renamed.

manishtomar:
LGTM!

cyli:
Thanks!

PR #1587, opened 2015-07-10
Treat CLB immutability consistently in error handling

### LINKED ISSUE #1559: Handle CLB in BUILD state by RETRY-ing

radix:
Fixes #1559

Instead of adding yet another exception type, I generalized CLBPendingUpdateError to CLBImmutableError and made our regex match that message no matter what status it is.

We still treat 'DELETED' specially, in that we raise CLBDeletedError in that case.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -326,16 +326,13 @@ def get_cloud_client_dispatcher(reactor, authenticator, log, service_configs):
 
 # ----- CLB requests and error parsing -----
 
-_CLB_PENDING_UPDATE_PATTERN = re.compile(
-    "^Load Balancer '\d+' has a status of 'PENDING_UPDATE' and is considered "
+_CLB_IMMUTABLE_PATTERN = re.compile(
+    "^Load Balancer '\d+' has a status of '[^']+' and is considered "
     "immutable\.$")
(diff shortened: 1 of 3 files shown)


radix:

radix:

otter/cloud_client.py
```diff
@@ -649,7 +652,7 @@ def _process_clb_api_error(api_error_code, json_body, lb_id):
           partial(CLBRateLimitError, lb_id=lb_id))] +
         _expand_clb_matches(
             [(422, _CLB_DELETED_PATTERN, CLBDeletedError),
-             (422, _CLB_PENDING_UPDATE_PATTERN, CLBPendingUpdateError),
+             (422, _CLB_IMMUTABLE_PATTERN, CLBImmutableError),
```
cyli: Non-blocking - wondering if we should comment here that this particular ordering matters, otherwise a PENDING_DELETE server will be recognized as just a `CLBImmutableError`?

cyli:
Aside from the above nitpick, LGTM if jenkins +1's!

Issue #1578, opened 2015-07-09
"error_reasons" field in group state


manishtomar:
This new field will be returned inside "state" whereever group state is returned in REST API (`/groupId/state`, `/groups/groupId` and `groups/`). It will contain `;`-separated list of reasons why the group is in ERROR. When the group is ACTIVE, this field will not be there. Hence state will look like 
```
      "state":{
         "active":[
         ],
         "status": "ERROR",
         "error_reasons": "Cloud load balancer 2134 has been deleted; Invalid imageRef 605e13f6-1452-4588-b5da-ac6bb468c5bf".
         "activeCapacity":0,
         "desiredCapacity":0,
         "paused":false,
         "pendingCapacity":0,
         "name":"smallest possible launch config group"
```
The reasons will contain same content that is injected to CF [at the end of convergence](https://github.com/rackerlabs/otter/blob/7794ab65fa7b0b990a3318263158bcf51159735d/otter/convergence/service.py#L233-L235).

Can be split up into:
- [x] Convergence putting reasons in DB
- [x] REST returning reasons in state calls

cyli:
Just to keep track of what we're doing, I think from discussion, we are probably just going to name the field `"errors": []`.

It will be a list of dictionaries that look like:  `[{"message": ....}, {"message": ....}]`

That way, if we wanted to insert an error type or error UUID or error link to documentation that explains how to fix it or something, we won't have to deprecate the API or add a new field.

Does that sound accurate?

manishtomar:
Yes @cyli . Thank you so much for updating it here.

cyli:
Closing since this #1608 got merged and integration tests (against mimic) pass :)

PR #1588, opened 2015-07-12
Convergence updates group error reasons in DB

manishtomar:
First task of #1578

### DIFF
--- a/schema/migrations/control_52_add_error_reasons_col.cql
+++ b/schema/migrations/control_52_add_error_reasons_col.cql
@@ -1,0 +1,6 @@
+USE @@KEYSPACE@@;
+
+-- Add "deleting" column to scaling_groups table
+
+ALTER TABLE scaling_group
+ADD error_reasons list<ascii>;
--- a/schema/migrations/control_52_add_error_reasons_col.cql
+++ b/schema/migrations/control_52_add_error_reasons_col.cql
@@ -1,6 +1,6 @@
 USE @@KEYSPACE@@;
 
--- Add "deleting" column to scaling_groups table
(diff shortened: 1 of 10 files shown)


schema/setup/control_15_create_scaling_groups.cql
```diff
@@ -52,6 +52,7 @@ CREATE COLUMNFAMILY scaling_group (
     created_at timestamp,
     status ascii,
     deleting boolean,
+    error_reasons list<ascii>,
```
radix: It seems given that this is human-readable text it should be `text` instead of `ascii`. as I understand it, `ascii` only allows bytes 0-127, so we can't even encode utf-8 into this field.

manishtomar: Yeah. Good point. Let me change it to `text` and check if it works fine.

radix:

schema/migrations/control_52_add_error_reasons_col.cql
```diff
@@ -0,0 +1,6 @@
+USE @@KEYSPACE@@;
+
+-- Add "deleting" column to scaling_groups table
+
+ALTER TABLE scaling_group
+ADD error_reasons list<ascii>;
```
radix: looks like you forgot to fix this one :)

radix:
This LGTM after you fix the remaining `<ascii>` and jenkins passes :)

manishtomar:
Hmm.. The tests seem to be work on dev vm. I am just going to try it again for now.

manishtomar:

PR #1589, opened 2015-07-13
CLB trial tests retry on 413

cyli:
Handle 413 from CLB, and extend the timeouts for retrying because of transient errors.

### DIFF
--- a/otter/integration/lib/autoscale.py
+++ b/otter/integration/lib/autoscale.py
@@ -281,8 +281,8 @@ class ScalingGroup(object):
             This provides useful information to complete the request, like
             which endpoint to use to make the API request.
         """
-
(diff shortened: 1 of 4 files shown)


cyli:
Actually now I'm encountering intermittent network errors, so I'm thinking about just making all the tests use a retrying treq, which always retries on 413 overlimit (because all of Rackspace's cloud uses repose and produces the same request rate limiting message), and `twisted.web._newclient.ResponseNeverReceived` errors.

radix:
@cyli  :+1:

cyli:
Ok, closing this in favor of #1591

cyli:
Ok, I think retrying-treq may not be working out.  :|    We may need this again.

PR #1593, opened 2015-07-16
Fix sending events to CF multiple times

### LINKED ISSUE #1584: CF observer logs a message twice to CF
Created a group with minentities 1, and an invalid load balancer.

Logging to https://gist.github.com/cyli/1f6b1f832c2e3f4d79f6.

Seeing 4 CF messages with the same message ID.

This is likely related to #1580

cyli:
This addresses both the issue where we send an invalid CF event multiple times, and where we send a valid CF event multiple times.

Fixes #1584.

### DIFF
--- a/otter/test/log/test_cloudfeeds.py
+++ b/otter/test/log/test_cloudfeeds.py
@@ -255,14 +255,15 @@ class EventTests(SynchronousTestCase):
 
         return perform_sequence(seq, eff)
 
-    def test_add_event_succeeds_on_201_with_xml_body(self):
+    def test_add_event_succeeds_if_request_succeeds(self):
         """
-        Adding an event succeeds on a 201, and ignores whether the body is JSON
-        or not.
+        Adding an event succeeds without retrying if the service request
+        succeeds.
         """
-        resp = stub_pure_response("<some xml>", 202)
-        response = [lambda _: resp]
-        self.assertEqual(self._perform_add_event(response), resp)
+        body = "<some xml>"
+        resp = stub_pure_response(body, 202)
+        response = [lambda _: (resp, body)]
+        self.assertEqual(self._perform_add_event(response), (resp, body))
 
     def test_add_event_only_retries_5_times_on_non_4xx_api_errors(self):
         """
@@ -273,8 +274,8 @@ class EventTests(SynchronousTestCase):
             lambda _: raise_(Exception("oh noes!")),
             lambda _: raise_(ResponseFailed(Failure(Exception(":(")))),
             lambda _: raise_(APIError(code=100, body="<some xml>")),
+            lambda _: raise_(APIError(code=202, body="<some xml>")),
             lambda _: raise_(APIError(code=301, body="<some xml>")),
-            lambda _: raise_(APIError(code=500, body="<some xml>")),
             lambda _: raise_(APIError(code=501, body="<some xml>")),
         ]
         with self.assertRaises(APIError) as cm:
--- a/otter/test/log/test_cloudfeeds.py
+++ b/otter/test/log/test_cloudfeeds.py
@@ -10,6 +10,7 @@ from effect import Effect, TypeDispatcher
 import mock
 
 from twisted.internet.defer import fail, succeed
+from twisted.python.failure import Failure
 from twisted.trial.unittest import SynchronousTestCase
 
 from txeffect import deferred_performer
@@ -29,6 +30,7 @@ from otter.log.cloudfeeds import (
 from otter.log.formatters import LogLevel
 from otter.log.intents import Log, LogErr
 from otter.test.utils import CheckFailure, mock_log, patch, resolve_effect
+from otter.util.http import APIError
 from otter.util.retry import (
     ShouldDelayAndRetry,
     exponential_backoff_interval,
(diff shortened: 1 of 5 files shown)


otter/test/log/test_cloudfeeds.py
```diff
@@ -211,39 +224,87 @@ def _get_request(self, _type, _id, tenant_id):
...
+        def perform_retry_without_delay(retry_intention):
+            def should_retry(exc_info):
+                exc_type, exc_value, exc_traceback = exc_info
+                failure = Failure(exc_value, exc_type, exc_traceback)
+                return Effect(Constant(
+                    retry_intention.should_retry.can_retry(failure)))
+
+            seq = [(svrq, resp) for resp in response_sequence]
+
+            new_retry_intent = Effect(Retry(effect=retry_intention.effect,
```
manishtomar: Nitpick: `new_retry_effect`?

cyli: Oh yes, thanks!

otter/test/log/test_cloudfeeds.py
```diff
@@ -211,39 +224,87 @@ def _get_request(self, _type, _id, tenant_id):
...
+                (Retry(effect=mock.ANY, should_retry=ShouldDelayAndRetry(
+                    can_retry=mock.ANY,
+                    next_interval=exponential_backoff_interval(2))),
+                 # we actually want to perform the effect multiple times so
+                 # we can test retrying
+                 perform_retry_without_delay)
+                ]))
+        ]
+
+        return perform_sequence(seq, eff)
```
manishtomar: Nice way of writing test :). It took me sometime to understand this and I am not sure if I would've been able to do this.

cyli: Ah, sorry, is there any way I can make this more readable?  I was originally just going to test the `can_retry` separately, but that wouldn't have caught that weird issue with `json_response` yesterday, so I wanted to try the whole retry.  Maybe I can filter `perform_retry_without_delay` into a `retry_sequence` function that is sort of like `perform_sequence`?

otter/test/log/test_cloudfeeds.py
```diff
@@ -211,39 +224,87 @@ def _get_request(self, _type, _id, tenant_id):
...
+
+    def test_add_event_succeeds_on_201_with_xml_body(self):
+        """
+        Adding an event succeeds on a 201, and ignores whether the body is JSON
+        or not.
+        """
+        resp = stub_pure_response("<some xml>", 202)
```
manishtomar: Hmm. Unfortunately, we need to retry on 202 also. I remember having a disappointing discussion with them about this. They seem to have some condition which can trigger 202 without event being submitted and they refuse to consider it a bug.

cyli: Oops, sorry, this was supposed to be a 201.  And hmm... this test should have failed.  Good catch!

cyli: Oh right, it didn't fail because service response didn't raise - any result is valid.  I'll make sure to test the response code.

otter/test/log/test_cloudfeeds.py
cyli: Meh, actually I realized this doesn't actually test the `json_response` thing either, since I'm not stubbing out Request, just ServiceRequest.  :|

radix: I haven't looked at this test too closely, just this comment thread, but one option would be to use the real `ServiceRequest` performer (`concretize_service_request`) here, which would cause the pure_http.Request to be run (meaning you could put it in the `SequenceDispatcher` list right after the `ServiceRequest`)

that may be tedious/ugly, so other options may be better

otter/test/log/test_cloudfeeds.py
radix: also I'm not sure if `intention` was an intentional (*cough*) difference from `intent` but if not then I guess it'd be better to keep with the convention and call this `retry_intent`. (mega nitpick!)

cyli: No, it was just late at night :D  Will fix, thanks!

radix:
This LGTM! Beautiful changes!

otter/test/log/test_cloudfeeds.py
```diff
@@ -211,39 +223,72 @@ def _get_request(self, _type, _id, tenant_id):
...
+
+        return perform_sequence(seq, eff)
+
+    def test_add_event_succeeds_if_request_succeeds(self):
+        """
+        Adding an event succeeds without retrying if the service request
+        succeeds.
+        """
+        body = "<some xml>"
+        resp = stub_pure_response(body, 202)
```
manishtomar: Sorry for chiming in late, but this test informs that 202 succeeds whereas it should be retrying. Could you change it to 201? Also, am not sure how could you test that 202 should retry since any stub response with any code will succeed.

cyli: Yep, sorry, I was just testing that any successful request succeeds here, but you're right, a 201 would be clearer.  I moved the testing of 201/202 to https://github.com/rackerlabs/otter/pull/1593/files#diff-f7c901865267f33c308c719213352948R1147 instead.

I have added the issue here:  https://github.com/rackerlabs/otter/issues/1603

manishtomar: Oh that is nice. Thanks :)

PR #1594, opened 2015-07-16
Fuzzier CLB and Nova error matching

### LINKED ISSUE #1497: Slightly fuzzier regexp matching when parsing error messages from CLB and Nova
https://github.com/rackerlabs/otter/issues/1475 was an issue where a slight discrepancy in mimic broke the test.

While this is not necessarily a bad thing, we don't have a full enumeration of all possible CLB errors (in Nova at least maybe we can look through the source), so maybe it'd make sense to make the regex matches slightly fuzzier?

cyli:
Removed the `^` and `$`, do not definitively expect a period at the end of a sentence, do case-insensitive regex matching, and in particular, be able to match against "Load Balancer", "Load balancer" "LoadBalancer", "loadbalancer" and "load balancer" for every CLB message.

Fixes #1497

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -325,28 +325,34 @@ def get_cloud_client_dispatcher(reactor, authenticator, log, service_configs):
 
 
 # ----- CLB requests and error parsing -----
+def _regex(pattern):
+    """
+    Compile a case-insensitive pattern.
+    """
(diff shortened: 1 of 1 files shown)


cyli:

radix:
Nice! Now we'll be able to support `loAD     \t\t BALancer NOT FOUND` :-)

PR #1600, opened 2015-07-17
Make pure_http requests use effectful bound logs

### LINKED ISSUE #1598: convergence HTTP requests do not have tenant_id/scaling_group_id bound in logs
either we need to pass a proper bound log in somewhere, or we should make some effect performer get the log from the effectful context.

We may need to add a new intent: GetBoundLog.

radix:
Fixes #1598

The problem is that the new effectful log-binding system in otter.log.intents only works "with itself", when `intents.msg()` and `intents.err()` are used. To remedy this I added a `GetFields` / `get_fields` intent, which returns the bound fields as a dictionary, and then used that intent in the `pure_http.Request` performer to get those effect-bound fields and bind them into the log we pass to logging_treq.

### DIFF
--- a/otter/log/intents.py
+++ b/otter/log/intents.py
@@ -45,12 +45,9 @@ class LogErr(object):
 
 
 @attr.s
-class GetLog(object):
+class GetFields(object):
     """
-    Intent to get a BoundLog from the effectful context.
-
-    This is useful to pass a log object to "legacy" code using BoundLogs
-    directly.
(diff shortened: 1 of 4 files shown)


manishtomar:
LGTM! Tested this on dev vm. Works great :)

manishtomar:

radix:
thanks @manishtomar :)

PR #1602, opened 2015-07-18
make model intent performers also merge effectful logs

radix:
This should fix the logs we see in #otter-logs that look like this:

```
(%{tenant_id}/%{scaling_group_id}) NoSuchScalingGroupError('No such scaling group f23148ac-be66-4a6f-a0ef-fe1da19cbb86 for tenant 851153',)
```

so that they actually have tenant_id and scaling_group_id set.

Same problem as #1600, with the same solution. I've refactored the code to merge logs into a central place so all the performers can use it.

I also did a fair amount of bikeshedding on the model/test_intents.py code, because all the mutable instance-attributes were hurting my brain :confounded:

### DIFF
--- a/otter/test/models/test_intents.py
+++ b/otter/test/models/test_intents.py
@@ -8,12 +8,14 @@ import mock
 from twisted.internet.defer import succeed
 from twisted.trial.unittest import SynchronousTestCase
 
+from otter.log.intents import get_log_dispatcher
 from otter.models.intents import (
     DeleteGroup, GetScalingGroupInfo, UpdateGroupStatus, UpdateServersCache,
     get_model_dispatcher)
 from otter.models.interface import (
     IScalingGroupCollection, ScalingGroupStatus)
-from otter.test.utils import EffectServersCache, iMock, mock_group, mock_log
+from otter.test.utils import (
+    EffectServersCache, IsBoundWith, iMock, matches, mock_group, mock_log)
 
 
 class ScalingGroupIntentsTests(SynchronousTestCase):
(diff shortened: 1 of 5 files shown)


otter/test/models/test_intents.py
```diff
@@ -25,15 +27,26 @@ def setUp(self):
...
+    def perform_with_group(self, eff, expected_lookup, group, disp=None):
+        """Run an effect that will look up group info."""
+        def gsg(log, tenant_id, group_id):
+            assert (log, tenant_id, group_id) == expected_lookup
+            return group
+        store = self.get_store()
+        store.get_scaling_group.side_effect = gsg
+        dispatcher = self.get_dispatcher(store)
+        if disp is not None:
+            dispatcher = ComposedDispatcher([dispatcher, disp])
```
manishtomar: Non-blocking nitpick: `[dispatcher, disp]` seems a little confusing. I wonder if any of the two can have better names?

otter/test/models/test_intents.py
```diff
@@ -48,30 +63,63 @@ def view_manifest(with_policies, with_webhooks, get_deleting):
...
-            'del')
+        result = self.perform_with_group(
+            Effect(DeleteGroup(tenant_id='00', group_id='g1')),
+            (self.log, '00', 'g1'), self.group)
+        self.assertEqual(result, 'del')
+
+    def test_delete_group_log_context(self):
+        """
+        When run in an effectful log context, the fields are bound to the log
+        passed to get_scaling_group.
```
cyli: Nitpick: "passed to `delete_group`?

cyli:
The unit test fixes are lovely, thank you!  Aside from my docstring nitpick, LGTM!

radix:
I realized that, yeah. Well, it's still something we should do. I don't know what the heck is producing those un-bound errors.

cyli:
@radix Agreed, this PR is good as is, and mergeable - I guess we should file an issue to track down the unbound errors.

PR #1607, opened 2015-07-21
Generate the CF UUID when logging

### LINKED ISSUE #1603: Change CF test to reflect the correct status code
As mentioned here:  https://github.com/rackerlabs/otter/pull/1593/files#r35052167 - the correct status code is 201.  However, testing at the CF observer level means we can't really test that a 202 retries, unless we resolve the service request.

That is tested in the cloud_client instead.

But the test should be a 201 anyway to reflect the fact that a 201 is the correct response.

cyli:
So that our logs by default have the cloud feeds ID in them, so that if something goes wrong, or a message got mangled or dropped, we can find them in our logs.

Fixes #1144.
Fixes #1603.

### DIFF
--- a/otter/log/cloudfeeds.py
+++ b/otter/log/cloudfeeds.py
@@ -49,9 +49,10 @@ log_cf_mapping = {
 }
 
 
-def _cf_log(log_effect, *args, **kwargs):
+def _cf_log_with_id(log_effect, *args, **kwargs):
     """
-    Helper helper function to log cloud feed event.
+    Helper helper function to generate a cloud feed ID before logging a
+    log cloud feed event.
     """
     def log_to_cf(uid):
         kwargs.update({'cloud_feed': True, 'cloud_feed_id': uid})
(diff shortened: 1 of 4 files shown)


otter/log/cloudfeeds.py
```diff
@@ -50,25 +49,36 @@ def __init__(self, message):
 }
 
 
+def _cf_log(log_effect, *args, **kwargs):
```
manishtomar: How about `cf_log_with_id`? And similar docstr for the function?

otter/test/log/test_cloudfeeds.py
```diff
@@ -258,10 +271,11 @@ def _perform_add_event(self, response_sequence):
     def test_add_event_succeeds_if_request_succeeds(self):
         """
         Adding an event succeeds without retrying if the service request
-        succeeds.
+        succeeds.  Testing what response code causes a service request to
+        succeeds is beyond the scope of this test.
```
manishtomar: Nice comment :)

manishtomar:
LGTM!

PR #1608, opened 2015-07-21
Group error reasons returned in REST

manishtomar:
Second task #1578


manishtomar:

otter/models/interface.py
```diff
@@ -40,8 +42,8 @@ class GroupState(object):
     TODO: ``remove_active``, ``pause`` and ``resume`` ?
     """
     def __init__(self, tenant_id, group_id, group_name, active, pending,
-                 group_touched, policy_touched, paused, status, desired=0,
-                 now=timestamp.now):
+                 group_touched, policy_touched, paused, status,
+                 error_reasons=[], desired=0, now=timestamp.now):
```
cyli: Non-blocking - since we're just creating new lists and assigning the error reasons anyway (we're not mutating anything), this is fine, but just for the sake of making sure nothing gets broken by accident in the future, can we set `error_reasons` to a default of `()` instead?

otter/test/rest/test_groups.py
```diff
@@ -152,8 +166,30 @@ def test_format_state_different_status(self, config_value):
             True,
             ScalingGroupStatus.ERROR,
             desired=10)
-        result = format_state_dict(state, active)
+        result = self.format(state, active)
         self.assertEqual(result['status'], 'ERROR')
+        self.assertEqual(result['errors'], [])
```
cyli: Non-blocking, since this is just a test, but this should technically never happen, right?  All errors should have at least one reason?

otter/models/interface.py
manishtomar: Good idea.

otter/test/rest/test_groups.py
manishtomar: That is true but that condition is not something which this function can ensure. I could change the test with a reason but I not sure if it buys us something apart from avoiding confusion?

cyli: Mostly I wanted to verify.  But yes, it's only for avoiding confusion, so someone new to the codebase who reads this won't be like "oh, it's totally valid for errors to be empty".

cyli:
This LGTM!

manishtomar:

cyli:
LGTM!  Thanks!

PR #1611, opened 2015-07-21
present reasons fallback

manishtomar:
If none of the errors occurred can be presented then a default unknown error is presented


cyli:
LGTM, thanks for finishing this so quickly @manishtomar!

PR #1612, opened 2015-07-21
Error reasons integration test

### LINKED ISSUE #1609: Integration test for error reasons
1. Create group with an invalid key name, because autoscale does not
   verify the key and hence won't just 400 the create group request.
2. Assert group goes into error state with the reason being invalid
   Server config.
3. Delete the key, and converge.
4. Assert group goes into active state.
5. Change the launch config to have an invalid load balancer.
6. Scale up.
7. Assert group goes into error state with the reason being an invalid
   CLB.

cyli:
Fixes #1609.

This is dependent upon #1608 to merge for this test to pass.

### DIFF
--- a/otter/convergence/errors.py
+++ b/otter/convergence/errors.py
@@ -38,18 +38,19 @@ def _present_exception(exception):
 
 @_present_exception.register(NoSuchCLBError)
 def _present_no_such_clb_error(exception):
-    return "Cloud Load Balancer does not exist: %s" % (exception.lb_id,)
+    return "Cloud Load Balancer does not exist: {0}".format(exception.lb_id)
 
 
 @_present_exception.register(CLBDeletedError)
 def _present_clb_deleted_error(exception):
-    return ("Cloud Load Balancer is currently being deleted: %s"
-            % (exception.lb_id,))
+    return "Cloud Load Balancer is currently being deleted: {0}".format(
+        exception.lb_id)
 
 
 @_present_exception.register(CreateServerConfigurationError)
 def _present_server_configuration_error(exception):
-    return "Server launch configuration is invalid: %s" % (exception.message,)
+    return "Server launch configuration is invalid: {0}".format(
+        exception.message)
 
 
 @match(ErrorReason)
--- a/otter/convergence/errors.py
+++ b/otter/convergence/errors.py
@@ -6,7 +6,11 @@ from sumtypes import match
 
 from toolz.functoolz import identity
 
-from otter.cloud_client import CLBDeletedError, NoSuchCLBError
+from otter.cloud_client import (
+    CLBDeletedError,
+    CreateServerConfigurationError,
+    NoSuchCLBError
+)
 from otter.convergence.model import ErrorReason
 from otter.log.formatters import serialize_to_jsonable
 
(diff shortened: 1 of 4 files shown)


otter/integration/tests/test_convergence.py
```diff
@@ -989,6 +990,101 @@ def test_recover_from_identity_auth_failures(self):
             max_entities=10)
         yield self.helper.start_group_and_wait(group, rcs, desired=5)
 
+    @inlineCallbacks
+    def test_error_reasons_are_updated(self):
```
manishtomar: How about skipping this with link to #1578 and merging it? Then I can merge #1608 with this enabled. That way we that merge will be synchronous with the test.

otter/convergence/errors.py
```diff
@@ -43,6 +47,11 @@ def _present_clb_deleted_error(exception):
             % (exception.lb_id,))
 
 
+@_present_exception.register(CreateServerConfigurationError)
+def _present_server_configuration_error(exception):
+    return "Server launch configuration is invalid: %s" % (exception.message,)
```
manishtomar: Non-blocking: How come not using `format`?

cyli: I was mostly just following the previous convention of the other `_present_server_configuration_error` things.  I generally prefer `format`, but I wasn't sure if we still follow that.  I'll be happy to change both this and all the rest to use format instead :)

otter/integration/tests/test_convergence.py
cyli: Sounds good!

otter/integration/tests/test_convergence.py
```diff
@@ -989,6 +990,101 @@ def test_recover_from_identity_auth_failures(self):
...
+            min_entities=1, max_entities=10,
+            server_name_prefix="no-such-key",
+            key_name="invalid_key_name"
+        )
+        launch_config = deepcopy(group.group_config['launchConfiguration'])
+
+        if not not_mimic():
+            # if this is mimic, we have to make server creation fail.
+            # in production, it's actually an invalid key and hence would
+            # naturally fail.
```
manishtomar: Non-blocking thought: I wonder if mimic should error in these situations. Does it error on invalid image currently? I always thought mimic to simulate behavior that is not possible through _normal_ APIs. Maybe create ticket in mimic to replicate this?

otter/integration/tests/test_convergence.py
```diff
@@ -989,6 +990,101 @@ def test_recover_from_identity_auth_failures(self):
...
+        yield group.wait_for_state(
+            self.rcs,
+            MatchesAll(
+                ContainsDict({
+                    'pendingCapacity': Equals(0),
+                    'desiredCapacity': Equals(1),
+                    'active': HasLength(1),
+                    'status': Equals("ACTIVE"),
+                }),
+                Not(Contains('error'))
```
manishtomar: Did you mean `errors`?

manishtomar:
LGTM modulo comments :)

otter/integration/tests/test_convergence.py
cyli: Doh!  yes, good catch, thanks!

otter/integration/tests/test_convergence.py
cyli: Mimic doesn't actually implement glance or the key store or anything.  It has some preset invalid images that you can use to fail the server, but it doesn't  have a set of preset keys.  We should probably add that, yeah.

cyli: https://github.com/rackerlabs/mimic/issues/346

cyli:

cyli:
Merging, thanks @manishtomar!

PR #1614, opened 2015-07-22
stop setting the `build_config` servermill option to `core`

radix:
This is what's spamming us with support tickets. Apparently `core` is an invalid option for us to use on the public net so it's not having any effect anyway.

It was also mentioned in all of our documentation and examples for some reason (probably because it got copied from our tests?) and so I went through and removed it from every place that mentioned it.


radix:
I'M A DORK

PR #1615, opened 2015-07-22
handle errors during gathering CLB data

radix:
This handles any `NoSuchCLBError` while gathering, and just treats the nodes for those LBs as non-existent.

One tricky part in particular is that during feed fetching, if we get one of these exceptions, we know that _all_ other feeds/nodes we've fetched for that LB are gone, so the function now throws out all nodes for any LB that raised one of these errors.

Points of note for reviewers:

1. I refactored the CLB requests to cloud_client.py
2. I changed get_clb_contents into a @do-using function, instead of doing manual callbacks. This shortened the code a fair amount, but if reviewers prefer the old style I can very quickly change it back.


radix:

radix:
ok I think I have a real bug, since I'm  getting an error in the integration test that checks for error reasons.

radix:

radix:

radix:

radix:
This is ready for review :)

otter/convergence/gathering.py
```diff
@@ -172,57 +171,46 @@ def get_scaling_group_servers(tenant_id, group_id, now,
...
+    """Get Rackspace Cloud Load Balancer contents as list of `CLBNode`."""
+    # If we get a CLBNotFoundError while fetching feeds, we should throw away
+    # all nodes related to that load balancer, because we don't want to act on
+    # data that we know is invalid/outdated (for example, if we can't fetch a
+    # feed because CLB was deleted, we don't want to say that we have a node in
+    # DRAINING with draining time of 0; we should just say that the node is
+    # gone).
+
+    def gone(r):
+        return catch(CLBNotFoundError, lambda exc: r)
```
cyli: Non-blocking - we can always file an issue or something - do we also want to catch `NoSuchCLBNodeError` here, in case the node itself (but not the CLB) is gone?  This would only probably affect the atom feed, and tbh I don't know if that's the exact error for getting the atom feed of a node.

radix: Yeah, good idea, but it would require more than just adding `NoSuchCLBNodeError` here, since the behavior later in the function would throw out all the nodes for that CLB. I'd be happy to do this in another PR :)

cyli:
Aside from the above nitpick, LGTM!  :)

PR #1617, opened 2015-07-22
unify CLBDeletedError and NoSuchCLBError

radix:
Just deletes CLBDeletedError and replaces it with NoSuchCLBError, since we always treat them the same.


cyli:

PR #1618, opened 2015-07-22
introduce a CLBNotFoundError superclass to CLBDeletedError and NoSuchCLBError

radix:
This makes it simpler to handle "there is no CLB" as one error.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -394,15 +394,20 @@ class CLBImmutableError(Exception):
 
 
 @attributes([Attribute('lb_id', instance_of=six.text_type)])
-class CLBDeletedError(Exception):
+class CLBNotFoundError(Exception):
+    """A CLB doesn't exist. Superclass of other, more specific exceptions."""
+    message = attr.ib(default='')
+    lb_id = attr.ib()
+
+
+class CLBDeletedError(CLBNotFoundError):
     """
     Error to be raised when the CLB has been deleted or is being deleted.
     This is distinct from it not existing.
     """
 
 
-@attributes([Attribute('lb_id', instance_of=six.text_type)])
-class NoSuchCLBError(Exception):
+class NoSuchCLBError(CLBNotFoundError):
     """
     Error to be raised when the CLB never existed in the first place (or it
     has been deleted so long that there is no longer a record of it).
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -396,8 +396,6 @@ class CLBImmutableError(Exception):
 @attributes([Attribute('lb_id', instance_of=six.text_type)])
 class CLBNotFoundError(Exception):
     """A CLB doesn't exist. Superclass of other, more specific exceptions."""
-    message = attr.ib(default='')
-    lb_id = attr.ib()
 
 
 class CLBDeletedError(CLBNotFoundError):
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -20,6 +20,7 @@ from otter.cloud_client import (
     CLBNotFoundError,
     CreateServerConfigurationError,
     CreateServerOverQuoteError,
+    NoSuchCLBNodeError,
     add_clb_nodes,
     create_server,
     has_code,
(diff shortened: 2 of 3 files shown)


radix:

otter/cloud_client.py
```diff
@@ -394,15 +394,20 @@ class CLBImmutableError(Exception):
 
 
 @attributes([Attribute('lb_id', instance_of=six.text_type)])
-class CLBDeletedError(Exception):
+class CLBNotFoundError(Exception):
+    """A CLB doesn't exist. Superclass of other, more specific exceptions."""
+    message = attr.ib(default='')
```
cyli: This is `attr`-specific attribute declaration, but the decorator is `characteristic`-specific, and already declares the `lb_id` type

radix: woops, that was a stray line from when I was messing around with attr.s :) thanks

radix:

radix:

radix:

otter/convergence/steps.py
```diff
@@ -300,8 +297,7 @@ def as_effect(self):
         # Since we're deleting a node, we'll ignore any errors which indicate
         # that the node doesn't exist.
         return eff.on(
-            error=_ignore_errors(
-                CLBDeletedError, NoSuchCLBError, NoSuchCLBNodeError)
+            error=_ignore_errors(CLBNotFoundError)
```
manishtomar: Shouldn't it also check `NoSuchCLBNodeError` also?

radix: Ah! I think you found the reason the integration tests are failing! Yeah, this was a simple typing mistake, I accidentally deleted too much...

radix:

radix:

radix:

radix:
finally :)

otter/test/convergence/test_steps.py
```diff
@@ -469,9 +469,9 @@ def test_add_nodes_to_clb_terminal_failures(self):
         if there is any other 4xx error, then
         the error is propagated up and the result is a failure.
         """
-        terminals = (CLBDeletedError(lb_id=u"12345"),
+        terminals = (CLBNotFoundError(lb_id=u"12345"),
+                     CLBDeletedError(lb_id=u"12345"),
                      CLBNodeLimitError(lb_id=u"12345"),
-                     NoSuchCLBError(lb_id=u"12345"),
```
manishtomar: Maybe these tests should remain the same since in reality specific exceptions will be raised and these tests already check those?

radix: I think it's important to pass in an instance of CLBNotFoundError too since it proves that we handle all of them equally. And I left the CLBDeletedError there to prove that subclasses are honored

manishtomar: In that case, do you think it would help to leave `CLBDeletedError` when checking or would that be duplicate work?

radix: you mean in the steps.py code? the whole point of this change is to make it so we don't have to enumerate all the various ways in which CLBs can disappear... but maybe I'm misunderstanding you

manishtomar: Sorry, I meant in the test code just to be sure that both the errors are handled. I can understand if you consider that duplicate effort knowing the fact the `CLBNotFoundError` is based and it is checked. However, I am just thinking of the test as black box not knowing about the inheritance which is why I suggested no changes in them in the first place. This is just bike-shedding. I am actually fine either way :). This LGTM if you don't change it.

radix: Oh, ok, I was confused because you suggested leaving `CLBDeletedError`, and that's exactly the error which I _did_ leave in the test :-) So I guess you mean I should add `NoSuchCLBError` back, to cover both subclasses and the superclass?

manishtomar: Oh yes. That is what I meant.

otter/test/convergence/test_steps.py
```diff
@@ -609,8 +611,7 @@ def test_remove_nodes_from_clb_success_failures(self):
         :obj:`AddNodesToCLB` succeeds if the CLB is not in existence (has been
         deleted or is not found).
         """
-        successes = (CLBDeletedError(lb_id=u"12345"),
-                     NoSuchCLBError(lb_id=u"12345"))
+        successes = [CLBNotFoundError(lb_id=u'12345')]
```
cyli: Non-blocking - similar to @manishtomar's comment above, do we want to ensure that the subclasses are also counted as success as well as the superclass?

cyli:
Other than the nitpick above, this LGTM!

radix:
merging, thanks :)

Issue #1619, opened 2015-07-23
Better trial test reporting


cyli:
It's hard to debug stuff for trial, so we should do some better reporting in the tests.

- [x] logging-treq
   - ~~easiest to just upgrade to twisted 15.2.1 and use its new logging system so we don't need to mess around with our giant observer chain for the trial tests~~ Removing this, since we want better logging tests in place (so we don't break ES) before doing so
   - [x] create a logging-treq in the tests and pass it to all the test libraries
- ~~instrument mimic to do verbose logging of request/responses (since logging-treq does not log request/response bodies)~~ (https://github.com/rackerlabs/mimic/issues/362)
- [ ] Log more information in tests themselves about what we are trying to do.

PR #1595, opened 2015-07-16
Remove audit log formatter and the history endpoint

cyli:
Since this is all going to be done via cloud feeds now.

PR #1621, opened 2015-07-23
Upgrade to Twisted 15.2.1

### LINKED ISSUE #1580: We cannot upgrade to Twisted 15.2.0, because the new logging system already pre-formats the messages
We see errors such as:

```
2015-07-09_21:40:30.86939 {
2015-07-09_21:40:30.86940   "@timestamp": "2015-07-09T21:40:30.866551", 
2015-07-09_21:40:30.86940   "@version": 1, 
2015-07-09_21:40:30.86941   "cloud_feed": true, 
2015-07-09_21:40:30.86941   "converger_run_id": "1f0f4358-2683-11e5-ade9-000c29b8bc05", 
2015-07-09_21:40:30.86941   "format": "%(log_legacy)s", 
2015-07-09_21:40:30.86941   "host": "vm", 
2015-07-09_21:40:30.86942   "level": 6, 
2015-07-09_21:40:30.86942   "log_format": "{log_text}", 
2015-07-09_21:40:30.86942   "log_legacy": "<twisted.logger._stdlib.StringifiableFromEvent object at 0x41f7050>", 
2015-07-09_21:40:30.86943   "log_level": "<LogLevel=info>", 
2015-07-09_21:40:30.86943   "log_namespace": "log_legacy", 
2015-07-09_21:40:30.86943   "log_system": "otter", 
2015-07-09_21:40:30.86943   "log_text": "convergence-create-servers", 
2015-07-09_21:40:30.86944   "log_time": 1436478030.866551, 
2015-07-09_21:40:30.86944   "message": "Creating 1 with config {'server': pmap({u'flavorRef': u'performance1-1', u'name': u'A', u'imageRef': u'aab63bcf-89aa-440f-b0c7-c7a1c611914b'
, u'metadata': pmap({'rax:autoscale:group:id': u'ddfb3217-5092-49c9-94a2-142c855d5887', u'create_server_failure': u'{\"message\": \"nope\", \"code\": 400}', 'rax:auto_scaling_group
_id': u'ddfb3217-5092-49c9-94a2-142c855d5887'})})}", 
2015-07-09_21:40:30.86944   "message_formatting_error": "[Failure instance: Traceback: <type 'exceptions.KeyError'>: \"'server'\"\n/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/logger/_observer.py:131:__call__\n/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/logger/_legacy.py:93:__call__\n/opt/otter/otter/log/formatters.py:376:emit\n/opt/otter/otter/log/spec.py:138:validating_observer\n--- <exception caught here> ---\n/opt/otter/otter/log/formatters.py:141:PEP3101FormattingObserver\n]", 
2015-07-09_21:40:30.86946   "num_servers": 1, 
2015-07-09_21:40:30.86946   "otter_facility": "otter", 
2015-07-09_21:40:30.86946   "otter_msg_type": "convergence-create-servers", 
2015-07-09_21:40:30.86946   "otter_service": "converger", 
2015-07-09_21:40:30.86947   "scaling_group_id": "ddfb3217-5092-49c9-94a2-142c855d5887", 
2015-07-09_21:40:30.86947   "server_config": {
2015-07-09_21:40:30.86947     "server": "pmap({u'flavorRef': u'performance1-1', u'name': u'A', u'imageRef': u'aab63bcf-89aa-440f-b0c7-c7a1c611914b', u'metadata': pmap({'rax:autoscale:group:id': u'ddfb3217-5092-49c9-94a2-142c855d5887', u'create_server_failure': u'{\"message\": \"nope\", \"code\": 400}', 'rax:auto_scaling_group_id': u'ddfb3217-5092-49c9-94a2-142c855d5887'})})"
2015-07-09_21:40:30.86948   }, 
```

because `otter.log.formatters.PEP3103FormattingWrapper` calls `.format(**eventDict)` on the "message" and the "why" field, but before it can do that, the twisted logging system has already gotten to it and done the same thing.

This causes a problem when JSON or dictionaries get embedded in the message, because then calling `.format` a second time raises `KeyErrors`.

cyli:
This adds a new formatter that copies the event dictionary to all our observer chains.

All our trial tests succeed under twisted 15.2.1.

Will run cafe tests on jenkins.

This adds a couple of new fields to our logs, mostly having to do with twisted noting that we are using the legacy logging system:

- `"log_system": ends up being the same as our `otter_facility`
- `"log_time": "<epoch>"`
- `"log_text": "..."` - This ends up being the same as message
- `"log_format": "{log_text}"` - it's always this value
- `"format": "%(log_legacy)s"` - its' always this value
- `"log_legacy": "<twisted.logger._stdlib.StringifiableFromEvent object at 0x....>"`
- `"log_level": "<LogLevel=...>"`

I don't think any of those new fields conflict with our logging stuff, so it should probably be ok.

Fixes #1580.
Part of #1619.

Ran master on 15.0.0 and compared the logs with this branch on 15.2.1, both with and without CF setup.

Looked at logs with the script in https://github.com/rackerlabs/otter/pull/1595#issuecomment-122376658, and the numbers all seemed to match up more or less, modulo a couple of extra get states from the tests.  And nothing particularly seemed to stand out when I grepped for Traceback.

### DIFF
--- a/otter/log/cloudfeeds.py
+++ b/otter/log/cloudfeeds.py
@@ -17,7 +17,11 @@ from otter.cloud_client import TenantScope, publish_to_cloudfeeds
 from otter.effect_dispatcher import get_legacy_dispatcher
 from otter.log import log as otter_log
 from otter.log.formatters import (
-    ErrorFormattingWrapper, LogLevel, PEP3101FormattingWrapper)
+    ErrorFormattingWrapper,
+    LogLevel,
+    PEP3101FormattingWrapper,
+    copying_wrapper
+)
 from otter.log.intents import err as err_effect, msg as msg_effect
 from otter.log.spec import SpecificationObserverWrapper
 from otter.util.http import APIError
@@ -219,6 +223,7 @@ def get_cf_observer(reactor, authenticator, tenant_id, region,
     cf_observer = CloudFeedsObserver(
         reactor=reactor, authenticator=authenticator, tenant_id=tenant_id,
         region=region, service_configs=service_configs)
-    return SpecificationObserverWrapper(
+    return copying_wrapper(
+        SpecificationObserverWrapper(
             PEP3101FormattingWrapper(
-                ErrorFormattingWrapper(cf_observer)))
+                ErrorFormattingWrapper(cf_observer))))
--- a/otter/log/formatters.py
+++ b/otter/log/formatters.py
@@ -269,7 +269,6 @@ def throttling_wrapper(observer):
         return True
 
     def emit(event):
-        event = event.copy()
         template = _get_matching_template(event)
         if template is not None:
             event_counts[template] += 1
@@ -281,4 +280,14 @@ def throttling_wrapper(observer):
         else:
             return observer(event)
 
+    return emit
+
+
+def copying_wrapper(observer):
+    """
+    An observer that copies the event-dict, so if there is more than one
+    observer chain that mutates events, we don't get any errors.
+    """
+    def emit(event_dict):
+        return observer(event_dict.copy())
     return emit
--- a/otter/log/setup.py
+++ b/otter/log/setup.py
@@ -12,6 +12,7 @@ from otter.log.formatters import (
     StreamObserverWrapper,
     SystemFilterWrapper,
     throttling_wrapper,
+    copying_wrapper,
 )
 from otter.log.spec import SpecificationObserverWrapper
 
@@ -20,17 +21,18 @@ def make_observer_chain(ultimate_observer, indent):
     """
     Return our feature observers wrapped our the ultimate_observer
     """
-    return throttling_wrapper(
-        SpecificationObserverWrapper(
-            PEP3101FormattingWrapper(
-                SystemFilterWrapper(
-                    ErrorFormattingWrapper(
-                        ObserverWrapper(
-                            JSONObserverWrapper(
-                                ultimate_observer,
-                                sort_keys=True,
-                                indent=indent or None),
-                            hostname=socket.gethostname()))))))
+    return copying_wrapper(
+        throttling_wrapper(
+            SpecificationObserverWrapper(
+                PEP3101FormattingWrapper(
+                    SystemFilterWrapper(
+                        ErrorFormattingWrapper(
+                            ObserverWrapper(
+                                JSONObserverWrapper(
+                                    ultimate_observer,
+                                    sort_keys=True,
+                                    indent=indent or None),
+                                hostname=socket.gethostname())))))))
 
 
 def observer_factory():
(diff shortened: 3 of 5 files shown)


cyli:

cyli:

glyph:
This PR basically looks good to me.

manishtomar:
I don't know how but it will be nice if this can be tested against our ELK stack _before_ deploying it in prod.

cyli:
@manishtomar I'm not entirely sure how to either.  We can revert this instead if you prefer.

cyli:
Or rather, I have an idea about how we can test logging, but it's not efficient.  :|  It involves getting the whole ELK stack installed on your local machine, on the dev vm, or another VM, copying over a sample of existing mappings from our real elk stack, and running both mimic and prod tests and pushing dev vm tests to an index created with multiple existing mappings).  

Probably better would be if we just had a default ES mapping that we create every index with in prod, and disable dynamic mappings altogether.  Then we can test this mapping in our automated tests.  That means that every time we make a logging change we have to change this mapping and modify our deployment.  :|

While I think those would be good things to do, I guess it's sorta low priority compared with other stuff at the moment, so I'd understand if we wanted to revert this deal with logging changes later.

glyph:
@manishtomar - This should leave all the existing mappings in place, since the new mapping is just _adding_ keys.  Is there something else that I'm missing about the new logs?  Is adding keys potentially problematic for elasticsearch?

cyli:
@glyph It *should*, but we've had so much trouble with logs it'd be good to verify

glyph:
@cyli I can see why that might be a concern :cry:

glyph:
For what it's worth, a prerelease is also out right now, so this would be a good time to at least run a PR with the prerelease version of Twisted and make sure nothing jumps out.

glyph:
How about this: there are 2 parts to this change, the actual code change, which adds `copying_wrapper`, and the `requirements.txt` bump which upgrades Twisted.  How about we roll back just the `requirements.txt` change, so that we can start testing in the dev VM and such with 15.2.1 and 15.3pre, but leave in the change that *allows* it to work on Twisted, since that is really just a correct change that should have been in place for 15.0 as well?

manishtomar:
We rely heavily on our logs for operational metrics at beginning of every week and we've had some really bad downtimes due to ES mapping issues. If its too much effort/difficult to test before deploy, then we might want to deploy on Monday, _expect_ a downtime, wait for it to roll over the next day and fix it by reindexing.

glyph:
Mainly I want to decouple the issues of 'make it possible to use newer twisted', 'upgrade to newer twisted' and 'disturb the log format'.  To further decouple these, we could write another observer-decorator which simply strips out the new namespaced log keys that @cyli so thoughtfully enumerated in the description; this would have the added benefit of getting `make_observer_chain` past the 10-indents mark.

manishtomar:
> How about we roll back just the requirements.txt change, so that we can start testing in the dev VM and such with 15.2.1 and 15.3pre, but leave in the change that allows it to work on Twisted, since that is really just a correct change that should have been in place for 15.0 as well?

Sounds good. I think it is safe to just add `copying_wrapper`.

PR #1623, opened 2015-07-23
switch the defaults in config.sh-template to those suitable for using mimic on the devvm

radix:
this way we don't need to create a custom config.sh to run the tests


otter/integration/config.sh-template
```diff
@@ -51,11 +51,11 @@ export AS_NOVA_SC_KEY=cloudServersOpenStack
 export AS_CLB_SC_KEY=cloudLoadBalancers
 
 # Uncomment this if you are running tests against Mimic.
```
cyli: Maybe "comment this out if you are not running tests..." - similarly with the next comment?

cyli:
Other than the comment wording, this LGTM!  Thanks - this will streamline setup  :)

PR #1626, opened 2015-07-27
fix unhandled noscalinggroup error

### LINKED ISSUE #1625: NoSuchScalingGroupError unhandled deferred
We seem to be getting lot of `Unhandled error in Deferred: NoSuchScalingGroupError` errors in kibana. These occur when a deferred has errored and it has not been handled. These are logged by twisted when garbage collecting the deferreds. 

These are not harmful yet but may be hiding some bigger problem in the code somewhere.

manishtomar:
Fixes #1625.

Earlier, get_func deferred in gatherResults errored which was propogated above in different deferred. The originial get_func deferred was not handled and resulted in "unhandled error" message. Adding
`consumeError=True` avoids this.

### DIFF
--- a/otter/rest/groups.py
+++ b/otter/rest/groups.py
@@ -487,7 +487,8 @@ class OtterGroup(object):
                 self.group_id)
         else:
             cache_d = succeed(None)
-        return gatherResults([get_func(*args, **kwargs), cache_d])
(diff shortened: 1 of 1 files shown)


cyli:
LGTM, thanks for finding this!  Just ran this with the cafe functional tests, and there are no Unhandled errors in the logs, but there are in master.  :D

PR #1630, opened 2015-07-29
Don't use deepcopy in assoc_obj

radix:
assoc_obj should use shallow copying since if there are things like twisted constants in an object they become uncomparable when copied.

### DIFF
--- a/otter/test/util/test_fp.py
+++ b/otter/test/util/test_fp.py
@@ -4,7 +4,8 @@ from pyrsistent import pmap
 
 from twisted.trial.unittest import SynchronousTestCase
 
-from otter.util.fp import predicate_all, predicate_any, set_in
+from otter.util.fp import (
+    assoc_obj, predicate_all, predicate_any, set_in)
 
 
 class PredicateAllTests(SynchronousTestCase):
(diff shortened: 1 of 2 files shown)


manishtomar:
`assoc_obj` is used to copy `GroupState` which has dict attributes. I am not sure if its a good idea to default it to normal copy. Or maybe use something else to copy `GroupState`s.

radix:
@manishtomar yeah, but only to get the effect of changing the one attribute. I looked at that code and don't see any problem with using a shallow copy for it -- the returned group state is very short-lived, just to write to the db.

I'll run integration tests to be safe.

radix:

radix:

manishtomar:
FWIW, I think its fine w.r.t that _one_ time `assoc_obj` is used to update `GroupState.desired`. I just don't feel comfortable changing the semantics of an existing function. Having default deepcopy sounds like a better choice. How about having separate `shallow_assoc_obj` function that does shallow copying instead?

radix:
I disagree. It's not like there's a lot of users of the function now. It should be expected that assoc_obj only does a shallow copy. I think needing a deepcopy should be a much less common case. I can update the docstring of assoc_obj to explicitly indicate that it only does a shallow copy so it's more clear.

manishtomar:
Hmm.. in that case, having a clear doc for `assoc_obj` should be fine. Otherwise, LGTM

radix:
thanks, done :)

PR #1632, opened 2015-07-29
Use SequenceDispatcher in CLB gathering tests

radix:
This makes them drastically more clear IMO. Also did some other cleanups to let us specify the test data more tersely.

FYI This is in preparation for #1615 which does some further refactoring/changes to the CLB gathering code.

### DIFF
--- a/otter/test/convergence/test_gathering.py
+++ b/otter/test/convergence/test_gathering.py
@@ -16,6 +16,8 @@ from effect.async import perform_parallel_async
 from effect.testing import (
     EQDispatcher, EQFDispatcher, Stub)
 
+import mock
+
 from pyrsistent import freeze
 
 from toolz.curried import map
@@ -24,14 +26,14 @@ from toolz.functoolz import compose
 from twisted.trial.unittest import SynchronousTestCase
 
 from otter.auth import NoSuchEndpoint
-from otter.cloud_client import service_request
+from otter.cloud_client import CLBDeletedError, service_request
 from otter.constants import ServiceType
 from otter.convergence.gathering import (
     UnexpectedBehaviorError,
     extract_CLB_drained_at,
     get_all_convergence_data,
-    get_all_server_details,
     get_all_scaling_group_servers,
+    get_all_server_details,
     get_clb_contents,
     get_rcv3_contents,
     get_scaling_group_servers,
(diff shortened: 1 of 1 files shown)


otter/test/convergence/test_gathering.py
```diff
@@ -419,145 +424,128 @@ def test_invalid_first_entry(self):
...
-        body = self.reqs[(req.method, req.url, req.json_response)]
-        fake_response = object()
-        return resolve_effect(eff, (fake_response, body))
-
-    def _resolve_lb(self, eff):
-        """Resolve the tree of effects used to fetch LB information."""
-        # first resolve the request to list LBs
-        lb_nodes_fetch = self._resolve_request(eff)
-        if type(lb_nodes_fetch) is not Effect:
-            # If a parallel effect is *empty*, resolve_stubs will
```
radix: SO happy to get rid of this horrible hack!

otter/test/convergence/test_gathering.py
```diff
@@ -419,145 +424,128 @@ def test_invalid_first_entry(self):
...
+        ])
+    )
+
+
+def nodes_req(lb_id, nodes):
+    return lb_req('GET', 'loadbalancers/{}/nodes'.format(lb_id),
+                  True, {'nodes': nodes})
+
+
+def feed_req(lb_id, node_id, response):
```
cyli: Non-blocking - since we also deal with cloud feeds, and may in the observer in the future, for clarity can we name this `node_feed_req` instead?

otter/test/convergence/test_gathering.py
```diff
@@ -419,145 +424,128 @@ def test_invalid_first_entry(self):
         self.assertRaises(ValueError, extract_CLB_drained_at, feed)
 
 
+def lb_req(method, url, json_response, response):
```
cyli: Totally non-blocking - since this method returns a 200 every time, which is usually a response code to a GET request (since the PUTs and POSTs usually return 202 or 201 or something), can we just assume whatever calls this method is going to call it with a GET?

otter/test/convergence/test_gathering.py
cyli: Yay!

otter/test/convergence/test_gathering.py
radix: Good point! I'll update it.

radix: And while I'm doing that, I'll also always add on `loadbalancers` as the first part of the URL.

radix: hmm. or not, since append_segments('loadbalancers', '') returns 'loadbalancers/' which breaks the tests. oh well.

cyli:
Yay 103 deletions and 92 additions!  :D  Thanks for cleaning this up.   LGTM!

PR #1634, opened 2015-07-29
servers cache queries use timestamp

### LINKED ISSUE #1627: convergence view_state buggy?
`test_delete_server_not_found_in_diff_group` test in https://as-jenkins.rax.io/job/otter-cloudcafe-syd-convergence-system-tests/53/console failed because the group created with min entities 1 never showed 1 active server. But as per the [logs](https://dev-otterlogs.fedsso.rackspace.com/#dashboard/temp/adyJ8bTsTO-Hze8sFP0C3A), the server did become active and was updated in cache. But getting state on the group never showed the active server. 

I am not sure where the problem is. Did CASS didn't actually update it? Did the view state returned correctly but test didn't see it for some reason? The latter doesn't seem true since test logs shows otherwise. 

I am temporarily creating issue here for tracking purpose. Might debug if it occurs again.

manishtomar:
Fixes #1627.

Since delete and insert are really close in time, there is chance that delete can take precedence over insert. We avoid this by sending timestamp from client which should be serial between the queries. Thanks to @cyli for suggesting this.

It would be nice to provide real reactor to `CassScalingGroupServersCache` but it seems difficult in current code base since that class's object is created in gathering on the fly.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -600,9 +600,9 @@ class WeakLocks(object):
 
 def get_client_ts(reactor):
     """
-    Return EPOCH with microseconds precision as a `Deferred`
+    Return EPOCH with microseconds precision synchronously
     """
-    return defer.succeed(int(reactor.seconds() * 1000000))
+    return int(reactor.seconds() * 1000000)
 
 
 def _check_deleting(group, get_deleting=False):
@@ -694,7 +694,7 @@ class CassScalingGroup(object):
         """
         @functools.wraps(func)
         def wrapper(*args):
-            d = get_client_ts(self.reactor)
+            d = defer.maybeDeferred(get_client_ts, self.reactor)
             d.addCallback(lambda ts: func(ts, *args))
             return d
         return wrapper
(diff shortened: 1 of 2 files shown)


manishtomar:

cyli:

otter/models/cass.py
```diff
@@ -1789,17 +1794,18 @@ def insert_servers(self, last_update, servers, clear_others):
             queries.append(query.format(cf=self.table, i=i))
         if clear_others:
             return self.delete_servers().on(
-                lambda _: cql_eff(batch(queries), params))
+                lambda _: cql_eff(
+                    batch(queries, self.clock.seconds()), params))
         else:
-            return cql_eff(batch(queries), params)
+            return cql_eff(batch(queries, self.clock.seconds()), params)
```
cyli: Was there some weirdness with Cassandra and the format of the timestamps needing to be milliseconds?  I think elsewhere we use https://github.com/rackerlabs/otter/blob/master/otter/models/cass.py#L601, but I don't remember the exact details.

manishtomar: Great catch @cyli. This should be microseconds. See [docs](http://docs.datastax.com/en/cql/3.0/cql/cql_reference/delete_r.html). Will change it.

manishtomar:

otter/models/cass.py
```diff
@@ -694,7 +694,7 @@ def with_timestamp(self, func):
         """
         @functools.wraps(func)
         def wrapper(*args):
-            d = get_client_ts(self.reactor)
+            d = defer.maybeDeferred(get_client_ts, self.reactor)
```
cyli: Non-blocking - since it's a synchronous function, can this just call `func(get_client_ts(self.reactor), *args)` and return the result?  It looks like all the functions that are decorated returned Deferreds anyway, and the docstring does not say that this will always wrap the function result in a Deferred.

cyli:
Other than the nitpick, this LGTM!

otter/models/cass.py
manishtomar: Sounds good :)

cyli:
+1 :)

PR #1636, opened 2015-07-30
fix unnecessary retry in delete server

manishtomar:
instead of `servers/id/details` which is incorrect and will always return 404 with non-json body. Thanks a lot to @cyli for finding this problem. Also reduced retry times since convergence will be retried anyways.


cyli:
LGTM pending jenkins +1 :)

cyli:

radix:
Yay!

Issue #1635, opened 2015-07-29
Logging responses


cyli:
There was some discussion about `logging_treq` logging responses for easier debugging in #1619 and #1629.

One requirement is that for otter prod, since we *definitely* need to not log some responses, or filter out some responses.  For instance, we do not want to log identity auth responses.  We don't want to log impersonation responses.  We do not want to log server root passwords when we create a server.  One suggestion from @glyph is that we encrypt all our logged responses, so we can't view them until there is some need to debug, but that might be pretty slow.

When discussing this more with @glyph and @radix, it turns out that we only want to log the success response, since APIErrors are already logged with bodies.  And we probably want to filter out certain JSON fields (like the password) and not just have a bunch of regexes.

`logging_treq` doesn't know about the content, it only logs the response itself.  `cloud_client` knows about all the different operations we're making, and what kind of data it should be getting back.

So 2 possibilities so far:

1. `cloud_client` does some extra logging of  filtered responses, in addition to the logging done by `logging_treq`.  This has the added bonus that we can truncate data (for instance, when listing servers we can log stats, and abbreviated server data, rather than a huge blob of all server details).

1. `logging_treq` automatically logs all responses, and filtering is done by matching the URL/method against regex, to determine what if any filtering is needed.  This can be done by our log observers, or by `logging_treq`.  This has the advantage of everything being logged, but that we have to re-constitute what operations are being performed (which `cloud_client` already knows) and what data needs to be synthesized given the response object and body.

`logging_treq` doing the logging also makes it usable by the trial tests.  On the other hand, we wanted to move toward using `cloud_client` on the trial tests anyway due to the retrying logic we already have in place with otter, and when hitting Mimic at least, we can instrument mimic to log all requests and responses.

manishtomar:
I like 1. `cloud_client` functions can provide filters depending on the kind of requests it is making. The filters can be separately written functions in which one of them could be log everything that can be used by tests until it switches to `cloud_client`.

cyli:
Sounds good.  :D  I'm leaning towards that.  I'm thinking:

- [ ] propagate the treq request ID as a header in all our requests?  I think that addresses part of #843, and then we can link the cloud client logs with the treq logs
- [ ] add a filtering and logging thing for create server
- [ ] move some of the gathering Nova functions to `cloud_client` and add logging for them too

Issue #843, opened 2014-12-17
transaction id header in requests to Nova/CLB


manishtomar:
Add request id header from otter's transaction id when sending request to Nova/CLB. This allows us to correlate requests from otter to Nova.

manishtomar:
Right now in worker, otter has 2 ids associated with each request:
1. Per request ID: Every HTTP request sent from otter to Nova/CLB will have unique ID. This can be sent as `otter-request-id` header when talking to Nova/CLB.
2. Per transaction ID: A transaction corresponds to HTTP request received. For example, policy execution request generates a transaction ID that gets sent to all the logs that generate from that request. These logs include creating/deleting servers/CLB nodes. This can be sent as `otter-transaction-id` header when talking to Nova/CLB. Scheduled policy works differently. Each schedule run has an ID that can be sent as transaction id header to HTTP request sent during scheduled policy executions in that run. Please note that there can be different tenant's different groups getting executed in same schedule run. This ID would be different in convergence. Each converge run can have ID that can correspond to this ID. Does this make sense?

cyli:
I have no idea why I missed the second comment.  Yes that sounds reasonable.

cyli:
Update:  the per-request ID is done.  The transaction ID has not been done yet, since at least with the converger, it may not be part of a transaction, but a convergence run.

PR #1637, opened 2015-07-30
Propagate treq request ID as a header

cyli:
This addresses part of #843 and gives us a treq transaction ID to trace additional log messages from `cloud_client` logs, which will be done as part of #1635.

### DIFF
--- a/otter/util/logging_treq.py
+++ b/otter/util/logging_treq.py
@@ -38,9 +38,6 @@ def _log_request(treq_call, url, **kwargs):
                    treq_request_id=treq_transaction)
     start_time = clock.seconds()
 
-    kwargs.setdefault('headers', {})
-    kwargs['headers'].setdefault('x-otter-request-id', [treq_transaction])
-
     log.msg("Request to {method} {url} starting.")
     d = treq_call(url=url, **kwargs)
 
--- a/otter/util/logging_treq.py
+++ b/otter/util/logging_treq.py
@@ -30,8 +30,8 @@ def _log_request(treq_call, url, **kwargs):
     method = kwargs.get('method', treq_call.__name__)
 
     kwargs.setdefault('headers', {})
-    treq_transaction = kwargs['headers'].setdefault('x-otter-request-id',
-                                                    [str(uuid4())])[0]
+    treq_transaction = str(uuid4())
+    kwargs['headers']['x-otter-request-id'] = [treq_transaction]
 
     log = log.bind(system='treq.request', url=url, method=method,
                    url_params=kwargs.get('params'),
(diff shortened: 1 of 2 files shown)


manishtomar:
Can we have a test that calls treq with headers and ensures that existing header is not changed before sending?

cyli:
@manishtomar Yes, good idea, thanks!

otter/util/logging_treq.py
```diff
@@ -24,16 +24,23 @@ def _log_request(treq_call, url, **kwargs):
...
     log = kwargs.pop('log', None)
+
     if not log:
         log = default_log
     method = kwargs.get('method', treq_call.__name__)
 
-    treq_transaction = str(uuid4())
+    kwargs.setdefault('headers', {})
+    treq_transaction = kwargs['headers'].setdefault('x-otter-request-id',
+                                                    [str(uuid4())])[0]
```
manishtomar: Non-blocking nitpick: This is a little confusing. How about
```python
kwargs.setdefault('headers', {})
treq_transaction = str(uuid.uuid4())
kwargs['headers'].setdefault('x-otter-request-id', [treq_transaction])
```

otter/util/logging_treq.py
```diff
@@ -24,16 +24,23 @@ def _log_request(treq_call, url, **kwargs):
...
+
     log = log.bind(system='treq.request', url=url, method=method,
                    url_params=kwargs.get('params'),
                    treq_request_id=treq_transaction)
     start_time = clock.seconds()
 
+    kwargs.setdefault('headers', {})
+    kwargs['headers'].setdefault('x-otter-request-id', [treq_transaction])
```
manishtomar: This is already set above right? Why is this set again? Sorry, I might be missing something here.

otter/util/logging_treq.py
cyli: The intention was to actually uses the request ID as the treq transaction ID, if it is provided (if we wanted to overload it.

So possibly:

```python
kwargs.setdefault('headers', {})
treq_transaction = str(uuid.uuid4())
treq_transaction = kwargs['headers'].setdefault('x-otter-request-id', [treq_transaction])[0]
```

Alternately, we can just always overwrite this header, no matter what the user provides (I can't think of a case where we might want to respect this ID after all).

otter/util/logging_treq.py
cyli: Oh that's weird - this should have been deleted.  I must have messed up my conflict resolution from stashing, thanks!

otter/util/logging_treq.py
manishtomar: Oh. Didn't realize about the update. I am fine either way. The above code that honors user given id or ignoring it. Any one is fine as long as it is clear :) I am leaning towards ignoring user provided for now.

cyli: Sounds good - if we actually come up with a use case we can try to preserve it.  Will remove this and add in a test :D

manishtomar:
LGTM! Merging :)

PR #1638, opened 2015-07-30
Log server create response

cyli:
First attempt at logging responses from `cloud_client`, as per #1635.  Also associating these logs by getting the request ID put into the request in #1637.

Does this look like a good place to put the logging?

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -29,6 +29,7 @@ from txeffect import deferred_performer, perform as twisted_perform
 
 from otter.auth import Authenticate, InvalidateToken, public_endpoint_url
 from otter.constants import ServiceType
+from otter.log.intents import msg as msg_effect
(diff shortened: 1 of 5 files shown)


cyli:
jenkisn please run integration tests

manishtomar:

radix:
You'll need to add an entry to spec.py for `request-create-server`. Other than that, this looks good to me :)

radix:
(the reason this is necessary isn't just to get the fancy english-text message, but the existence of an entry in spec.py properly maps the `request-create-server` string into the `otter_msg_type` field so we can filter on it properly)

cyli:
@radix: oh right!  thanks, I totally forgot about the filtering.  :)

cyli:

radix:
LGTM :)

PR #1639, opened 2015-07-31
Fix bug where if headers=None is passed, propagating the treq ID fails

cyli:
Because 'headers' DOES have a default value it's just None.


radix:
Dang. LGTM :)

PR #1640, opened 2015-07-31
Group pause implementation

### LINKED ISSUE #1604: Implement pause
The endpoint is already implemented (https://github.com/rackerlabs/otter/blob/master/otter/rest/groups.py#L665), but the functionality is not (https://github.com/rackerlabs/otter/blob/master/otter/controller.py#L84).

As per discussions, "paused" state and "status" are different - a group can both be paused and in error.

If in error, a group will stop it's current converge cycle, but any converging-action taken by the user (group config change, a policy execution, a scheduled policy execution, a request to the converge endpoint) will cause the group to converge again, which will either fix the group or leave it in ERROR again.

(Question: will adding the observer mean that changes to the tenant's servers also cause the group to converge if the group is in ERROR?)

If the group is paused, a group will actively prevent all future convergences and policy executions.  Scheduled policies will not execute.  Regular policies will not execute.  Changes observed by the observer will not cause a convergence (?).

I believe the outcome was:

1. Config changes will be accepted, but not acted upon (no converging)
2. Launch config changes will be accepted, but not acted upon (no converging)
3. Policy creations/deletions will be accepted, but not acted upon (no converging, no scheduled executions).

But please confirm the above 3 points.

Tasks:
- [ ] raise `CannotExecutePolicyError` when attempting to execute a policy on a paused group (this should cause the execute policy endpoint to 403)
- [ ] Return 403 on converge endpoint (only for convergence tenants) when attempting to converge a paused group
- [ ] New convergence cycles should abort if the group is paused
- [ ] After a convergence cycle ends, the dirty flag should be deleted if the group is in paused state (?)
- [ ] Integration test for pause: 
   - Once paused, executing a policy: 403
   - Once paused, hitting convergence endpoint: 403
   - Once paused, creating a scheduled policy works but it never runs
   - Once paused, changing the launch config and group config both work
   - Once paused, deleting the group also works (?)
   - Not sure what else, because it's hard to test that convergence stops, since it's hard to distinguish between convergence stopping and convergence not stopping but taking a really long time.

manishtomar:
Implements #1604.

Sorry this is big. It also contains integration tests and some changes to get that working. I will be splitting this up into multiple PRs:
- [ ] Integration tests: #1642
- [ ] Intent to modify group state paused: #1643
- [ ] `delete_divergent_flag` ignores `NoNodeError`: #1644
- [ ] Remaining stuff: Mainly REST and controller changes

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -27,6 +27,7 @@ from functools import partial
 
 from effect import (
     Effect,
+    parallel,
     parallel_all_errors)
 
 from effect.do import do, do_return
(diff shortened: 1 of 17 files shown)


manishtomar:
Closing in preference to smaller PRs

PR #1641, opened 2015-08-01
run all trial tests

manishtomar:
by making otter.integration.tests a module. Changed the Makefile to make unittests from `otter.test`. This way we can run all trial tests by running `trial otter.integration.tests`


cyli:
LGTM!

PR #1642, opened 2015-08-02
Group pause integration tests

manishtomar:
First PR of #1640

### DIFF
--- a/otter/integration/tests/test_pause.py
+++ b/otter/integration/tests/test_pause.py
@@ -13,12 +13,14 @@ from otter.integration.lib.mimic import MimicNova
 from otter.integration.lib.resources import TestResources
 from otter.integration.lib.trial_tools import (
     TestHelper,
+    convergence_interval,
     get_identity,
     get_resource_mapping,
     get_utcstr_from_now,
-    otter_build_timeout,
+    not_mimic,
     region,
     scheduler_interval,
+    skip_if,
     sleep
 )
 
@@ -56,6 +58,7 @@ class PauseTests(unittest.TestCase):
         yield self.helper.assert_group_state(
             group, ContainsDict({"paused": Equals(True)}))
 
+    @skip_if(not_mimic, "This requires mimic for long builds")
     @inlineCallbacks
     def test_pause_stops_convergence(self):
         """
@@ -67,12 +70,13 @@ class PauseTests(unittest.TestCase):
         5. Notice that group continues to think that server is building
         """
         mimic_nova = MimicNova(pool=self.helper.pool, test_case=self)
+        server_build_time = convergence_interval + 5
         yield mimic_nova.sequenced_behaviors(
             self.rcs,
             criteria=[{"server_name": "pause-stops-convergence" + ".*"}],
             behaviors=[
                 {"name": "build",
-                 "parameters": {"duration": otter_build_timeout - 5}}
+                 "parameters": {"duration": server_build_time}}
             ])
         group, _ = self.helper.create_group(min_entities=1)
         yield group.start(self.rcs, self)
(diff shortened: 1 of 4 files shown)


otter/integration/tests/test_pause.py
```diff
@@ -0,0 +1,159 @@
...
+            group, ContainsDict({"paused": Equals(True)}))
+
+    @inlineCallbacks
+    def test_pause_stops_convergence(self):
+        """
+        Pausing a group will stop any further convergence cycle. We do this by
+        1. Setup mimic to build server for just before timeout
+        2. Creating a group with 1 min entity
+        3. Pause the group
+        4. Wait for server to finish building in mimic
```
cyli: Should this test only be run in the mimic environment?  It seems like with a few changes, it could also work in production.

1.  If it's mimic only, this test should probably have a `@skip_if` decorator.
2.  If we want this to also work in production, maybe we can poll Nova to wait until Nova says the server is active, then wait some more time for a convergence cycle to kick off, and then assert something about the state?

Possibly even on mimic though this could spuriously pass, if pause does not stop convergence cycles:  If the server on mimic goes active at time `t=otter_build_timeout - 5`, and a convergence cycle happened right before that, then if we check the state at time `t=otter_build_timeout`, another convergence cycle may not have happened yet (but will) and hence it'd still say that the serve was building.  

To be sure that convergence cycles have stopped, should we wait at least 1 convergence cycle of time?

cyli:
Do we also want to add a test that ensures that creating a webhook succeeds even if the group is paused, and a test that executing a webhook will fail to execute the policy (even though it returns 202)?

otter/integration/tests/test_pause.py
```diff
@@ -0,0 +1,187 @@
...
+        # execute webhook and wait for sometime for execution to take place
+        resp = yield treq.post(webhook.capurl, pool=self.helper.pool)
+        self.assertEqual(resp.code, 202)
+        yield treq.content(resp)
+        yield sleep(reactor, 2)
+        # The group has no impact
+        yield self.helper.assert_group_state(
+            group,
+            ContainsDict({"pendingCapacity": Equals(0),
+                          "activeCapacity": Equals(0)}))
```
cyli: Should this also check `desiredCapacity` to assert that there's been no change?  I know that a change to `desiredCapacity` would be reflected in `pendingCapacity`, since the pending is now just (desired - active), but it might be nice to check all 3 because desired is the actual number that gets changed when a policy gets executed, and also if there's ever a bug in the pending/active/desired calculation this would catch it.  :)

otter/integration/tests/test_pause.py
```diff
@@ -0,0 +1,187 @@
...
+            criteria=[{"server_name": "pause-stops-convergence" + ".*"}],
+            behaviors=[
+                {"name": "build",
+                 "parameters": {"duration": server_build_time}}
+            ])
+        group, _ = self.helper.create_group(min_entities=1)
+        yield group.start(self.rcs, self)
+        one_building = ContainsDict({"pendingCapacity": Equals(1),
+                                     "activeCapacity": Equals(0),
+                                     "status": Equals("ACTIVE")})
```
cyli: Similarly to https://github.com/rackerlabs/otter/pull/1642/files#r36238678, would it make sense to also add a `desiredCapacity` check here?

cyli:
I think this LGTM aside from the nitpicks :D

otter/integration/tests/test_pause.py
manishtomar: Sounds good :)

manishtomar:
@cyli Jenkins test are not triggered and these will not run since they are skipped.

cyli:
@manishtomar Er, apologies I meant travis.  :)

cyli:
Speaking of which, I'll close and open and hope travis doesn't hang again this time.  :|

cyli:
Yay!  Travis finally finished.  LGTM.

manishtomar:
Thanks @cyli :) Merging.

PR #1643, opened 2015-08-02
intent to modify group state paused

manishtomar:
Second PR of #1640

### DIFF
--- a/otter/test/models/test_intents.py
+++ b/otter/test/models/test_intents.py
@@ -10,10 +10,11 @@ from twisted.trial.unittest import SynchronousTestCase
 
 from otter.log.intents import get_log_dispatcher
 from otter.models.intents import (
-    DeleteGroup, GetScalingGroupInfo, UpdateGroupErrorReasons,
-    UpdateGroupStatus, UpdateServersCache, get_model_dispatcher)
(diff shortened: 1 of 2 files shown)


otter/test/models/test_intents.py
```diff
@@ -153,3 +156,15 @@ def test_perform_update_error_reasons(self):
...
         self.assertIsNone(sync_perform(dispatcher, Effect(intent)))
         self.group.update_error_reasons.assert_called_once_with(['r1', 'r2'])
+
+    def test_modify_group_state_paused(self):
+        dispatcher = self.get_dispatcher(self.get_store())
+        r = sync_perform(dispatcher,
+                         Effect(ModifyGroupStatePaused(self.group, False)))
+        self.assertIsNone(r)
+        # Returned state has updated paused
+        modified_state = self.group.modify_state_values[-1]
```
cyli: Do we want to add a sanity check here that `self.state != modified_state`, just to ensure that they're two different objects and not the same object that's been mutated?

manishtomar: Sounds good :)

cyli:
LGTM, thanks!  :D

PR #1644, opened 2015-08-02
delete_divergent_flag ignores NoNodeError

manishtomar:
Refactored to use do generator. 

3rd PR of #1640.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -11,11 +11,11 @@ from datetime import datetime
 from functools import partial
 from hashlib import sha1
 
-from effect import Effect, FirstError, Func, catch, parallel
+from effect import Effect, FirstError, Func, parallel
 from effect.do import do, do_return
 from effect.ref import Reference
 
-from kazoo.exceptions import BadVersionError
+from kazoo.exceptions import BadVersionError, NoNodeError
 from kazoo.recipe.partitioner import PartitionState
 
 from pyrsistent import pset
@@ -316,6 +316,7 @@ def mark_divergent(tenant_id, group_id):
     return eff
 
 
+@do
 def delete_divergent_flag(tenant_id, group_id, version):
     """
     Delete the dirty flag, if its version hasn't changed. See comment in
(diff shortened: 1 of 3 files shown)


radix:
Do you have an idea of why this is happening?

manishtomar:
This doesn't happen now. It may happen when a group is paused at which time it tries to delete divergent flag. Sorry if this wasn't clear in #1640.

otter/convergence/service.py
```diff
@@ -325,17 +326,20 @@ def delete_divergent_flag(tenant_id, group_id, version):
...
         # BadVersionError shouldn't be logged as an error because it's an
         # expected occurrence any time convergence is requested multiple times
         # rapidly.
-        error=catch(
-            BadVersionError, lambda e: msg('mark-clean-skipped',
-                                           path=path, dirty_version=version))
-    ).on(
-        error=lambda e: err(exc_info_to_failure(e), 'mark-clean-failure',
-                            path=path, dirty_version=version))
+        yield msg('mark-clean-skipped', **fields)
```
cyli: Yay this is easier to read, thanks!

cyli:
Oh I see, because there may be two attempts to delete the divergence flag, one by pause, and then one if an existing convergence cycle (which is unaffected by pause) finishes?

I think this PR needs to add `mark-clean-not-found` to `otter/log/spec.py` so it will actually get logged rather than filtered out, but otherwise I think it LGTM.   Or at least I can't think of a case where this would cause a problem, since we log if that happens so it's not just disappearing.  @radix?

manishtomar:
> Oh I see, because there may be two attempts to delete the divergence flag, one by pause, and then one if an existing convergence cycle (which is unaffected by pause) finishes?

That could be one case but the group pause implementation in #1646 tries to delete divergent not knowing or concerned about any current convergence cycle. The group could be silent (no convergence activity) or it could be running convergence. In either case, deleting the flag will stop convergence from continuing. 

> I think this PR needs to add mark-clean-not-found to otter/log/spec.py so it will actually get logged rather than filtered out

Oh yes. Thanks for catching this. Will add it :)

cyli:
LGTM!

PR #1646, opened 2015-08-03
rest of group pause implementation

manishtomar:
Final PR of #1640. Please review this after merging #1643 and #1644. This does following:

- pausing the group by updating group state paused and deleting divergent flag
- returning 403 on REST APIs that should not execute on paused group

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -27,6 +27,7 @@ from functools import partial
 
 from effect import (
     Effect,
+    parallel,
     parallel_all_errors)
 
 from effect.do import do, do_return
@@ -48,10 +49,12 @@ from otter.cloud_client import (
     set_nova_metadata_item)
 from otter.convergence.composition import tenant_is_enabled
 from otter.convergence.model import DRAINING_METADATA, group_id_from_metadata
-from otter.convergence.service import get_convergence_starter
+from otter.convergence.service import (
+    delete_divergent_flag, get_convergence_starter)
 from otter.json_schema.group_schemas import MAX_ENTITIES
 from otter.log import audit
-from otter.models.intents import GetScalingGroupInfo
+from otter.log.intents import with_log
+from otter.models.intents import GetScalingGroupInfo, ModifyGroupStatePaused
 from otter.models.interface import GroupNotEmptyError, ScalingGroupStatus
 from otter.supervisor import (
     CannotDeleteServerBelowMinError,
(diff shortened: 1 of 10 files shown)


manishtomar:
Closing and reopening to trigger travis again.

radix:
should the 'in progress' tag be removed?

manishtomar:
@radix Yes. Done :)

otter/controller.py
```diff
@@ -91,7 +118,10 @@ def pause_scaling_group(log, transaction_id, scaling_group):
 
     :return: None
     """
-    raise NotImplementedError('Pause is not yet implemented')
+    if not tenant_is_enabled(scaling_group.tenant_id, config_value):
+        raise NotImplementedError("Pause is not yet implemented")
```
radix: Probably better to make this say "Pause is not implemented for legacy groups" or something, since "yet" implies it will be

otter/controller.py
```diff
@@ -174,6 +204,10 @@ def delete_group(log, trans_id, group, force):
     """
 
     def check_and_delete(_group, state):
+        if state.paused:
```
radix: last discussion I remember I *think* we concluded that it should just be fine to delete a paused group; is there some specific reason we did it this way?

otter/test/test_controller.py
```diff
@@ -46,9 +51,63 @@
...
+                     (ModifyGroupStatePaused(self.group, True), noop),
+                     (DeleteNode(path="/groups/divergent/tid_gid", version=-1),
+                      noop),
+                     (Log("mark-clean-success", {}), noop)
+                 ])
+             ]))
+        ]
+        self.assertEqual(perform_sequence(seq, eff), None)
+
+    @mock.patch("otter.controller.conv_pause_group_eff",
```
radix: It seems the mock here is buying us very little, since I think it should be pretty easy to check the real intents used. You could factor these tests so that they share an intent-sequence.

radix:
Nice refactoring :) LGTM

otter/controller.py
manishtomar: I am sorry for miscommunication but we [thought](https://github.com/rackerlabs/otter/issues/1604#issuecomment-125747741) of allowing only force deletion of a paused group. I am not sure why but I thought it is inappropriate to normally allow deleting a paused group. On second thoughts, it should be fine to just delete the group if it doesn't have servers.

otter/controller.py
manishtomar: Good point. Will change it.

otter/test/test_controller.py
manishtomar: I wouldn't completely agree with you. The intention of this test is to check that it is performing _any_ effect returned from `conv_pause_group_eff` and returning its result which it does. I do realize that avoiding mock in any case is better but I am slightly leaning towards this since this only unit tests `pause_scaling_group` function. If you feel still feel strongly about it, I can change it :)

otter/controller.py
manishtomar: @cyli Thoughts?

cyli: I'm in favor of allowing deletes during pause.

manishtomar: Ok cool :) I'll change it allow deletes during pause.

manishtomar:

PR #1647, opened 2015-08-03
Skip test scaling

cyli:
Skip this test for now while we decide whether this should be fixed up or removed (since there is cloudcafe coverage for scaling up and down)


manishtomar:
LGTM! Merging.

PR #1650, opened 2015-08-04
Add an argument to logging-treq to log responses for tests

cyli:
This is intended to be used only for the trial tests, because it doesn't do any filtering, and I didn't want to write a different logging treq just for the tests.

The otter codebase will just log responses in the `cloud_client.py`.

I moved most of the `logging_treq` logic into an object because I wanted to be able to pass around a treq with the log-type arguments already bound in the trial tests.

### DIFF
--- a/otter/test/test_logging_treq.py
+++ b/otter/test/test_logging_treq.py
@@ -2,9 +2,10 @@
 Tests for logging treq
 """
 import mock
+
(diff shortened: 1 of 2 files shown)


cyli:

radix:
Nice, LGTM :)

PR #1651, opened 2015-08-04
Add another file log observer for the trial tests

cyli:
There is currently no good way to replace the log observer that trial uses, because I think we'd have to get ahold of the current instance of the tests (I'm sure we can do this from a test module), obtain its observer, remove it from the global log publisher, and then replace it with our own.

For less grossness, I'm just adding a log observer at the start of every test that writes to a file, and closing it after the test.

So the `_trial_temp` directory will look like:

```
_trial_temp
\_ #                               <-- process number when running with -j#
   \_ test.log                     <-- where normal trial logs go
   \_ test_module_name.TestCaseName.test_method_name1.log
   \_ test_module_name.TestCaseName.test_method_name2.log
```

The logs we write are all in separate files, because for some reason the test name demarcation that appears in `test.log` doesn't seem to appear in our logs.

Part of #1619

### DIFF
--- a/otter/integration/lib/trial_tools.py
+++ b/otter/integration/lib/trial_tools.py
@@ -1,7 +1,7 @@
 """
 A set of helpers for writing trial tests
 """
-
+import json
 import os
 
 from testtools.matchers import (
@@ -14,6 +14,8 @@ from testtools.matchers import (
 from twisted.internet import reactor
 
 from twisted.internet.defer import gatherResults, inlineCallbacks, returnValue
+
+from twisted.python.log import addObserver, removeObserver
 
 from twisted.web.client import HTTPConnectionPool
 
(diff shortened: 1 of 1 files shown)


cyli:

otter/integration/lib/trial_tools.py
```diff
@@ -98,6 +107,66 @@ def get_resource_mapping():
     return res
 
 
+def filter_logs(observer):
+    """
+    Filter out logs like
+    "Starting factory <twisted.web.client_HTTP11ClientFactory".
+    """
+    def filter(eventdict):
```
manishtomar: Non-blocking nitpick: Could you rename this to `observer` or `filter_starting_messages` since `filter` is a predefined function?

cyli: Good point, thanks!

manishtomar:
Code LGTM however I am getting formatting errors when testing on dev vm. Is that expected?

cyli:
@manishtomar In otter logs or in trial logs?

cyli:
@manishtomar Oh yeah, that's because I'm logging dicts in the wait_for_state stuff.  :|  Yeah, that's fine, although not ideal for now.

manishtomar:
Cool. No problem :). This is good to go :+1:

PR #1654, opened 2015-08-05
use uuid4() instead of uuid1() for converger run id

manishtomar:
because that is more random making it more easier to debug with.


cyli:
LGTM - were we getting collisions?

manishtomar:
No. But they seemed to be pretty similar making them hard to differentiate. For example, 
screen shot 2015-08-05 at 3 06 26 pm
shows the 2 UUIDs (one starting with `4a6` and other starting with `4ce`) that are very similar.

cyli:
Ah ok.  :)

PR #1655, opened 2015-08-05
Inject a logging treq into all the trial integration test calls.

cyli:
Create a logging treq that also logs the response, and use this treq to do all the trial requests.

(except for auth, which is handled by the otter code, not the trial test code).

Part of #1619

### DIFF
--- a/otter/test/test_logging_treq.py
+++ b/otter/test/test_logging_treq.py
@@ -338,3 +338,18 @@ class LoggingTreqTest(SynchronousTestCase):
         On timed out call to delete, failure is returned and request logged
         """
         self._test_method_timeout('delete')
+
+    def test_contents(self):
+        """
+        ``content``, ``json_content``, and ``text_content`` are just directly
+        mapped to treq.
+        """
+        ltreq_instance = logging_treq.LoggingTreq()
+        for name in ('content', 'json_content', 'text_content'):
+            for obj in (ltreq_instance, logging_treq):
+                # patching won't change things that were defined at import time
+                expected = (getattr(self.treq, name), getattr(treq, name))
+                actual = getattr(obj, name)
+                self.assertIn(actual, expected,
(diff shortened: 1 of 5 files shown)


cyli:

otter/test/test_logging_treq.py
```diff
@@ -338,3 +338,18 @@ def test_delete_timeout(self):
...
+        ``content``, ``json_content``, and ``text_content`` are just directly
+        mapped to treq.
+        """
+        ltreq_instance = logging_treq.LoggingTreq()
+        for name in ('content', 'json_content', 'text_content'):
+            for obj in (ltreq_instance, logging_treq):
+                # patching won't change things that were defined at import time
+                expected = (getattr(self.treq, name), getattr(treq, name))
+                actual = getattr(obj, name)
+                self.assertIn(actual, expected,
```
manishtomar: Correct me if I am wrong but shouldn't `actual` always be attribute from `self.treq` since it is patched before the test is called?

manishtomar:
LGTM modulo comment above :)

otter/test/test_logging_treq.py
cyli: I thought so too, but apparently not because for the module level functions at least, they're already set to the `treq` functions at import time.  So after I patch, while `treq` then becomes a mock, those functions have already been defined and don't change.

Instead of this, I can patch, and then reload the module, and then reload again after unpatching.

What do you think?

manishtomar: I guess this needs deeper understanding of import machinery. If what you are saying is true and is expected behavior as per python spec, then how about checking that `expected` is same object as main `treq`'s functions only?

cyli: Sounds good.

manishtomar:
LGTM!

PR #1658, opened 2015-08-06
Resume group implementation

### LINKED ISSUE #1605: Implement resume
The endpoint is already implemented (https://github.com/rackerlabs/otter/blob/master/otter/rest/groups.py#L684), but the functionality is not (https://github.com/rackerlabs/otter/blob/master/otter/controller.py#L97).

As per discussions, "paused" state and "status" are different - a group can both be paused and in error - please see #1604 for pause behavior.

Once the resume endpoint is hit, I'm guessing the "paused" bit should be turned off, and a convergence cycle started?

manishtomar:
Closes #1605. This just changes group state `paused` and triggers convergence on the group.

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -50,10 +50,10 @@ from otter.cloud_client import (
 from otter.convergence.composition import tenant_is_enabled
 from otter.convergence.model import DRAINING_METADATA, group_id_from_metadata
 from otter.convergence.service import (
-    delete_divergent_flag, get_convergence_starter)
+    delete_divergent_flag, get_convergence_starter, mark_divergent)
 from otter.json_schema.group_schemas import MAX_ENTITIES
 from otter.log import audit
-from otter.log.intents import with_log
+from otter.log.intents import msg, with_log
 from otter.models.intents import GetScalingGroupInfo, ModifyGroupStatePaused
 from otter.models.interface import GroupNotEmptyError, ScalingGroupStatus
 from otter.supervisor import (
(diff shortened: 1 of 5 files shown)


manishtomar:

otter/controller.py
```diff
@@ -124,7 +124,19 @@ def pause_scaling_group(log, transaction_id, scaling_group, dispatcher):
...
 
 
-def resume_scaling_group(log, transaction_id, scaling_group):
+def conv_resume_group_eff(trans_id, group):
+    """
+    Resume scaling group of convergence enabled tenant
+    """
+    eff = parallel([
+        Effect(ModifyGroupStatePaused(group, False)),
+        mark_divergent(group.tenant_id, group.uuid).on(
```
cyli: Totally non-blocking nitpick - since it seems like we want to log "mark-dirty-success" every time we mark a group as divergent, I'm wondering if the logging should be done as part of `mark_divergent` so that functions that use it don't have to look up the exact log event name.

cyli:
LGTM!

otter/controller.py
manishtomar: Yes. Sorry I struggled with this since `ConvergenceStarter` does that I didn't want to use it because it returns deferred. At the same time I didn't want to add it in `mark_divergent` since it `ConvergenceStarter` does it :confused:. So, I ended up putting it here. I guess it would be nice to put it in `mark_divergent` once all functions use it directly instead of `ConvergenceStarter`?

cyli: Seems like?  But we can always just file an issue - this is fine as is :)

otter/integration/tests/test_pause.py
```diff
@@ -104,7 +102,7 @@ def test_resume(self):
         webhook or trigger convergence
         """
         group = yield self.test_pause_stops_convergence()
-        yield group.resume()
+        yield group.resume(self.rcs)
```
cyli: Oh nice catch!  Sorry, I didn't even notice this in the previous review

PR #1659, opened 2015-08-06
Update APIError and UpstreamError.

cyli:
This does two things:

1. Add the method and URL in the API error
1. Update UpstreamError to be slightly more permissive and to display the operation in the string representation.

Both of these are needed for a utility that will help debug APIErrors in the trial tests.  Both of these don't change the string representation much, just adds a little bit to the end, so shouldn't break our logs or dashboards.

### DIFF
--- a/otter/test/test_util.py
+++ b/otter/test/test_util.py
@@ -268,7 +268,7 @@ class UpstreamErrorTests(SynchronousTestCase):
         body = json.dumps({"computeFault": {"message": "b"}})
         apie = APIError(404, body, {})
         err = UpstreamError(Failure(apie), 'nova', 'add', 'xkcd.com')
-        self.assertEqual(str(err), 'nova error: 404 - b')
+        self.assertEqual(str(err), 'nova error: 404 - b (add)')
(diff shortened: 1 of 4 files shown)


otter/test/test_util.py
```diff
@@ -289,30 +292,34 @@ def test_apierror_identity(self):
...
         self.assertEqual(err.details, {
             'system': 'identity', 'operation': 'stuff', 'url': 'xkcd.com',
             'message': 'ba', 'code': 410, 'body': body, 'headers': {}})
 
     def test_apierror_unparsed(self):
         """
-        Wraps APIError from identity and uses default string if unable to parses
-        error body
+        Wraps APIError from identity and uses default string if unable to
+        parses error body
```
manishtomar: Sorry nitpick from my earlier code: This should be `if unable to parse error body` :)

manishtomar:
LGTM :)

otter/test/test_util.py
cyli: Oh I didn't even notice that when I wrapped it.  Thanks!

Issue #1629, opened 2015-07-29
Cannot recover from issues with changes-since


cyli:
Noticed an issue with a test tenant where autoscale lost track of a bunch of created servers - not sure what caused this, so this issue is just for tracking.

(https://dev-otterlogs.fedsso.rackspace.com/#/dashboard/temp/VYnnhWDrRfmKdytwbPt-wg)

1. Policy scaled up by 1
1. Convergence cycle gathered from Nova at `2015-07-23T21:05:01.883Z` (finished at `2015-07-23T21:05:04.806Z` with parameters `changes-since=2015-07-23T20%3A30%3A11.857000Z&limit=100`, saw that there were 2 DELETED servers.
1. Autoscale created a server (the created server is not tracked, nor is its ID logged) to satisfy a desired capacity of 1.  Request started at `2015-07-23T21:05:05.837Z` and finished at `2015-07-23T21:05:07.773Z`.
1. Next convergence cycle gathered from Nova at `2015-07-23T21:05:11.875Z` (finished at 2015-07-23T21:05:11.966Z) with parameters `changes-since=2015-07-23T21%3A05%3A01.857000Z&limit=100 `.  Saw only the same 2 DELETED servers as before, and not the previously created server.
1. Autoscale created another server (the created server is not tracked) to satisfy a desired capacity of 1 ...
1. repeat

And so on 31 times until `2015-07-23T21:10:09.440Z`, when the create failed with a 403 because we were over quota:  "Quota exceeded for instances: Requested 1, but already used 100 of 100 instances" and the group went into ERROR state.

Not sure why this happened, but one possibility is that the `changes-since` feed could have been behind by a few seconds, just over the 10 second threshold.  So if it takes 15 seconds for a created server to appear in the `changes-since` listing, autoscale will never catch it if it converges every 10 seconds, since every time it re-gathers it moves the `changes-since` date 10 seconds forward.

As @manishtomar points out, if there is any problem with the `changes-since` list, autoscale can not currently recover because we just build on our server cache by polling `changes-since`.

Possible mitigations/debug logging:

- moving the convergence cycle to once every 30 seconds or minute, as @manishtomar had rightly suggested would be probably more Nova-friendly anyway
- logging (with some filters for private data, such as root server passwords) responses from `logging_treq`
- keeping track of servers we create (@radix also suggest CLB nodes we create) and logging if on a re-gather, we don't see the changes we made take effect
- getting the full list of current servers from Nova every so often (although when should we do this?)

[...]

PR #1660, opened 2015-08-07
not using changes-since when gathering servers

manishtomar:
This implements
https://github.com/rackerlabs/otter/issues/1629#issuecomment-126787343. DELETED servers are found by comparing cache with current list. Any server in cache not found currently is considered deleted and marked accordingly. 

This should not impact existing caches. They will be used and updated with latest servers.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -83,12 +83,13 @@ def get_all_scaling_group_servers(changes_since=None,
     return get_all_server_details(changes_since).on(servers_apply)
 
 
-def updated_deleted_servers(old, new):
+def mark_deleted_servers(old, new):
     """
-    Return servers with servers not found in new list marked as deleted
+    Given dictionaries containing old and new servers, return a list of all
+    servers, with the deleted ones annotated with a status of DELETED.
 
     :param list old: List of old servers
-    :param list second: List of latest servers
+    :param list new: List of latest servers
     :return: List of updated servers
     """
 
@@ -132,7 +133,7 @@ def get_scaling_group_servers(tenant_id, group_id, now,
         servers = (yield all_as_servers()).get(group_id, [])
     else:
         current = yield all_servers()
-        servers = updated_deleted_servers(cached_servers, current)
+        servers = mark_deleted_servers(cached_servers, current)
         servers = list(filter(server_of_group(group_id), servers))
     yield do_return(servers)
 
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -5,7 +5,7 @@ from effect import catch, parallel
 from effect.do import do, do_return
 
 from toolz.curried import filter, groupby, keyfilter, map
-from toolz.dicttoolz import get_in, merge
+from toolz.dicttoolz import assoc, get_in, merge
 from toolz.functoolz import compose, curry, identity
 from toolz.itertoolz import concat
 
@@ -99,8 +99,8 @@ def mark_deleted_servers(old, new):
     old = sdict(old)
     new = sdict(new)
     deleted_ids = set(old.keys()) - set(new.keys())
-    for s in deleted_ids:
-        old[s]["status"] = "DELETED"
+    for sid in deleted_ids:
+        old[sid] = assoc(old[sid], "status", "DELETED")
     return merge(old, new).values()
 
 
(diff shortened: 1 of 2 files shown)


manishtomar:

otter/convergence/gathering.py
```diff
@@ -84,21 +83,27 @@ def group_id(s):
...
-    return merge(sdict(first), sdict(second)).values()
+    old = sdict(old)
+    new = sdict(new)
+    deleted_ids = set(old.keys()) - set(new.keys())
+    for s in deleted_ids:
+        old[s]["status"] = "DELETED"
+    return merge(old, new).values()
 
 
+@curry
```
cyli: Nice!  I this seems neater than partial :)

otter/test/convergence/test_gathering.py
```diff
@@ -337,51 +330,62 @@ def test_no_cache(self):
...
-        changes = self.servers1
-        current = self.servers2
+        asmetakey = "rax:autoscale:group:id"
+        cache = [
+            {'id': 'a', 'metadata': {asmetakey: "gid"}},
+            {'id': 'b', 'metadata': {asmetakey: "gid"}},
+            {'id': 'd', 'metadata': {asmetakey: "gid"}}]
+        current = [
+            {'id': 'a', 'b': 'c', 'metadata': {asmetakey: "gid"}},
+            {'id': 'd', 'metadata': {"changed": "yes"}}]
```
cyli: Great test!  I like that this also tests that 'd' is not marked as DELETED.

cyli:
This LGTM!  Love how small this change is.

otter/convergence/gathering.py
```diff
@@ -84,21 +83,27 @@ def group_id(s):
     return get_all_server_details(changes_since).on(servers_apply)
 
 
-def merge_servers(first, second):
+def updated_deleted_servers(old, new):
     """
-    Return servers got by merging the two sets of servers where second
-    takes precedence over first
+    Return servers with servers not found in new list marked as deleted
```
radix: The name of this method is a bit weird, I guess maybe it was meant to be `update_deleted_servers` (i.e. `update` instead of `updated`?)

Also the docstring is pretty confusing. Maybe "Given dictionaries containing old and new servers, return a list of all servers, with the deleted ones annotated with a `status` of `DELETED`."

otter/convergence/gathering.py
```diff
@@ -84,21 +83,27 @@ def group_id(s):
...
+
     def sdict(servers):
         return {s['id']: s for s in servers}
 
-    return merge(sdict(first), sdict(second)).values()
+    old = sdict(old)
+    new = sdict(new)
+    deleted_ids = set(old.keys()) - set(new.keys())
+    for s in deleted_ids:
+        old[s]["status"] = "DELETED"
```
radix: I think we can get rid of this mutation *and* make the function a lot shorter by using pyrsistent's `transform`:

`return pmap(merge(old, new)).transform([deleted_ids.__contains__, "status"], "DELETED")`

This is a little tricky, but exactly the kind of use case that `transform` was meant for. Plus it's purely functional :)

(I double-checked that this should do the right thing:

```python
>>> servers = pmap({'s1': pmap(), 's2': pmap(), 's3': pmap()})
>>> deleted = ['s1', 's3']
>>> servers.transform([deleted.__contains__, "status"], "DELETED")
pmap({'s3': pmap({'status': 'DELETED'}), 's2': pmap({}), 's1': pmap({'status': 'DELETED'})})
```

Okay!)

radix: To explain this a little more, this is how transform works:

`structure.transform(*path*, *value-or-callable*)`

It basically walks the existing data in `structure` by following `path`. Each element in `path` can either be an index or a function of `index -> Bool`. It continues walking if the Bool is True. `*value-or-callable*` is either a literal value to set as the new value at the path, or a callable to transform the existing value.

In this case, we're using `deleted_ids.__contains__` as an index -> Bool function, to say that we should operate on all values for which the server ID is in the `deleted_ids`. The next and last part of the path is the `status` key, so the value, `DELETED` will be set there.

radix:
Everything else looks great :)

otter/convergence/gathering.py
manishtomar: > The name of this method is a bit weird, I guess maybe it was meant to be update_deleted_servers (i.e. update instead of updated?)

How about `mark_deleted_servers`?

> Maybe "Given dictionaries containing old and new servers, return a list of all servers, with the deleted ones annotated with a status of DELETED."

This is great docstring. Thanks!

manishtomar:

radix:
LGTM :)

PR #1661, opened 2015-08-07
Log the list servers response from cloud_client

cyli:
This might be too big to be logged, but I'm hoping that lack of other fields in this particular logging message will decrease the size of the log event in total.

Addresses part of #1635


radix:
Nice change, thanks for refactoring the tests :-)

Issue #1649, opened 2015-08-04
Metrics script should email about divergent groups


manishtomar:
Since metrics is continuously getting all the group's desired and actual, it can know if a group is diverged for a long time. If so, it can send out email listing those groups. This way we know proactively and can debug or inform customer/support.

Since running the dashboard, I've been trying to manually find divergent groups and fixing them (by converging) apart from debugging why they are divergent. With this email, I can avoid the extra step of finding the divergent groups.

This needs https://github.com/rackerlabs/autoscaling-chef/issues/762 completed.

manishtomar:
@wirehead / @radix / @cyli Thoughts?

cyli:
Just to clarify, is there a threshold for how long a group has been divergent before the email goes out?

manishtomar:
> Just to clarify, is there a threshold for how long a group has been divergent before the email goes out?

@cyli Thats a good question. The same has been asked in #1287 but not yet answered. While thinking about this a little more I was thinking of emailing when the group has not _changed_ for too long, i.e. its `active + pending != desired` and their respective values has remained exactly same for say an hour or so. Does that make sense?

cyli:
@manishtomar Yes, that seems like a good way of going about it.  We had previously discussed converging on a schedule anyway (like once an hour or so just schedule a convergence), and this would give us more data about whether that would help things (since if the solution is just to re-converge, automating the process would be nice).

radix:
Is this something that could be done by Sam's new thing?

manishtomar:
> Is this something that could be done by Sam's new thing?

@sam-falvo thing is something that reacts based on otter's logs. This email is primarily OOB checking if otter is doing its job correctly. So, I think it should not be done there.

wirehead:
I really like the idea of:
 - Having a group that's been unstable for a while get convergence automatically triggered
 - Having a group that fails to react to being triggered send out an alert.
 - Having it documented in such a way that, once we've guaranteed correct operation, we can engage another team to actually read those alerts.

OOB checking and having it send out an email is a pretty good first step.

radix:
I like the idea of auto-converging and stuff.

But I'd rather not do our own alerting/emailing stuff: it'd be a lot nicer if we could be pumping data into an actual monitoring system that we could configure to do notifications.

manishtomar:
> it'd be a lot nicer if we could be pumping data into an actual monitoring system that we could configure to do notifications.

[...]

PR #1662, opened 2015-08-07
Metrics log refactor

manishtomar:
Passing it everywhere from service and script beginning instead of directly using global `metrics_log`. This is precursor to implementing https://github.com/rackerlabs/otter/issues/1649#issuecomment-128787673


radix:
LGTM :)

Issue #1656, opened 2015-08-06
Prevent small time between convergence cycles on same group


manishtomar:
Sometimes when a convergence cycle gets executed due to creation of divergent flag and its completion time corresponds with next interval of `ConvergenceService` it causes the cycle to run almost immediately on that group. The next cycle's gathering does not end up getting the servers created in previous cycle (possibly due to using changes-since and having minor clock drifts between otter and nova servers) and causing extra server to be created. This gets fixed in next cycle by extra server being deleted. However, the tests expects this particular server (that gets deleted) for its purpose and fails when it doesn't find it. 

I've found 2 tests failure in `otter-cloudcafe-syd-convergence-system-tests` till now due to this. I can see other tests failures which _seem_ to have similar symptoms. Even though the end result from convergence is correct we should probably fix this anyways due to 1) reduced load on Nova, 2) tests passing and 3) it may be unpleasant to customers when their (possibly beefy) servers get created and deleted for no reason.

The problem _may_ be reduced when we [stop using changes-since](https://github.com/rackerlabs/otter/issues/1629) but still cannot guarantee it. 

One solution I can think of is to have each divergent group run in fixed 10s intervals starting from the time they got diverged. If there is request to execute convergence again it between those 10s, it will just need to wait for the next cycle. I haven't thought about code changes yet. Any other ideas are welcome.

PR #1663, opened 2015-08-10
Reorganize giant comments in the service.py file into GHC-style "Notes"

radix:
... note that I also bikeshedded an inner function to use `@do` notation to be more clear

While I was thinking about #1656 I looked for a nice simple explanation of convergence cycles and iterations, because I always forget the precise order of things. I couldn't find one, so I wrote one, and then I realize that service.py was filling up with giant comments. So I decided to take the lead from the GHC source code, which uses [a very simple system of organizing long "Notes"](https://ghc.haskell.org/trac/ghc/wiki/Commentary/CodingStyle#Commentsinthesourcecode) which are just top-level comments with tags `[Like this]` which other parts of the source code can reference. The patch should be pretty self-explanatory.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -5,6 +5,100 @@ The top-level entry-points into this module are :obj:`ConvergenceStarter` and
 :obj:`Converger`.
 """
 
+# # Note [Single convergence iteration]
+#
+# An individual convergence iteration goes through these steps:
+# - gather data (from nova, CLB, RCv3, etc)
+# - calculate some steps to reach the desired state (not necessarily all of
+#   them)
+# - run those steps
+#
+# That's only one iteration -- See [Convergence cycles] for the whole cycle.
+
+
+# # Note [Convergence cycles]
+#
+# To sum up the whole process with some pseudo-code:
+#
+# - Regularly list divergent flags OR receive notification via ZK that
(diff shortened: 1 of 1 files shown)


radix:

radix:

otter/convergence/service.py
```diff
@@ -491,24 +535,22 @@ def converge_all_groups(currently_converging, my_buckets, all_buckets,
     yield msg('converge-all-groups', group_infos=group_infos,
               currently_converging=list(cc))
 
+    @do
```
cyli: This looks a lot nicer, thank you for the refactor!

otter/convergence/service.py
```diff
@@ -5,6 +5,85 @@
 :obj:`Converger`.
 """
 
+# # Note [Convergence cycles]
+#
+# A very abstract version of our convergence cycle:
+# - CYCLE (every N seconds)
+# - find all of my divergent groups
```
cyli: Non-blocking:  might want to note that this is done by looking at divergent flags, which is implied but not overtly stated.

cyli:
LGTM!  Thanks for writing this explanation and re-organizing all the notes.

radix:
Thanks ying :)

PR #1665, opened 2015-08-10
Use UpstreamError in the trial tests

### LINKED ISSUE #1619: Better trial test reporting

cyli:
So we can more easily see what the trial tests were doing when they failed, so if we get an API error, we can see what caused it without having to guess by looking at the body.

the error messages now look like:

```
[ERROR]
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1105, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/opt/otter/otter/integration/tests/test_convergence.py", line 1006, in test_recover_from_identity_auth_failures
    yield self.helper.start_group_and_wait(group, rcs, desired=5)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1105, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/opt/otter/otter/integration/lib/trial_tools.py", line 258, in start_group_and_wait
    timeout=600)
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 578, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/opt/otter/otter/integration/lib/utils.py", line 119, in wrap_failure
    raise UpstreamError(failure, system, message)
otter.util.http.UpstreamError: AS error: AS error: Connection was refused by other side: 111: Connection refused. (Getting scaling group state) (Wait for scaling group state to reach a particular point)
```

Giving us a sort of very ugly stack trace (failed during the get scaling group state part of waiting for state).

Fixes #1619.

### DIFF
--- a/otter/integration/lib/utils.py
+++ b/otter/integration/lib/utils.py
@@ -106,17 +106,21 @@ def diagnose(system, message):
     Decorator that adds a callback to the deferred return that will wrap a
     failure so that we'll know what we were doing when that caused the failure.
     """
-    def wrap_failure(failure):
+    def _unwrap_first(failure):
         if failure.check(FirstError):
-            return wrap_failure(failure.value.subFailure)
+            return _unwrap_first(failure.value.subFailure)
+        return failure
 
-        if failure.check(APIError):
-            raise UpstreamError(failure, system, message, failure.value.url)
+    def wrap_failure(failure):
+        new_f = _unwrap_first(failure)
 
-        if failure.check(ConnectionRefusedError, UpstreamError):
+        if new_f.check(APIError):
+            raise UpstreamError(new_f, system, message, new_f.value.url)
+
+        if new_f.check(ConnectionRefusedError, UpstreamError):
             # allowing UpstreamError too, since the traceback will be too short
             # and this gives us a sort of diagnosis stack
-            raise UpstreamError(failure, system, message)
+            raise UpstreamError(new_f, system, message)
 
         return failure
 
--- a/otter/integration/lib/utils.py
+++ b/otter/integration/lib/utils.py
@@ -1,8 +1,18 @@
+"""
+Shared utilities for trial test library.
+"""
+
+from functools import wraps
+
 from characteristic import Attribute, attributes
 
 from pyrsistent import PSet
 
+from twisted.internet.defer import FirstError
+from twisted.internet.error import ConnectionRefusedError
+
 from otter.convergence.model import ServerState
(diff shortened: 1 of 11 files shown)


otter/integration/lib/utils.py
```diff
@@ -89,3 +99,30 @@ def _count_dead_servers(servers):
     Count servers that are in error state.
     """
     return len([s for s in servers if s.state is ServerState.ERROR])
+
+
+def diagnose(system, message):
```
radix: what the hell, this is so awesome. I need something like this for Effect.

manishtomar:
This is amazing @cyli

manishtomar:
Non-blocking: `diagnose` doesn't have tests which I am fine with. 

This looks great otherwise: :+1: Nice decorator :)

cyli:
@manishtomar good point, I'll add them!

manishtomar:
@cyli I believe you were still working on this. Should I change it to "in progress"?

cyli:

cyli:
@manishtomar Apologies!  I'll add the in-progress for now.  I think it should be ready if they pass, but if not, then this can stop appearing as up for review :)

cyli:
Ok, probably ready for review :)

radix:
Ok, I've reviewed your changes since my last review. LGTM :)

PR #1666, opened 2015-08-12
Log unchanged divergent groups

manishtomar:
Implements #1649. However, metrics logs are not yet shipped to ES: https://github.com/rackerlabs/autoscaling-chef/issues/762

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -18,7 +18,7 @@ from silverberg.client import ConsistencyLevel
 from silverberg.cluster import RoundRobinCassandraCluster
 
 from toolz.curried import filter, get_in, groupby
-from toolz.dicttoolz import keyfilter, merge
+from toolz.dicttoolz import merge
 from toolz.functoolz import identity
 
 from twisted.application.internet import TimerService
@@ -342,6 +342,10 @@ class Options(usage.Options):
         self.update(json.load(self.open(self['config'])))
 
 
+def metrics_set(metrics):
+    return set((g.tenant_id, g.group_id) for g in metrics)
+
+
 def unchanged_divergent_groups(clock, current, timeout, group_metrics):
     """
     Return list of GroupMetrics that have been divergent and unchanged for
@@ -356,9 +360,11 @@ def unchanged_divergent_groups(clock, current, timeout, group_metrics):
     """
     converged, diverged = partition_bool(
         lambda gm: gm.actual + gm.pending == gm.desired, group_metrics)
-    # stop tracking all converged groups
-    converged = set((g.tenant_id, g.group_id) for g in converged)
-    updated = keyfilter(lambda k: k not in converged, current)
+    # stop tracking all converged and deleted groups
+    deleted = set(current.keys()) - metrics_set(group_metrics)
+    updated = current.copy()
+    for g in metrics_set(converged) | deleted:
+        updated.pop(g, None)
     # Start tracking divergent groups depending on whether they've changed
     now = clock.seconds()
     to_log, new = [], {}
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -19,7 +19,7 @@ from silverberg.cluster import RoundRobinCassandraCluster
 
 from toolz.curried import filter, get_in, groupby
 from toolz.dicttoolz import merge
-from toolz.functoolz import identity
+from toolz.functoolz import curry, identity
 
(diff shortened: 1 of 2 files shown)


otter/metrics.py
```diff
@@ -253,6 +254,40 @@ def connect_cass_servers(reactor, config):
...
+    for gm in converged:
+        dg.pop((gm.tenant_id, gm.group_id), None)
+    # Start tracking divergent groups depending on whether they've changed
+    now = clock.seconds()
+    for gm in diverged:
+        pair = (gm.tenant_id, gm.group_id)
+        if pair in dg:
+            last_time, values = dg[pair]
+            if values != hash((gm.desired, gm.actual, gm.pending)):
+                del dg[pair]
```
cyli: Non-blocking - if this is deleted, won't it not track this group as divergent until the next metrics cycle?  But this group is currently divergent, it has just moved.  This probably doesn't matter much, since we seem to be gathering metrics once a minute, and the timeout is once an hour.  If the metrics cycle is increased or timeout decreased, it may start to make more of a difference.

otter/metrics.py
```diff
@@ -253,6 +254,40 @@ def connect_cass_servers(reactor, config):
         seed_endpoints, config['keyspace'], disconnect_on_cancel=True)
 
 
+def log_divergent_groups(clock, divergent_groups, log, timeout, group_metrics):
+    """
+    Log groups that have not changed and been divergent for long time
```
cyli: Do we want to document the parameters here, particularly that this function is expecting to mutate the `divergent_groups` parameter?

otter/metrics.py
manishtomar: Yeah. I wanted to track only groups that are stuck.

otter/metrics.py
manishtomar: Yes. Excellent idea. I thought about it and forgot. Will do. Thanks!

otter/metrics.py
cyli: Right, but the time it has been stuck should start when we first see that particular combination of `(gm.desired, gm.actual, gm.pending)`, it seems.

cyli: Sorry, to be clear, I'm not saying this change should be made, I just wanted to check with you that this isn't an issue (I don't think it is).

But let's say metrics runs every **m** seconds, and the timeout is **n** seconds.

At time `T=0`, group 1 is divergent with hash `a`, so it's put in the divergent list.
At time `T=m`, group 1 is divergent but with a different hash `b`, so it's removed from the divergent list.
At time `T=2m`, group 1 is still divergent with hash `b`, so it is put back in the divergent queue.
At time `T=2m + n`, group 1 is logged as divergent for `n` seconds, but actually it's been divergent for `n + m` seconds, since it's has hash `b` since time `T=m`, not `T=2m`.

otter/metrics.py
manishtomar: On the other hand, I think it will be better to return updated copy of `divergent_groups` instead of mutating it. That way it remains pure. I'll do that.

otter/metrics.py
manishtomar: Thanks for the detailed explanation and sorry you had to do that. I should've expressed it better earlier. The log says `Group has been diverged _and_ unchanged for x time` which is off by `m` seconds. Considering that `m` is small, I think its fine. Moreover, this information is for diagnostic purpose only. I would imagine us to look for groups logs in kibana and find out why it is divergent in which case exact time is not really an issue. To be clear, I completely agree with you about the incorrect time :) but I think its fine.

cyli: Cool, thanks for clarifying!  :D  And sorry for wasting all this space for a non-issue!

otter/metrics.py
cyli: Sounds good :)

cyli:
Apologies, I may just be missing something - what happens if a group is deleted between one metrics gather and another?  The group would not appear in the new metrics, but if it was previously divergent wouldn't it remain in `self._divergent`?

manishtomar:
> The group would not appear in the new metrics, but if it was previously divergent wouldn't it remain in self._divergent?

Yes. It will remain in `self._divergent_groups` but it will not be logged. Great catch @cyli . I'll change it to remove it.

cyli:
LGTM!  Thanks for this!  It'd be a great way of automating fixes :)

PR #1667, opened 2015-08-12
Include a log message for server quota issues

cyli:
Otherwise quota issues appear as "Unknown error reasons"


manishtomar:
LGTM!

PR #1668, opened 2015-08-13
Otter stops after scheduler finishes

### LINKED ISSUE #1539: otter should stop after scheduler service stops
This is how it was earlier but has changed since partitioner has been split out for converger. Currently, return value of [`got_buckets`](https://github.com/rackerlabs/otter/blob/master/otter/util/zkpartitioner.py#L117) is ignored due to which scheduler can lose events that it was triggering at that time. The right solution for this #1400 but until then I think its best that otter continues to run until scheduler interval completes.

manishtomar:
Closes #1539.

In zkpartitioner.py `got_buckets` return value is propogated to `check_partition` return that ensures that partitioner service stops after `got_buckets` deferred is fired. This way scheduler will stop after its events are processed like before. 

Changed `ConvergerService` to return tuple with deferred instead of the deferred to allow it to stop right away. We don't want to block otter stopping when convergence cycle is going on since it idempotent and it will be started in next otter start.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -607,6 +607,10 @@ class Converger(MultiService):
         """
         ceff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
             partial(self._converge_all, my_buckets))
+        # Return deferred as 1-element tuple for testing only.
+        # Returning deferred would block otter from shutting down until
+        # it is fired which we don't need to do since convergence is itempotent
+        # and will be triggered in next start of otter
         return (perform(self._dispatcher, self._with_conv_runid(ceff)), )
 
     def divergent_changed(self, children):
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -607,7 +607,7 @@ class Converger(MultiService):
         """
         ceff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
             partial(self._converge_all, my_buckets))
-        return perform(self._dispatcher, self._with_conv_runid(ceff))
+        return (perform(self._dispatcher, self._with_conv_runid(ceff)), )
 
     def divergent_changed(self, children):
         """
--- a/otter/util/zkpartitioner.py
+++ b/otter/util/zkpartitioner.py
@@ -67,9 +67,10 @@ class Partitioner(MultiService):
 
     def stopService(self):
         """Release the buckets."""
-        MultiService.stopService(self)
+        d = MultiService.stopService(self)
         if self.partitioner.acquired:
-            return self.partitioner.finish()
+            d.addCallback(lambda _: self.partitioner.finish())
+        return d
 
     def reset_path(self, path):
         """Re-initialize the partitioner to use a new path."""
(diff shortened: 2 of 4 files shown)


otter/util/zkpartitioner.py
```diff
@@ -67,9 +67,10 @@ def startService(self):
 
     def stopService(self):
         """Release the buckets."""
-        MultiService.stopService(self)
+        d = MultiService.stopService(self)
```
cyli: Oh, nice catch!  Also, non-blocking, but I think the general suggestion is to avoid calling the superclass like this (since Python3 only has new style classes and this kind of superclass call may no longer work) and to use `super(Partitioner, self).stopService()`.  That only works if `Partitioner` inherits from both `MultiService` and `object`, though (which is Python3-compatible).

otter/convergence/service.py
```diff
@@ -607,7 +607,7 @@ def buckets_acquired(self, my_buckets):
         """
         ceff = Effect(GetChildren(CONVERGENCE_DIRTY_DIR)).on(
             partial(self._converge_all, my_buckets))
-        return perform(self._dispatcher, self._with_conv_runid(ceff))
+        return (perform(self._dispatcher, self._with_conv_runid(ceff)), )
```
cyli: Out of curiosity, does this even need to return at all?  Is the result used for anything except testing?

manishtomar: Its only used for testing.

cyli: Ok cool, thanks.  :)  Would it make sense to include that as a comment, or in the docstring (e.g. the return value only matters in that if a deferred is returned, the partitioner service will not stop until the deferred is callbacked?

cyli:

cyli:
Aside from non-blocking nitpicks above, LGTM pending +1 from jenkins.  :)

otter/convergence/service.py
manishtomar: Yes. Very much! I realized it after you asked the question and felt bad for not already putting it. Thanks :)

otter/util/zkpartitioner.py
manishtomar: Yeah. Unfortunately, `MultiService` is old-style class in 15.0.0.

cyli: Right, but Partitioner can be declared as `class Partitioner(MultiService, object)`, in which case the new-style class superclass call can be used.

```
>>> class A:
...    def foo(self):
...       print "I am old"
... 
>>> class B(A):
...    def foo(self):
...       super(B, self).foo()
...       print "Should not get here"
... 
>>> B().foo()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in foo
TypeError: must be type, not classobj
>>> class C(A, object):
...    def foo(self):
...       super(C, self).foo()
...       print "I am new"
... 
>>> C().foo()
I am old
I am new
```

cyli: Again, non-blocking.  :)

manishtomar: Ooh yeah. Forgot I could do that :) Will do.

cyli:
LGTM, thanks for fixing this!!

Issue #1653, opened 2015-08-05
Spec.py has a callable for each event type that knows how to split that type of log


cyli:
So for instance the execute-convergence log can split if the servers list is too long, etc.

radix:
I started on a branch a whiiiile back that would make it so you could put callables instead of formattables as the values of the log spec dictionary, and they would be called with the event dict to produce 1 or more log message. I'm not sure if I actually produced anything valuable, I'll have to check.

manishtomar:
This can be used to split up CF messages too and help in implementing #1583. @radix Did you manage to find your branch that contains this?

manishtomar:
`spec.py` is observing a single log event. Is it possible for it to fan it to multiple log events? As in, no matter what it does, the event is going to the next observer in the list right?

cyli:
man

radix:
I found the branch, it's https://github.com/rackerlabs/otter/tree/callable-specs. super simple, it doesn't allow returning a list but that shouldn't be too hard to add.

cyli:
@radix thanks!

cyli:
Done

PR #1671, opened 2015-08-17
Event type splitting

cyli:
Based on @radix's branch https://github.com/rackerlabs/otter/tree/callable-specs for spec taking a callable.

This modifies it so that the callable can return multiple events.

Part of #1653


otter/log/spec.py
```diff
@@ -78,44 +80,58 @@ class MsgTypeNotFound(Exception):
     """
 
 
-def try_msg_types(*tries):
+def try_msg_types(event, specs, tries):
```
manishtomar: +1 for passing `specs`. Should've done that earlier.

manishtomar:
LGTM! :)

Issue #1677, opened 2015-08-19
Run all our integration tests, etc. against an otter running Python 2.7.10


cyli:
Since the new OS X update upgrades the native Python to be 2.7.10, and also there are a lot of bug and security fixes for 2.7.10.

https://hg.python.org/cpython/raw-file/15c95b7d81dc/Misc/NEWS

(one issue that we've run into we check `partial(...).keywords` in the unit tests, and we assert that it's None - in py2.7.10, it's always a dictionary now).

PR #1678, opened 2015-08-19
Make unit tests pass on Python 2.7.10

cyli:
https://hg.python.org/cpython/raw-file/15c95b7d81dc/Misc/NEWS  

"The keywords attribute of functools.partial is now always a dictionary."

But on earlier versions, `functools.partial.keywords` might be `None`.

So change the tests so we're not testing for that specifically.

Part of #1677 

If you're running unit tests on Mac OS 10.10.5, you'll probably need to run unit tests with this merged in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,7 @@ script:
 
 after_script:
     - coveralls
+
+branches:
+  only:
+    - master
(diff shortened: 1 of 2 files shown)


radix:
thank you so much for cleaning up those tests and not just fixing the ugly `partial` use :)

PR #1676, opened 2015-08-19
RCv3 cafe convergence tests

cyli:
Some changes to make RCv3 cafe tests run better and in convergence.

~~Still waiting on the following:~~
- [x] #1675 
- [x] https://github.com/rackerlabs/autoscaling-chef/pull/763

#1681 Should also be done, but possibly as a different PR.

PR #1680, opened 2015-08-20
Tell convergence to deal with RCv3 load balancers

### LINKED ISSUE #1675: Include RCv3 load balancer pools in the desired LBs when parsing LBs from JSON
In `otter.convergence.composition.json_to_LBConfigs`

cyli:
Fixes #1675.

Tested this with #1676.  All those tests against production pass (at least they did once, which was the number of times I've run this branch against production so far).

Update: Testing with #1676 works against mimic too.

### DIFF
--- a/otter/test/convergence/test_composition.py
+++ b/otter/test/convergence/test_composition.py
@@ -31,12 +31,17 @@ class JsonToLBConfigTests(SynchronousTestCase):
                  {'loadBalancerId': 20, 'port': 800},
                  {'loadBalancerId': 20, 'type': 'RackConnectV3'},
                  {'loadBalancerId': 200, 'type': 'RackConnectV3'},
-                 {'loadBalancerId': 21, 'port': 81}]),
-            pset([CLBDescription(lb_id='20', port=80),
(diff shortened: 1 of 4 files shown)


otter/test/convergence/test_composition.py
```diff
@@ -10,36 +10,33 @@
...
-    def test_with_rackconnect(self):
-        """
-        LB config with rackconnect
+        LB config with both CLBs and rackconnect.
         """
         self.assertEqual(
             json_to_LBConfigs(
                 [{'loadBalancerId': 20, 'port': 80},
+                 {'loadBalancerId': 20, 'port': 800},
+                 {'loadBalancerId': 20, 'type': 'RackConnectV3'},
```
manishtomar: Will be nice to check with string `loadBalancerId` since RCv3 Ids are UUIDs

manishtomar:
LGTM modulo comment :)

otter/test/convergence/test_composition.py
cyli: Oh good point, thanks!  :D

PR #1682, opened 2015-08-20
Cloud client logging

### LINKED ISSUE #1635: Logging responses

cyli:
Finish adding response logging to cloud_client functions that need them.

Fixes #1635.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -573,7 +573,8 @@ def add_clb_nodes(lb_id, nodes):
         _match_errors(mappings, code, json_body)
         _process_clb_api_error(code, json_body, lb_id)
 
-    return eff.on(error=_parse_known_errors)
+    return eff.on(error=_parse_known_errors).on(
+        log_success_response('request-add-clb-nodes', identity))
 
 
 def change_clb_node(lb_id, node_id, condition, weight):
@@ -612,6 +613,7 @@ def change_clb_node(lb_id, node_id, condition, weight):
             json_body)
 
     return eff.on(error=_parse_known_errors)
+    # CLB 202 response here has no body, so no response logging needed
 
 
 def remove_clb_nodes(lb_id, node_ids):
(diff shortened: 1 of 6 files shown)


cyli:

radix:
Thanks for the nice test refactoring, this LGTM :)

Issue #1669, opened 2015-08-14
Sometimes convergence does not get the updated state


manishtomar:
Scaling group f1547f43-9280-4204-b9f9-df8fdeab579c executed policy on 2015-08-13T13:08:31.550Z to create 12 servers but convergence got desired state of 0 servers. This probably happened because convergence gets triggered _before_ state gets updated. `maybe_execute_policy` calls `converge` which triggers convergence. But state gets updated _after_ `maybe_execute_policy` gets fired since it is called with `modify_state`. After discussing the @cyli , one decent solution is to have a wrapper function that modifies the state and triggers convergence _after_ updating the state. This function can then be called instead of `modify_state` from rest and scheduler when executing policies or updating config.

PR #1683, opened 2015-08-20
Refactor start convergence

manishtomar:
In the process of removing `ConvergenceStarter` and replacing it with effect based `trigger_convergence` function. This is possible since rest APIs objects have dispatcher that they can use to perform the effects. 

This is precursor to implementing #1669 which will remove `ConvergenceStarter` completely.

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -205,7 +205,7 @@ def obey_config_change(log, transaction_id, config, scaling_group, state,
     return d
 
 
-def delete_group(log, trans_id, group, force):
+def delete_group(dispatcher, log, trans_id, group, force):
     """
     Delete group based on the kind of tenant
 
@@ -221,7 +221,7 @@ def delete_group(log, trans_id, group, force):
 
     def check_and_delete(_group, state):
         if state.desired == 0:
-            d = trigger_convergence_deletion(log, group)
+            d = trigger_convergence_deletion(dispatcher, group)
             return d.addCallback(lambda _: state)
         else:
             raise GroupNotEmptyError(group.tenant_id, group.uuid)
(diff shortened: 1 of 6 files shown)


manishtomar:

otter/convergence/service.py
```diff
@@ -386,6 +386,17 @@ def delete_divergent_flag(tenant_id, group_id, version):
         yield msg('mark-clean-success')
 
 
+def trigger_convergence(tenant_id, group_id):
+    """
+    Trigger convergence on a scaling group
+    """
+    eff = with_log(mark_divergent(tenant_id, group_id),
+                   tenant_id=tenant_id, scaling_group_id=group_id)
+    return eff.on(success=lambda _: msg("mark-dirty-success"),
```
cyli: Apologies, I probably just don't understand performing `BoundFields` - since "mark-dirty-success" and "mark-dirty-failure" are no longer within the `with_log` context (and hence within `BoundFields`, does this mean that "mark-dirty-success" and "mark-dirty-failure" will not be logged with the `tenant_id` and `scaling_group_id` fields?  Whereas they seem to still be, in `ConvergenceStarter.start_convergence`.

otter/controller.py
```diff
@@ -629,35 +629,8 @@ def convergence_remove_server_from_group(
         yield do_return(state)
 
 
-def perform_convergence_remove_from_group(
```
cyli: Yay so much deletion!

cyli:

cyli:
Aside from the logging nitpick, this LGTM!

otter/convergence/service.py
manishtomar: You understood it correct @cyli. I didn't realize that the logs were not bound. Thanks for catching it! Now that I think of it, the bounding may not be required since this function is going to be called by rest or scheduler both of which should bind it tenant, group and trans ids before calling it. Let me do that.

manishtomar:

cyli:
LGTM!  Thanks - propagating a ZK error up as a 500 makes more sense than accepting a change, and then doing nothing!  Nice catch!

PR #1686, opened 2015-08-25
Change the inheritance order of SupervisorService and FakeSupervisor

cyli:
Because:

```
>>> class A(object):
...    pass
... 
>>> class B(object, A):
...    pass
... 
  File "<stdin>", line 1, in <module>
TypeError: Error when calling the metaclass bases
    Cannot create a consistent method resolution
order (MRO) for bases object, A
>>> class B(A, object):
...    pass
... 
>>> 
```

And it seems like `Service`, in Twisted, is now (15.3.0) a new-style class.  We get the same error for `SupervisorService` when running otter with the newest version of Twisted.

### DIFF
--- a/otter/supervisor.py
+++ b/otter/supervisor.py
@@ -77,7 +77,7 @@ class RequestBag(object):
 
 
 @implementer(ISupervisor)
-class SupervisorService(object, Service):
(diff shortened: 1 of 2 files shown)


radix:
Python.

PR #1688, opened 2015-08-25
Don't converge too fast

### LINKED ISSUE #1656: Prevent small time between convergence cycles on same group

radix:
Fixes #1656 

This adds a `recently_converged` `Reference(pmap())`, which keeps track of groups that have recently been converged. If a group has recently been converged (within 10 seconds), it is not converged. The group will be left dirty, so the next time we check it, we will converge it.

Unrelated changes that I happened to make in this branch:

- did a big `s/lambda i: None/noop/` in service.py
- refactored some tests to more tersely describe some sequences

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -483,6 +483,12 @@ def get_my_divergent_groups(my_buckets, all_buckets, divergent_flags):
     return converging
 
 
+def eff_finally(eff, after_eff):
+    """Run some effect after another effect, whether it succeeds or fails."""
+    return eff.on(success=lambda r: after_eff.on(lambda _: r),
+                  error=lambda e: after_eff.on(lambda _: six.reraise(*e)))
+
+
 @do
 def converge_one_group(currently_converging, recently_converged,
                        tenant_id, group_id, version,
@@ -501,16 +507,15 @@ def converge_one_group(currently_converging, recently_converged,
     :param callable execute_convergence: like :func`execute_convergence`, to
         be used for test injection only
     """
-    @do
-    def cvg():
-        result = yield execute_convergence(tenant_id, group_id, build_timeout)
-        time_done = yield Effect(Func(time.time))
-        yield recently_converged.modify(
-            lambda rcg: rcg.set(group_id, time_done))
-        yield do_return(result)
+    mark_recently_converged = Effect(Func(time.time)).on(
+        lambda time_done: recently_converged.modify(
+            lambda rcg: rcg.set(group_id, time_done)))
+    cvg = eff_finally(
+        execute_convergence(tenant_id, group_id, build_timeout),
+        mark_recently_converged)
 
     try:
-        result = yield non_concurrently(currently_converging, group_id, cvg())
+        result = yield non_concurrently(currently_converging, group_id, cvg)
     except ConcurrentError:
         # We don't need to spam the logs about this, it's to be expected
         return
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -590,8 +590,6 @@ def converge_all_groups(currently_converging, recently_converged,
         tenant_id, group_id = info['tenant_id'], info['group_id']
         if group_id in recent_groups:
             # Don't converge a group if it has recently been converged.
-            # TODO: clean up the pmap so it doesn't accumulate an entry for all
-            # groups over time!
             continue
         eff = converge(tenant_id, group_id, info['dirty-flag'])
         effs.append(
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -540,10 +540,10 @@ def converge_one_group(currently_converging, recently_converged,
 
 
 @do
-def converge_all_groups(currently_converging, recently_converged,
-                        my_buckets, all_buckets,
-                        divergent_flags, build_timeout,
-                        converge_one_group=converge_one_group):
+def converge_all_groups(
+        currently_converging, recently_converged, my_buckets, all_buckets,
+        divergent_flags, build_timeout, interval,
+        converge_one_group=converge_one_group):
     """
     Check for groups that need convergence and which match up to the
     buckets we've been allocated.
@@ -558,6 +558,9 @@ def converge_all_groups(currently_converging, recently_converged,
     :param divergent_flags: divergent flags that were found in zookeeper.
     :param number build_timeout: number of seconds to wait for servers to be in
         building before it's is timed out and deleted
+    :param number interval: number of seconds between attempts at convergence.
+        Groups will not be converged if less than this amount of time has
+        passed since the end of its last convergence.
     :param callable converge_one_group: function to use to converge a single
         group - to be used for test injection only
     """
@@ -589,7 +592,8 @@ def converge_all_groups(currently_converging, recently_converged,
             result = yield Effect(TenantScope(eff, tenant_id))
             yield do_return(result)
 
-    recent_groups = yield get_recently_converged_groups(recently_converged)
+    recent_groups = yield get_recently_converged_groups(recently_converged,
+                                                        interval)
     effs = []
     for info in group_infos:
         tenant_id, group_id = info['tenant_id'], info['group_id']
@@ -604,7 +608,7 @@ def converge_all_groups(currently_converging, recently_converged,
 
 
 @do
-def get_recently_converged_groups(recently_converged):
+def get_recently_converged_groups(recently_converged, interval):
     """
     Return a list of recently converged groups, and garbage-collect any groups
     in the recently_converged map that are no longer 'recent'.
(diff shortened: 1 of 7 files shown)


radix:

radix:

radix:

cyli:

otter/convergence/service.py
```diff
@@ -581,6 +600,23 @@ def converge(tenant_id, group_id, dirty_flag):
...
+def get_recently_converged_groups(recently_converged):
+    """
+    Return a list of recently converged groups, and garbage-collect any groups
+    in the recently_converged map that are no longer 'recent'.
+    """
+    # STM would be cool but this is synchronous so whatever
+    recent = orig = yield recently_converged.read()
+    now = yield Effect(Func(time.time))
+    for group in recent.keys():
+        if now - recent[group] > 10:
```
cyli: Should this be the same value as the converger interval, which is configurable?

otter/convergence/service.py
```diff
@@ -566,14 +578,21 @@ def converge(tenant_id, group_id, dirty_flag):
...
             result = yield Effect(TenantScope(eff, tenant_id))
             yield do_return(result)
 
+    recent_groups = yield get_recently_converged_groups(recently_converged)
     effs = []
     for info in group_infos:
         tenant_id, group_id = info['tenant_id'], info['group_id']
+        if group_id in recent_groups:
+            # Don't converge a group if it has recently been converged.
+            # TODO: clean up the pmap so it doesn't accumulate an entry for all
```
cyli: Is this TODO covered by the cleanup in `get_recently_converged_groups`?

radix: yes, it is! thanks for spotting that. I'll remove it.

otter/convergence/service.py
radix: Yeah, I guess it should be. Time to thread some more config through the app :)

otter/test/convergence/test_service.py
```diff
@@ -330,13 +359,13 @@ def test_unexpected_errors(self):
         dirty flag.
         """
         expected_error = RuntimeError('oh no!')
-        sequence = SequenceDispatcher([
+        sequence = [
```
cyli: Non-blocking - just wondering if it's worth converting this test (and maybe the failure test) to not use the fallback dispatcher, to ensure that if there is an error during convergence that `recently_converged` is not updated, and also that if the convergence finishes but fails, `recently_converged` *is* updated?

otter/test/convergence/test_service.py
```diff
@@ -504,32 +539,61 @@ def test_filter_out_currently_converging(self):
...
+        """
+        eff = self._converge_all_groups(['00_g1'])
+        sequence = [
+            (ReadReference(ref=self.currently_converging), lambda i: pset([])),
+            (Log('converge-all-groups',
+                 dict(group_infos=[self.group_infos[0]],
+                      currently_converging=[])),
+             noop),
+            (ReadReference(ref=self.recently_converged),
+             lambda i: pmap({'g1': 5})),
```
cyli: Do we also want to include other, expired groups in the `recently_converged` that are not this group, to show that other groups also get removed from `recently_converged` every time it's modified?  To test the cleanup function.

otter/test/convergence/test_service.py
radix: Oh, that's a really good point. I'd really rather avoid using the reference_dispatcher at all in our tests, and make sure everything is explicitly tracked, but I didn't want to go and update all of them now. I will make the change for these failure tests as you suggest :)

radix: For the record, I actually decided to go a different way with this -- I mean, I still updated the test to be more explicit, but I decided to add the group to "recently converged" even in the case of error, so that even when a group is messing up it won't be converged too quickly again.

otter/test/convergence/test_service.py
radix: I'm not sure *exactly* what you mean but I think I made this test better by adding a couple more groups to currently_converging to make sure more cases are handled. thanks :)

radix:

radix:
@cyli I think I've addressed all your comments :)

radix:
Oh no, I didn't, I forgot about the configurable 10 seconds thing :(

radix:
ok, this time for sure :)

otter/test/convergence/test_service.py
cyli: Your next test handles it - basically I just wanted to make sure that a group that was still in the `recently_converged` group that isn't being converged this time gets removed, because otherwise deleted groups will remain forever.

cyli:

cyli:
This LGTM pending +1 from jenkins!  :)  I think @manishtomar wanted to look at it too, though.

manishtomar:
@radix I am looking at it. However, feel free to merge if you are eager since @cyli has already +1 it :)

otter/convergence/service.py
```diff
@@ -581,6 +607,23 @@ def converge(tenant_id, group_id, dirty_flag):
...
+    """
+    Return a list of recently converged groups, and garbage-collect any groups
+    in the recently_converged map that are no longer 'recent'.
+    """
+    # STM would be cool but this is synchronous so whatever
+    recent = orig = yield recently_converged.read()
+    now = yield Effect(Func(time.time))
+    for group in recent.keys():
+        if now - recent[group] > interval:
+            recent = recent.remove(group)
```
manishtomar: Is it safe to remove entry while enumerating it?

cyli: `recent.keys()` produces an iterable, which is being iterated over, so changing `recent` in the loop should be ok (since it's not affecting the iterable of keys).

manishtomar: You mean it returns a _copy_ of the keys? I am guessing if its just iterating directly in the pmap it can cause problems? I remember I had got into this problem when implementing heartbeat check in [participate](https://github.com/manishtomar/participate/blob/master/participate.py#L134-L144)

cyli: Python `dict.keys()` returns a copy of the keys, and it'd be strange for PMap to break that contract.  I think it does return a copy:  it makes a `PVector` of all the keys (https://github.com/tobgu/pyrsistent/blob/master/pyrsistent/_pmap.py#L117).

But also, `recent` is not being mutated, the variable is just being bound to [a new PMap](https://github.com/tobgu/pyrsistent/blob/master/pyrsistent/_pmap.py#L161), which I think is fine?

```
>>> a = [2, 3, 1]
>>> for i in a:
...    print i
...    a = []
... 
2
3
1
```

radix: @manishtomar the important thing to realize is that there is no mutation going on. Those are pmaps, and .remove() returns a new pmap. notice the `recent = ` part, on the line with the call to `.remove(group)`. The only "mutation" is the rebinding of the `recent` name.

radix: But, yeah, the loop is weird and I am not happy about rebinding `recent`. I'll try to come up with a more pure formulation.

radix: ok, I have a MUCH better formulation now :) thanks for the comments!

otter/test/convergence/test_service.py
```diff
@@ -504,33 +566,66 @@ def test_filter_out_currently_converging(self):
...
+        self.assertEqual(perform_sequence(sequence, eff), [])
+
+    def test_dont_filter_out_non_recently_converged(self):
+        """
+        If a group was converged in the past but not recently, it will be
+        cleaned from the ``recently_converged`` map, and it will be converged.
+        """
+        # g1: converged a while ago; divergent -> removed and converged
+        # g2: converged recently; not divergent -> not converged
+        # g3: converged a while ago; not divergent -> removed and not converged
```
manishtomar: +1. Nice comments :) Really help.

manishtomar:
LGTM!

radix:
Thanks :D

PR #1689, opened 2015-08-25
Bump to effect 0.10 and use its new testing utilities

radix:
I've moved the perform_sequence function to effect, and also moved (and modified) `nested_parallel` to effect. I renamed it to `parallel_sequence` and also changed its behavior, because I realized the version we had didn't allow distinguishing chains of parallelized effects from top-level parallel effects.

e.g. `nested_parallel([(a, noop), (b, noop)])` would match both of the following: `parallel([a, b])` and `parallel([a.on(lambda r: b)])`. So, parallel_sequence requires a list of lists, where each list represents the sequence of effects that each parallelized effect goes through.

unfortunately lots of whitespace changes, but the changes themselves are trivial.

### DIFF
--- a/otter/test/convergence/test_gathering.py
+++ b/otter/test/convergence/test_gathering.py
@@ -14,7 +14,7 @@ from effect import (
 
 from effect.async import perform_parallel_async
 from effect.testing import (
-    EQDispatcher, EQFDispatcher, Stub)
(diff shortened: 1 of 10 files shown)


cyli:
Nice catch on `nested_parallel`!  And yay modular libraries.  This is beautiful, thanks!  LGTM.

PR #1690, opened 2015-08-26
Modify state and trigger convergence

### LINKED ISSUE #1669: Sometimes convergence does not get the updated state

manishtomar:
Closes #1669.

This also completely removes `ConvergenceStarter` since the dispatcher created in `tap/api.py` is passed to REST and scheduler objects. Sorry this turned out to be much bigger than I anticipated :confused:

### DIFF
--- a/otter/controller.py
+++ b/otter/controller.py
@@ -54,7 +54,7 @@ from otter.convergence.service import (
     trigger_convergence)
 from otter.json_schema.group_schemas import MAX_ENTITIES
 from otter.log import audit
-from otter.log.intents import msg, with_log
+from otter.log.intents import BoundFields, msg, with_log
 from otter.models.intents import GetScalingGroupInfo, ModifyGroupStatePaused
 from otter.models.interface import GroupNotEmptyError, ScalingGroupStatus
 from otter.supervisor import (
@@ -304,7 +304,7 @@ def empty_group(log, trans_id, group):
 
 
 @defer.inlineCallbacks
-def modify_and_trigger(group, log, modifier, *args, **kwargs):
+def modify_and_trigger(dispatcher, group, logargs, modifier, *args, **kwargs):
     """
     Modify group state and trigger convergence after that
 
@@ -320,8 +320,10 @@ def modify_and_trigger(group, log, modifier, *args, **kwargs):
     except CannotExecutePolicyError as ce:
         cannot_exec_pol_err = ce
     if tenant_is_enabled(group.tenant_id, config_value):
-        cs = get_convergence_starter()
-        yield cs.start_convergence(log, group.tenant_id, group.uuid)
+        eff = Effect(
+            BoundFields(
+                trigger_convergence(group.tenant_id, group.uuid), logargs))
+        yield perform(dispatcher, eff)
     if cannot_exec_pol_err is not None:
         raise cannot_exec_pol_err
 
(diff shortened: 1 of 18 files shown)


manishtomar:

cyli:

otter/controller.py
```diff
@@ -303,6 +302,31 @@ def modify_state(group_info):
...
+
+    :return: Deferred with None
+    """
+    cannot_exec_pol_err = None
+    try:
+        yield group.modify_state(modifier, *args, **kwargs)
+    except CannotExecutePolicyError as ce:
+        cannot_exec_pol_err = ce
+    if tenant_is_enabled(group.tenant_id, config_value):
+        eff = Effect(
```
radix: why not use with_log here?

otter/controller.py
```diff
@@ -303,6 +302,31 @@ def modify_state(group_info):
...
+            BoundFields(
+                trigger_convergence(group.tenant_id, group.uuid), logargs))
+        yield perform(dispatcher, eff)
+    if cannot_exec_pol_err is not None:
```
radix: can you explain why we need this unusual error handling? Is this a change in behavior? i.e. why do we need to converge even when CannotExecutePolicyError is raised? What about different exceptions?

otter/test/rest/request.py
```diff
@@ -313,6 +313,25 @@ def test_invalid_methods_are_405(self):
             self.assert_status_code(405, method=method)
 
 
+def setup_mod_and_trigger(testcase):
```
radix: Is there a reason this can't just be added to the mixin's setUp?

radix:
looks pretty good, just have some questions :)

otter/controller.py
manishtomar: The current expectation is convergence will still be triggered if policy execution or min/max change does not affect current desired. This is being [tested](https://github.com/rackerlabs/otter/blob/master/otter/integration/tests/test_convergence.py#L553-L602)

otter/controller.py
manishtomar: `with_log` seems to be useful when specific keyword arguments are known. Here, since I already have all the args I thought of using `BoundFields` directly. Sorry I assumed it would be _cheaper_ to do this. Premature optimization :)

otter/controller.py
cyli: Slightly more context for this currently expected behavior:  We don't distinguish between the two following cases of `CannotExecutePolicyError` on `delta=0`: 

1.  The user is already at their max, and hence can't scale up.  (similarly with min)
2.  The user tries to execute a policy which sets the desired capacity to the current desired capacity.

I think it makes sense in the first case to not converge.  I can see wanting to converge in the second case.

(I'm not arguing one way or another - @manishtomar and I discussed just removing that test expectation instead, but I think we decided it'd be easier to keep current functionality for now, and if we decided to change the behavior, do it in a different PR).

otter/test/rest/request.py
manishtomar: The main reason was that not all tests inheriting from mixin need to mock the controller. Ideally, it shouldn't affect even if it is mocked but I thought of keeping it separate for this reason.

otter/controller.py
radix: Sounds good :)

radix:
LGTM :)

PR #1695, opened 2015-08-31
Metrics timely tenants

### LINKED ISSUE #1691: Add more tenants in metrics on timely basis
Metrics service should start polling say 5 more tenants every day to slowly increase load on Nova. Eventually, after some number of days all tenants will be enabled.

manishtomar:
Closes #1691. Every day 5 new tenants info are fetched from Nova and ingested in CM. It does this by storing number of non-convergence tenants fetched till now along with last time they were fetched in a file. 

Also did some extra stuff:
- Number of tenants and number of groups are also ingested as metrics
- `LogErr` intent handles failure as `exc_info` tuple also
- Effect based file IO in `util/fileio.py`

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -126,7 +126,7 @@ def get_scaling_groups(client):
     """
     Get valid scaling groups grouped on tenantId from Cassandra
 
-    :param :class:`silverber.client.CQLClient` client: A cassandra client
+    :param :class:`silverberg.client.CQLClient` client: A cassandra client
     :return: `Deferred` fired with ``dict`` of the form
         {"tenantId1": [{group_dict_1...}, {group_dict_2..}],
          "tenantId2": [{group_dict_1...}, {group_dict_2..}, ...]}
@@ -144,7 +144,7 @@ def get_scaling_group_rows(client, props=None, batch_size=100):
     dict has 'tenantId', 'groupId', 'desired', 'active', 'pending'
     and any other properties given in `props`
 
-    :param :class:`silverber.client.CQLClient` client: A cassandra client
+    :param :class:`silverberg.client.CQLClient` client: A cassandra client
     :param ``list`` props: List of extra properties to extract
     :param int batch_size: Number of groups to fetch at a time
     :return: `Deferred` fired with ``list`` of ``dict``
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -14,7 +14,7 @@ from functools import partial
 
 import attr
 
-from effect import Effect, Func, TypeDispatcher, ComposedDispatcher
+from effect import ComposedDispatcher, Effect, Func, TypeDispatcher
 from effect.do import do, do_return
 
 from silverberg.client import ConsistencyLevel
@@ -22,8 +22,6 @@ from silverberg.cluster import RoundRobinCassandraCluster
 
 from toolz.curried import filter, get_in, groupby
 from toolz.dicttoolz import keyfilter, merge
-from toolz.functoolz import curry, identity
-from toolz.itertoolz import mapcat
 
 from twisted.application.internet import TimerService
 from twisted.application.service import Service
@@ -42,7 +40,7 @@ from otter.log import log as otter_log
 from otter.log.intents import err
 from otter.util.fileio import (
     ReadFileLines, WriteFileLines, get_dispatcher as file_dispatcher)
-from otter.util.fp import partition_bool, predicate_all
+from otter.util.fp import partition_bool
 from otter.util.timestamp import datetime_to_epoch
 
 
@@ -311,6 +309,7 @@ def add_to_cloud_metrics(ttl, region, total_desired, total_actual,
             for metric, value in totals]
     yield service_request(ServiceType.CLOUD_METRICS_INGEST,
                           'POST', 'ingest', data=data, log=log)
+
 
 def connect_cass_servers(reactor, config):
     """
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -9,17 +9,21 @@ import operator
 import sys
 import time
 from collections import namedtuple
-from datetime import timedelta
+from datetime import datetime, timedelta
 from functools import partial
 
-from effect import Effect
+import attr
+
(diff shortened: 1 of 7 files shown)


otter/test/utils.py
```diff
@@ -899,6 +899,18 @@ def noop(_):
     pass
 
 
+def const(v):
```
cyli: Very nice!  I've been thinking about adding a function like this for a while!  Thank you, it will be very useful!! :)

otter/metrics.py
```diff
@@ -44,54 +50,114 @@
...
+    return d.addCallback(filter(valid_group_row))
+
+
+def get_last_info(fname):
+    eff = Effect(ReadFileLines(fname)).on(
+        lambda lines: (int(lines[0]),
+                       datetime.utcfromtimestamp(float(lines[1]))))
+
+    def log_and_return(e):
+        _eff = err(e, "error reading last tenant")
```
cyli: "error reading last number of tenants"?  Since It seems to be the length of the tenants, as opposed to the actual tenant ID?

otter/metrics.py
```diff
@@ -44,54 +50,114 @@
...
+        return _eff.on(lambda _: (None, None))
+
+    return eff.on(error=log_and_return)
+
+
+def update_last_info(fname, tenants_len, time):
+    eff = Effect(
+        WriteFileLines(
+            fname, [tenants_len, datetime_to_epoch(time)]))
+    return eff.on(error=lambda e: err(e, "error updating last tenant"))
```
cyli: Similarly, "error updating the previous number of tenants"?

otter/metrics.py
manishtomar: Yes. Thanks! Same for other error too :)

otter/metrics.py
```diff
@@ -44,54 +50,114 @@
...
+    Get tenants that are enabled till today
+    """
+    batch_size = 5
+    tenants = sorted(tenants)
+    if last_tenants_len is None:
+        return tenants[:batch_size], batch_size, today
+    days = (today - last_date).days
+    if days <= 0:
+        return tenants[:last_tenants_len], last_tenants_len, today
+    if len(tenants) < last_tenants_len + batch_size:
```
cyli: Non-blocking: the return value as a result of this `if` statement seems to basically be the same logic as the return below it.  Essentially:

```
tenants = tenants[:last_tenants_len + batch_size]
return (tenants, len(tenants), today)
```

manishtomar: The length returned in last line is `last_tenants_len + batch_size` where here is `len(tenants)` since it has reached the limit.

otter/metrics.py
```diff
@@ -44,54 +50,114 @@
...
+    yield update_last_info(fname, last_tenants_len, last_date)
+    yield do_return(
+        keyfilter(lambda t: t in set(tenants + convergence_tids), groups))
+
+
+def get_scaling_groups(client):
+    """
+    Get valid scaling groups grouped on tenantId from Cassandra
+
+    :param :class:`silverber.client.CQLClient` client: A cassandra client
```
cyli: typo nitpick: silverber**g**.client.CQLClient :)

otter/metrics.py
cyli: Right, they just seem to be equivalent.  Since if there are fewer tenants than the `last_tenants_len + batch_size`, you just return the tenants that are there, which `tenants = tenants[:last_tenants_len + batch_size]` would also return:

```
>>> [1, 2, 3][:50]
[1, 2, 3]
```

And the length is `len(tenants)`.

If there are more than `last_tenants_len + batch_size` tenants, then `tenants = tenants[:last_tenants_len + batch_size]` would truncate them, and `len(tenants) == last_tenants_len + batch_size`, so the logic seems to work for both cases.

otter/metrics.py
```diff
@@ -44,54 +50,114 @@
...
-@defer.inlineCallbacks
+def valid_group_row(row):
+    return (
+        row.get('created_at') is not None and
+        row.get('desired') is not None and
+        row.get('status') not in ('DISABLED', 'ERROR') and
+        not row.get('deleting', False))
+
+
 def get_specific_scaling_groups(client, tenant_ids):
```
cyli: I think I may just be missing it - is this function still used by metrics?

manishtomar: No. Not really, I just left it in case it may be required later. Should I just delete it?

otter/metrics.py
```diff
@@ -229,19 +292,23 @@ def add_to_cloud_metrics(ttl, region, total_desired,
         there in the region
     :param int total_pending: Total number of servers currently
         building in a region
+    :param int no_tenants: total number of tenants
+    :param int no_tenants: total number of groups
```
cyli: `:param int no_groups:`  :)

otter/metrics.py
manishtomar: Eesh. Sorry, I didn't notice the `tenants = tenants[:last_tenants_len + batch_size]` line above. Good call :). Will change it.

manishtomar: It also makes the code more readable :)

otter/test/test_metrics.py
```diff
@@ -447,6 +422,127 @@ def test_diverged(self):
...
+            ([{"tenantId": "t1", "a": "1"}, {"tenantId": "t1", "a": "2"}] +
+             [{"tenantId": "t{}".format(i), "b": str(i)}
+              for i in range(2, 10)]))
+
+    def test_success(self):
+        """
+        Returns todays scaling groups based on number of tenants fetched
+        since last time. Updates the current fetch in file
+        """
+        seq = [
```
cyli: So neat :D  I love that it tests both the convergence IDs + the additional tenants in each test.

otter/metrics.py
cyli: We can always bring it back using git history if we need it back for something later.  :)

cyli:
LGTM other than the above nitpicks, which are mostly docs :D

manishtomar:
Thanks @cyli . Merging as I've addressed your comments.

PR #1696, opened 2015-09-01
Scheduler CF logging

manishtomar:
cloud feed event is logged every time a scheduled policy is executed

### DIFF
--- a/otter/scheduler.py
+++ b/otter/scheduler.py
@@ -198,7 +198,7 @@ def execute_event(store, log, event, deleted_policy_ids):
     policy_id = event['policyId']
     log = log.bind(tenant_id=tenant_id, scaling_group_id=group_id,
                    policy_id=policy_id)
-    log.msg('Scheduler executing policy {policy_id}')
+    log.msg('Scheduler executing policy {policy_id}', cloud_feed=True)
     group = store.get_scaling_group(log, tenant_id, group_id)
     d = group.modify_state(
         partial(maybe_execute_scaling_policy,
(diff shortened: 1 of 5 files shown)


otter/scheduler.py
```diff
@@ -197,8 +197,9 @@ def execute_event(store, log, event, deleted_policy_ids):
     group_id = event['groupId']
     policy_id = event['policyId']
     log = log.bind(tenant_id=tenant_id, scaling_group_id=group_id,
-                   policy_id=policy_id)
-    log.msg('Scheduler executing policy {policy_id}')
+                   policy_id=policy_id,
+                   scheduled_time=event["trigger"].isoformat() + "Z")
+    log.msg('sch-exec-pol', cloud_feed=True)
```
cyli: Should this be `'sch-exec-policy'`, since that's what's been registered in spec?

manishtomar: > Should this be 'sch-exec-policy', since that's what's been registered in spec? (or it can just be changed in the spec :))

Ooh. Thanks for catching that :) I'll change the spec. 

> Also, we seem to use cloudfeeds.cf_msg to do logging with cloud_feed=True - I'm not sure if we want to continue trying to use that abstraction?

`cf_msg` is Effect based and this is regular logging. I was thinking of having separate function for regular logging but skipped it as it was only adding `cloud_feed=True`. Also, like you said will need to change `ignore_and_log` to use that function.

cyli: > cf_msg is Effect based and this is regular logging. I was thinking of having separate function for regular logging but skipped it as it was only adding cloud_feed=True. Also, like you said will need to change ignore_and_log to use that function.

Makes sense!

cyli:
LGTM!

PR #1697, opened 2015-09-01
Log filter

cyli:
Not sure if this would be useful to anyone, but I've been using this to read through logs to find logs relevant to test failures.


manishtomar:
Thanks @cyli :)

PR #1699, opened 2015-09-02
Test self-healing on disowned servers

cyli:
These tests do not pass due to #1698.

I have a fix for that issue, which does cause these tests to pass.

Fixes #1486.
Fixes #1487.

PR #1700, opened 2015-09-02
Fix scrubbing otter metadata

### LINKED ISSUE #1698: Disowning a server (deleting without purging) fails
Scrubbing the otter metadata of the rax autsocaling keys fails because it expects to be scrubbing something like:

```
{
   "rax:autoscale:group:id": ....
}
```

And is instead scrubbing:
```
{
   "metadata": {
      "rax:autoscale:group:id": ....
   }
}
```

So it is still recognized as a server in the scaling group.  If `replace=False`, then some arbitrary server is deleted instead to make up for the change in desired capacity.

cyli:
Fixes #1698.

Have tested with #1699, and tests pass with this fix.  Either PR can be merged first.

### DIFF
--- a/otter/test/worker/test_launch_server_v1.py
+++ b/otter/test/worker/test_launch_server_v1.py
@@ -2144,10 +2144,10 @@ class MetadataScrubbingTests(SynchronousTestCase):
         keys and correctly keeps other keys.
         """
         samples = [
-            ({}, {}),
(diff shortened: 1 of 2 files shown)


cyli:
Helloooo travis?

manishtomar:
LGTM :)

PR #1702, opened 2015-09-02
Support many more server states

### LINKED ISSUE #1694: Handle all the different Nova server states, also unknown Nova states
@sam-falvo Found a bug while scaling up 300 servers where convergence stopped working for a while.

This was due to at least one of the servers being in a shutoff state, which we do not currently handle in our server model.  Our server model attempts to convert the Nova server status to `otter.convergence.model.ServerState`, and there are only 5 states, so the name lookup fails.  This caused gathering to repeatedly fail until all the SHUTOFF servers were removed.

(I do actually mean `SHUTOFF`, even though it's undocumented in Nova's [list of server states](http://docs.rackspace.com/servers/api/v2/cs-devguide/content/List_Servers_Server_Statuses-d1e2078.html#server_status):

```
  ...
from_server_details_json
    server_state = ServerState.lookupByName(server_json['status'])
  File "/opt/otter/.ve/local/lib/python2.7/site-packages/twisted/python/constants.py", line 258, in lookupByName
    raise ValueError(name)
effect._intents.FirstError: (index=0) ValueError: SHUTOFF
```
)

Since `otter.convergence.model.ServerState` reflects the Autoscale conception of various state a server can be in (Nova does not have DRAINING, for instance, and Autoscale does not care about transitional states like MIGRATING, because Autoscale can take no action), we should probably translate those other unhandled Nova states to `ACTIVE`, `BUILD`, `ERROR`, `DRAINING`, or `DELETE`.

Most of the Nova states are transitional, and we can maybe translate them to `BUILD` state (since it means the server isn't active, but it's not in error or being deleted).

However, we're not sure how to handle `SUSPENDED` and `UNKNOWN`.  @manishtomar made the great suggestion of just disowning them and replacing them, since `SUSPENDED` can be by user request.  We may want to handle undocumented Nova states this way too (for instance, the `SHUTOFF` state that @sam-falvo noticed in his test).

radix:
Fixes #1694.

- Added a bunch of definitions to ServerState for all the rest of the states we know about
- added a new enum, "Destiny", which indicates what we want to _do_ with a server, separate from its state
- map states to destinies

Actual behavior changes:
- previously unhandled states are:
  - HARD_REBOOT -> WAIT
  - MIGRATING -> WAIT
  - PASSWORD -> WAIT
  - REBUILD -> WAIT
  - RESIZE -> WAIT
  - REVERT_RESIZE -> WAIT
  - RESCUE -> DO_NOT_REPLACE
  - VERIFY_RESIZE -> DO_NOT_REPLACE
  - SUSPENDED -> DO_NOT_REPLACE
  - SHUTOFF -> DELETE
  - UNKNOWN -> IGNORE

- The behavior of `Destiny.WAIT` is similar to how we handle `BUILD`, but without the timeout.
- The behavior of `Destiny.DO_NOT_REPLACE` is mostly the same as `ACTIVE`, except no load balancer convergence will happen for those nodes.

Request for comments on:
- A thought occurred to me as I was writing this description: perhaps we should explicitly _remove_ `WAIT` and `DO_NOT_REPLACE` servers from the load balancers?
- Better names for destinies?

[Review on Reviewable](https://reviewable.io/reviews/rackerlabs/otter/1702)

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -73,9 +73,9 @@ class ServerState(Names):
     Most of these constants correspond to Nova states of the same name.
     """
     ACTIVE = NamedConstant()
-    ERROR = NamedConstant()
     BUILD = NamedConstant()
     DELETED = NamedConstant()
+    ERROR = NamedConstant()
     HARD_REBOOT = NamedConstant()
     MIGRATING = NamedConstant()
     PASSWORD = NamedConstant()
@@ -83,9 +83,10 @@ class ServerState(Names):
     RESCUE = NamedConstant()
     RESIZE = NamedConstant()
     REVERT_RESIZE = NamedConstant()
-    VERIFY_RESIZE = NamedConstant()
     SHUTOFF = NamedConstant()
+    SUSPENDED = NamedConstant()
     UNKNOWN = NamedConstant()
+    VERIFY_RESIZE = NamedConstant()
 
     DRAINING = NamedConstant()
     """"
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -69,22 +69,23 @@ class CLBNodeType(Names):
 class ServerState(Names):
     """
     Constants representing the state of a Nova cloud server.
-    """
 
+    Most of these constants correspond to Nova states of the same name.
+    """
     ACTIVE = NamedConstant()
-    """
-    Corresponds to Nova's ``ACTIVE`` state.
-    """
-
     ERROR = NamedConstant()
-    """
-    Corresponds to Nova's ``ERROR`` state.
-    """
-
     BUILD = NamedConstant()
-    """
-    Corresponds to Nova's ``BUILD`` and ``BUILDING`` states.
-    """
+    DELETED = NamedConstant()
+    HARD_REBOOT = NamedConstant()
+    MIGRATING = NamedConstant()
+    PASSWORD = NamedConstant()
+    REBUILD = NamedConstant()
+    RESCUE = NamedConstant()
+    RESIZE = NamedConstant()
+    REVERT_RESIZE = NamedConstant()
+    VERIFY_RESIZE = NamedConstant()
+    SHUTOFF = NamedConstant()
+    UNKNOWN = NamedConstant()
 
     DRAINING = NamedConstant()
     """"
@@ -93,9 +94,10 @@ class ServerState(Names):
     is basically functional, but that Autoscale would like to delete it.
     """
 
-    DELETED = NamedConstant()
+    UNKNOWN_TO_OTTER = NamedConstant()
     """
-    Corresponds to Nova's ``DELETED`` state.
+    Indicates that some state was returned by Nova that Otter doesn't know
+    about. The real state will be on the NovaServer object.
     """
 
 
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -97,7 +97,7 @@ class ServerState(Names):
     UNKNOWN_TO_OTTER = NamedConstant()
     """
     Indicates that some state was returned by Nova that Otter doesn't know
-    about. The real state will be on the NovaServer object.
+    about. The real state will be in `NovaServer.json`.
     """
 
 
@@ -256,7 +256,10 @@ class NovaServer(object):
 
         :return: :obj:`NovaServer` instance
         """
-        server_state = ServerState.lookupByName(server_json['status'])
+        try:
+            server_state = ServerState.lookupByName(server_json['status'])
+        except ValueError:
+            server_state = ServerState.UNKNOWN_TO_OTTER
         if server_json.get("OS-EXT-STS:task_state", "") == "deleting":
             server_state = ServerState.DELETED
         metadata = server_json.get('metadata', {})
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from pyrsistent import pbag, pset
 
 from toolz.curried import filter
-from toolz.itertoolz import concat, groupby, mapcat
+from toolz.itertoolz import groupby, mapcat
 
 from twisted.python.constants import NamedConstant, Names
 
@@ -236,14 +236,8 @@ _DESTINY_TO_STATES = {
         ServerState.UNKNOWN_TO_OTTER],
 }
 
-# Ensure we've covered all the server states
-for st in ServerState.iterconstants():
-    assert st in concat(_DESTINY_TO_STATES.values()), st
-# Ensure states only map to one destiny
-_all_destiny_states = list(concat(_DESTINY_TO_STATES.values()))
-assert len(_all_destiny_states) == len(set(_all_destiny_states))
 
-STATE_TO_DESTINY = {
+_STATE_TO_DESTINY = {
     state: destiny
     for destiny, states in _DESTINY_TO_STATES.iteritems()
     for state in states}
@@ -251,7 +245,7 @@ STATE_TO_DESTINY = {
 
 def get_destiny(server):
     """Get the obj:`Destiny` of a server."""
-    return STATE_TO_DESTINY.get(server.state)
+    return _STATE_TO_DESTINY.get(server.state)
 
 
 def converge(desired_state, servers_with_cheese, load_balancer_contents, now,
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -1,9 +1,13 @@
 """Code related to creating a plan for convergence."""
 
+from collections import defaultdict
+
 from pyrsistent import pbag, pset
 
 from toolz.curried import filter
-from toolz.itertoolz import mapcat
+from toolz.itertoolz import concat, groupby, mapcat
+
+from twisted.python.constants import NamedConstant, Names
 
 from otter.convergence.model import (
     CLBDescription,
@@ -27,7 +31,7 @@ from otter.convergence.steps import (
     SetMetadataItemOnServer,
 )
 from otter.convergence.transforming import limit_steps_by_count, optimize_steps
-from otter.util.fp import partition_bool, partition_groups
+from otter.util.fp import partition_bool
 
 
 def _remove_from_lb_with_draining(timeout, nodes, now):
@@ -165,6 +169,55 @@ def _drain_and_delete(server, timeout, current_lb_nodes, now):
     return lb_draining_steps
 
 
+class Destiny(Names):
+    """
+    What otter will be doing with a server.
+    """
+    CONSIDER_AVAILABLE = NamedConstant()
+    WAIT_WITH_TIMEOUT = NamedConstant()
+    WAIT = NamedConstant()
+    DRAIN = NamedConstant()
+    DELETE = NamedConstant()
+    CLEANUP = NamedConstant()
+    IGNORE = NamedConstant()
+
+
+DESTINY_TO_STATES = {
+    Destiny.CONSIDER_AVAILABLE: [ServerState.ACTIVE],
+    Destiny.WAIT_WITH_TIMEOUT: [ServerState.BUILD],
+    Destiny.WAIT: [
+        ServerState.HARD_REBOOT,
+        ServerState.MIGRATING,
+        ServerState.PASSWORD,
+        ServerState.REBUILD,
+        ServerState.RESCUE,
+        ServerState.RESIZE,
+        ServerState.REVERT_RESIZE,
+        ServerState.VERIFY_RESIZE,
+        ],
+    Destiny.DRAIN: [ServerState.DRAINING],
+    Destiny.DELETE: [ServerState.SHUTOFF, ServerState.ERROR],
+    Destiny.CLEANUP: [ServerState.DELETED],
+    Destiny.IGNORE: [
+        ServerState.UNKNOWN,
+        ServerState.UNKNOWN_TO_OTTER],
+}
+
+for st in ServerState.iterconstants():
+    assert st in concat(DESTINY_TO_STATES.values()), st
+
+_all_destiny_states = list(concat(DESTINY_TO_STATES.values()))
+assert len(_all_destiny_states) == len(set(_all_destiny_states))
+
+STATES_TO_DESTINY = {
+    state: destiny
+    for destiny, states in DESTINY_TO_STATES.iteritems()
+    for state in states}
+
+
+get_destiny = STATES_TO_DESTINY.get
+
+
 def converge(desired_state, servers_with_cheese, load_balancer_contents, now,
              timeout=3600):
     """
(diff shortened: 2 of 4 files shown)


radix:

cyli:
Non-blocking questions:  

- This means that we count things in `WAIT_WITH_TIMEOUT`, `WAIT`, and `DO_NOT_REPLACE` all as pending capacity in the state?  (we are only counting servers specifically in `ACTIVE` as active).

- Should we maybe log to CF if there is an increase in the number of `DO_NOT_REPLACE` servers, since the customer will need to do something about it and if they don't know that they do (since we don't surface it) then it will look (to the customer) like autoscale just remains in building forever?

As for explicitly removing `WAIT` and `DO_NOT_REPLACE` servers from the load balancers - hopefully the load balancer health checks can detect those states?  Maybe we should do it anyway, but it seems like a separate feature that can happen later.  Might be good to open an issue to track it, though.

And I'm terrible at naming - the current destiny names seem great to me :)

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[otter/convergence/planning.py, line 239 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEo5BGjnfM5yzLIATw)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L239)):</sup>
Out of curiosity, why assertions here, rather than in a unit test?

---

<sup>**[otter/convergence/planning.py, line 306 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L306)):</sup>
Non-blocking - just wondering about `DO_NOT_REPLACE` - I presume we don't want to replace them because we don't want just delete them.  If that's the case, would it make sense for this to be `servers[Destiny.DO_NOT_REPLACE] + servers_in_active + servers[Destiny...`?

(Maybe not, just checking though)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[otter/convergence/planning.py, line 239 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEo5BGjnfM5yzLIATw)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L239)):</sup>
I guess my mindset was just "post-conditions" or "contracts" rather than "unit tests" when I was writing those assertions. I could definitely change them to be unit tests.

---

<sup>**[otter/convergence/planning.py, line 306 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L306)):</sup>
I was going to say that I think this would allow a scenario where a group has only DO_NOT_REPLACE servers, which seems bad...

But I realize that with this code that can happen _anyway_, if all CONSIDER_ACTIVE servers become DO_NOT_REPLACE. which is probably really bad... Maybe we should treat all these as IGNORE? That would mean we provision new servers to replace them. I'm just worried about overprovisioning servers against the user's will and charging them too much... but then I guess maximums can help alleviate this...

Clouds are hard :(

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

cyli:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[otter/convergence/planning.py, line 306 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L306)):</sup>
By "maximums" do you mean that scaling up would fail because we'd count `IGNORE` or `DO_NOT_REPLACE` as kind of a shadow capacity (since it's not active or pending) that blocks scaling up?  Or do you mean the Nova quota maximums?

But yeah, I dunno.  :|  Hence I was thinking maybe we should alert users any time the number of `DO_NOT_REPLACE` or `IGNORE` servers goes up, so they can do something about it? 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

cyli:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[otter/convergence/planning.py, line 239 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEo5BGjnfM5yzLIATw)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L239)):</sup>
I don't feel particularly strongly.  :)  Unit tests make more sense to me, but it's more like a +0.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:

radix:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[otter/convergence/planning.py, line 239 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEo5BGjnfM5yzLIATw)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L239)):</sup>
Okay, I'll change it. :)

---

<sup>**[otter/convergence/planning.py, line 306 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L306)):</sup>
No, I was just referring to max servers in a group.

Definitely agree about adding the CF logs. I think I might do that in another PR though. I'll look into it.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:

radix:
To answer cyli's questions:
- yes, WAIT, WAIT_WITH_TIMEOUT, and DO_NOT_REPLACE all act as pending capacity.
- yes, I agree we should be logging to CF. I think I'll do that in another PR.

fwiw I renamed DO_NOT_REPLACE to AVOID_REPLACING, since there are still situations in which a server will be deleted.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[otter/convergence/planning.py, line 239 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEo5BGjnfM5yzLIATw)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L239)):</sup>
Done.

---

<sup>**[otter/convergence/planning.py, line 306 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L306)):</sup>
Okay, so I think I'm just going to leave this as-is unless @manishtomar has a better idea. I think there's only so much we can do with these cloud statuses, and logging the fact that we've observed them in CF should give the user the alert they need to do something about them.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:

manishtomar:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

<sup>**[otter/convergence/planning.py, line 186 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJJNTfdnrg0_zjDh1z)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L186)):</sup>
I am guessing we wait indefinitely for this server to become active. If so, can this be renamed to `WAIT_INDEFINITELY`? Also, is it practical to wait forever? Ideally, this are transient states which _should_ change but we never know what can happen with Nova :)

---

<sup>**[otter/convergence/planning.py, line 211 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJNVg5qUJtIx4gzQiZ)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L211)):</sup>
I would've gone with STATES_TO_DESTINY :)

---

<sup>**[otter/convergence/planning.py, line 300 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L300)):</sup>
Given a choice between keeping CONSIDER_ACTIVE and AVOID_REPLACING server, I would go with keeping CONSIDER_ACTIVE server since from our perspective is healthy and working. In other words, this code is good :)

---

<sup>**[otter/convergence/planning.py, line 341 \[r2\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJXhTbrGKaZqQ8zg9_)** ([raw file](https://github.com/rackerlabs/otter/blob/f2c85506fa6ec9451835601fbfde56a6aa4124fd/otter/convergence/planning.py#L341)):</sup>
I am worried about convergence running forever due to servers remaining in `Destiny.WAIT` forever. Honestly, I do not trust Nova to ensure servers will become normal. I think we should put the group in ERROR if `Destiny.WAIT` servers do not become ACTIVE within a certain time. It can be done in another PR though :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

<sup>**[otter/convergence/planning.py, line 186 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJJNTfdnrg0_zjDh1z)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L186)):</sup>
Given how much pushback we've had from other people about timing out *builds* I'd rather not introduce more timeouts. I think logging with CF that we've observed these states is probably good enough, and customers using monitoring is probably the best solution to these situations.  That's my thought for now, anyway :P

---

<sup>**[otter/convergence/planning.py, line 211 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJNVg5qUJtIx4gzQiZ)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L211)):</sup>
not sure what you mean. there is already a _STATE_TO_DESTINY below which is the same data inverted. _DESTINY_TO_STATES maps destiny constants to lists of states. i.e. "x_to_y" describes a {x: y} map

---

<sup>**[otter/convergence/planning.py, line 300 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyEtltiX2hH6BGJgZpS)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L300)):</sup>
Thanks for the input :)

---

<sup>**[otter/convergence/planning.py, line 341 \[r2\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJXhTbrGKaZqQ8zg9_)** ([raw file](https://github.com/rackerlabs/otter/blob/f2c85506fa6ec9451835601fbfde56a6aa4124fd/otter/convergence/planning.py#L341)):</sup>
Hmm, yeah, that makes sense. So we wouldn't delete the servers, we would just put the group into error. That makes sense. I'll create another ticket for that. We could perhaps even do the same with BUILD servers? (given nova complains about us timing out servers anyway).

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

manishtomar:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[otter/convergence/planning.py, line 211 \[r1\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJNVg5qUJtIx4gzQiZ)** ([raw file](https://github.com/rackerlabs/otter/blob/860c9b89181ca721743acac3d09b3a213b2e51b7/otter/convergence/planning.py#L211)):</sup>
Yeah. Sorry I saw that below. This is fine. I was just saying that I would've started the code with STATES_TO_DESTINY. 

---

<sup>**[otter/convergence/planning.py, line 341 \[r2\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJXhTbrGKaZqQ8zg9_)** ([raw file](https://github.com/rackerlabs/otter/blob/f2c85506fa6ec9451835601fbfde56a6aa4124fd/otter/convergence/planning.py#L341)):</sup>
> We could perhaps even do the same with BUILD servers?

Thats a good question. BUILD servers may be a different issue. I think we should talk to Nova before deciding to put group in ERROR if server is building for too long. I think for now, we can go ahead with deleting building servers and discuss it in the ticket. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[otter/convergence/planning.py, line 341 \[r2\]](https://reviewable.io:443/reviews/rackerlabs/otter/1702#-JyJXhTbrGKaZqQ8zg9_)** ([raw file](https://github.com/rackerlabs/otter/blob/f2c85506fa6ec9451835601fbfde56a6aa4124fd/otter/convergence/planning.py#L341)):</sup>
Okay, https://github.com/rackerlabs/otter/issues/1706.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1702)

radix:

manishtomar:
woohoo.. tests pass :)

manishtomar:
Merging

PR #1710, opened 2015-09-04
next day bugfix

manishtomar:
the current time was always returned for same day resulting in current time being stored in file every minute. Now, it will return old date for same day and today only for new batch. This way the file will be updated with new date at beginning of each day.


radix:
LGTM!

---

Reviewed 2 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/rackerlabs/otter/1710)

PR #1711, opened 2015-09-04
Move draining logic to planning

radix:
Get rid of the `ServerState.DRAINING` state. Make it so `get_destiny` is responsible for looking at server metadata to determine if the server should have the `DRAIN` destiny.

Pretty straightforward change, and it reduces LoC (mostly by way of the unit tests), so it seems good :)

unrelated change:
- made the `otter.test.utils.server` utility accept a `metadata` map for convenience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ subunit-output.txt
 *.pydev*
 .DS_Store
 docbook/target
+_trial_temp.lock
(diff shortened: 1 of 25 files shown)


manishtomar:

manishtomar:
LGTM :) once build +1s

radix:
Thanks manish!

PR #1715, opened 2015-09-08
Converge later if any of the servers are draining

### LINKED ISSUE #1256: If load balancers are draining, and nothing else can be done, also return `ConvergeLater`
Maybe we can have an optimization step that concatenates all the `ConvergeLater`s into 1 `ConvergeLater`

manishtomar:
Closes #1256

### DIFF
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -84,7 +84,11 @@ def _remove_from_lb_with_draining(timeout, nodes, now):
 
     changes = [drain_lb_node(node=node) for node in to_drain]
 
-    return removes + changes
+    retry = (
+        [ConvergeLater(reasons=[ErrorReason.String('draining servers')])]
+        if in_drain else [])
+
+    return removes + changes + retry
 
 
 def _converge_lb_state(server, current_lb_nodes):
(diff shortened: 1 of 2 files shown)


otter/convergence/planning.py
```diff
@@ -84,7 +84,11 @@ def _remove_from_lb_with_draining(timeout, nodes, now):
 
     changes = [drain_lb_node(node=node) for node in to_drain]
 
-    return removes + changes
+    retry = (
+        [ConvergeLater(reasons=[ErrorReason.String('draining servers')])]
+        if in_drain else [])
```
radix: So I've been considering if we should be creating one of these per server. It would mean we can clearly see which server caused which step (if we include a message like `'draining server %s' % (server.id,)`). But maybe that would make the log messages too long. What do you think?

(I was thinking about this since I'm also adding another case of ConvergeLater in another PR).

radix:
LGTM! :) You can merge this. Let me know what you think about my comment.

otter/convergence/planning.py
manishtomar: Sounds like a good idea. However I am worried the number of steps will be too much when there are many servers. I think we can leave it like this and implement this or add separate log inside `log_steps` if such a need arises.

manishtomar:
Thanks for the review :)

PR #1716, opened 2015-09-09
Convergence gathering logging

manishtomar:
Logging beginning of convergence cycle and time it takes to gather all the data.

### DIFF
--- a/otter/log/spec.py
+++ b/otter/log/spec.py
@@ -102,7 +102,7 @@ msg_types = {
     # event -> [(event, format_str)]
     "add-server-clb": ("Adding {server_id} with IP address {ip_address} "
                        "to CLB {clb_id}"),
-    "beginning-convergence": "Beginning to execute convergence",
+    "begin-convergence": "Beginning to execute convergence",
     "converge-all-groups": "Attempting to converge all dirty groups",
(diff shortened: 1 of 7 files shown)


manishtomar:

otter/log/spec.py
```diff
@@ -102,6 +102,7 @@ def length_calc(e):
     # event -> [(event, format_str)]
     "add-server-clb": ("Adding {server_id} with IP address {ip_address} "
                        "to CLB {clb_id}"),
+    "beg-exec-convergence": "Beginning to execute convergence",
```
radix: Just nitpicks: given that we don't do abbreviations so far (plus "beg" kinda sounds like begging ;) I think it'd be better to call this 'begin-convergence', which is just as short

otter/log/spec.py
```diff
@@ -115,6 +116,8 @@ def length_calc(e):
     "execute-convergence": split_execute_convergence,
     "execute-convergence-results": (
         "Got result of {worst_status} after executing convergence"),
+    "get-conv-exec-data": (
```
radix: and another nitpick: I think this could be called "gather-data"

radix:
LGTM! You can ignore my comments if you want :)

PR #1719, opened 2015-09-16
splitting list server details response

manishtomar:
Each part displays subset of response body since that is the biggest part of the event.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -328,13 +328,15 @@ def get_cloud_client_dispatcher(reactor, authenticator, log, service_configs):
 # ----- Logging responses -----
 
 
-def log_success_response(msg_type, response_body_filter):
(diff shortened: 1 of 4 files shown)


manishtomar:

otter/cloud_client.py
```diff
@@ -345,13 +347,14 @@ def _log_it(result):
         # So we can link it to any non-cloud_client logs
         request_id = resp.request.headers.getRawHeaders(
             'x-otter-request-id', [None])[0]
-
+        resp_body = (
+            json.dumps(response_body_filter(json_body), sort_keys=True)
```
radix: do you know why we wanted json in the first place? why not just always log it structured?

manishtomar: I am guessing due to ES mapping issues. This way it always remains string and doesn't confuse ES indexes if for some reason different kind of values are returned.

radix:
Thanks :)

PR #1737, opened 2015-09-24
force delete in rbac tests

manishtomar:
Since sometimes server from previous policy exec can still be building when test tries to delete the group. This fixes https://as-jenkins.rax.io/job/otter-cloudcafe-prod-iad/308/console failures.


lekhajee:
+1

PR #1742, opened 2015-09-29
include the group ID in exception messages from wait_for_group

radix:
just a really minor improvement to error messages


manishtomar:
LGTM! Thanks for doing this. I feel bad for not doing this already.

radix:
@manishtomar no problem. it's only a minor improvement really, since it was already in the test case log file -- this just makes it a little easier to correlate :)

PR #1724, opened 2015-09-17
Limit LIMITED_RETRY

radix:
This branch fixes #1706 the rest of the way.

I tore through the tests a bit more refactoring them.

PR #1743, opened 2015-09-30
add sumtype for status of execute_convergence

radix:
This simplifies our code so that the weird two-tuple of `(worst_status, group_status)` doesn't need to escape from `execute_convergence` and processed with some opaque behavior (which was my own fault for introducing!)

Now there's sumtype representing the various results that executing convergence can have: Continue, Stop, and GroupDeleted.

(I decided to do this when I found a bug in the code I introduced in #1724, in the handling of the result of execute_convergence.  This change makes that change a lot simpler, not to mention the existing code).

### DIFF
--- a/otter/convergence/model.py
+++ b/otter/convergence/model.py
@@ -125,6 +125,14 @@ class StepResult(Names):
 
 
 @sumtype
+class ConvergenceIterationStatus(object):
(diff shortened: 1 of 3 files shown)


radix:

radix:

manishtomar:
LGTM! :)

PR #1747, opened 2015-10-02
Add orchestration to service config

dragorosson:


stretchservers:

stretchservers:

otter/constants.py
```diff
@@ -36,6 +37,10 @@ def get_service_configs(config):
             'name': config['cloudLoadBalancers'],
             'region': config['region'],
         },
+        ServiceType.CLOUD_ORCHESTRATION: {
```
manishtomar: Can this service be optional just like metrics and cloudfeeds is? It will be nice to run otter without mandating this service.

manishtomar:
Is it possible to only have `otter/constants.py` and `otter/test/test_constants.py` in this PR since that is simple cohesive change to review? :)

otter/constants.py
radix: I actually suggested making it mandatory to reflect the fact that cloud servers is also mandatory here -- in order to facilitate normal otter API operation we will need the orchestration service running just as much as we will need cloud servers.

radix:
@manishtomar I think this PR is showing more than it should because it's based on drago's renaming branch. once that PR is merged into master and this branch is updated it should be smaller.

radix:

manishtomar:

manishtomar:
LGTM! Merging :)

PR #1748, opened 2015-10-02
Add Heat requests to cloud_client

dragorosson:


stretchservers:

stretchservers:

otter/cloud_client.py
```diff
@@ -1051,3 +1051,62 @@ def publish_to_cloudfeeds(event, log=None):
...
         data=event, log=log, success_pred=has_code(201),
         json_response=False)
+
+
+# ----- Cloud orchestration requests -----
+def list_stacks_all(parameters=None):
+    """
+    List Heat stacks.
+
+    :ivar dict parameters: Query parameters to include.
```
radix: This should be `param` not `ivar`

otter/cloud_client.py
```diff
@@ -1051,3 +1051,62 @@ def publish_to_cloudfeeds(event, log=None):
...
+        .on(log_success_response('request-list-stacks-all', identity))
+        .on(lambda (response, body): body['stacks'])
+    )
+
+
+def create_stack(stack_args):
+    """
+    Create a stack using Heat.
+
+    :ivar dict stack_args:  The dictionary to pass to Heat specifying how
```
radix: same here (param not ivar)

otter/cloud_client.py
```diff
@@ -1051,3 +1051,62 @@ def publish_to_cloudfeeds(event, log=None):
...
+    List Heat stacks.
+
+    :ivar dict parameters: Query parameters to include.
+
+    Succeed on 200.
+
+    :return: a `tuple` of (:obj:`twisted.web.client.Response`, JSON `dict`)
+    """
+
+    # TODO Actually implement batch handling
```
radix: Is this going to be necessary before we deploy this feature? I assume if a user has over N(=20? 100? 500?) stacks then this will only get the first batch?

otter/cloud_client.py
```diff
@@ -1051,3 +1051,62 @@ def publish_to_cloudfeeds(event, log=None):
...
+# ----- Cloud orchestration requests -----
+def list_stacks_all(parameters=None):
+    """
+    List Heat stacks.
+
+    :ivar dict parameters: Query parameters to include.
+
+    Succeed on 200.
+
+    :return: a `tuple` of (:obj:`twisted.web.client.Response`, JSON `dict`)
```
radix: This seems to be incorrect, it's just returning the `stacks`, not a two-tuple.

otter/cloud_client.py
```diff
@@ -1051,3 +1051,62 @@ def publish_to_cloudfeeds(event, log=None):
...
+
+
+def create_stack(stack_args):
+    """
+    Create a stack using Heat.
+
+    :ivar dict stack_args:  The dictionary to pass to Heat specifying how
+        the stack should be built.
+
+    :return: a `tuple` of (:obj:`twisted.web.client.Response`, JSON `dict`)
```
radix: I recommend just making this return the body too, no need to return the Response object. i.e., just return the second element of this tuple, like you did in `list_stacks_all`.

radix:
Looking pretty good. Just a few minor review points.

dragorosson:
@radix Yeah, it's rough still. I'm going to make this WIP still and revisit it after more prep work is done.

manishtomar:
@dragorosson You can use "in progress" label for WIP PRs.

otter/cloud_client.py
dragorosson: Done.

otter/cloud_client.py
dragorosson: Done.

otter/cloud_client.py
dragorosson: Not sure. I don't know that it's necessary from the Heat side. Should it be done for performance reasons on the autoscale side?

otter/cloud_client.py
dragorosson: Done, actually a list of dicts.

otter/cloud_client.py
dragorosson: Done.

dragorosson:
Ready for review!

otter/cloud_client.py
radix: I don't think there's any particular performance concern, I'm more worried about correctness. If the Heat API batches by default, and we're only processing the first batch, then otter will think that those stacks don't exist when they actually do.

I've looked at both the OpenStack Heat API docs (http://developer.openstack.org/api-ref-orchestration-v1.html) and the Rackspace ones (https://developer.rackspace.com/docs/cloud-orchestration/v1/developer-guide/) and neither of them document the pagination behavior of a query without `limit` supplied, so I don't know what will happen.

radix:
This PR looks good to go, but I'm still curious about what the default pagination behavior is. Might need to check the heat code, since I can't find any info in the docs.

dragorosson:
@radix the default pagination behavior (by omitting `limit`) is to return all stacks. https://github.com/openstack/oslo.db/blob/master/oslo_db/sqlalchemy/utils.py#L178-L179

radix:
Alright! Thanks for checking :)

PR #1749, opened 2015-10-02
Move group-fetching code from metrics.py to core model code

radix:
This is a fairly simple refactoring to move the get_scaling_groups and GetAllGroups intent into otter.models. I figure we will want to use this code in order to implement naive self-healing in the convergence codepath, so it should be in a more accessible place than otter.metrics.

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -12,15 +12,12 @@ from collections import namedtuple
 from datetime import datetime, timedelta
 from functools import partial
 
-import attr
-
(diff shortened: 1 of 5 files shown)


radix:

manishtomar:
Unfortunately, there are no integration tests for metrics. Let me run metrics on dev vm and see.

radix:
@manishtomar you're a saint!

manishtomar:
LGTM! Metrics works fine :)

PR #1751, opened 2015-10-02
allow space at end of message

manishtomar:
since sometimes the error has an extra space


radix:
thanks :) and sigh :(

PR #1752, opened 2015-10-02
Implement ConvergenceExecutor

dragorosson:
ConvergenceExecutor allows simple polymorphism when running convergence, which is useful for changing behavior depending on launch config type. Interestingly, the way this is implemented makes `ConvergenceExecutor`'s methods static, so I was able to directly replace free function calls with the executor version without having to add `self` to the original function.

Testing works almost the same way as before, except a `get_executor` method is passed which can return a custom executor that has the mock methods attached, because all of its methods can be overridden. (See how `_invoke` has changed and see `test_launch_server_overrides`)

The exact methods contained in `ConvergenceExecutor` may change. For example, just `UpdateServersCache` instead of the entirety of `convergence_succeeded_servers` since most of the code there isn't specific to launch_server.

@radix, @manishtomar, let me know what you think!

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -133,6 +133,9 @@ from otter.util.zk import CreateOrSet, DeleteNode, GetChildren, GetStat
 
 
 def get_executor(launch_config):
+    """
+    Returns a ConvergenceExecutor based upon the launch_config type given.
+    """
     if launch_config['type'] == 'launch_server':
         return launch_server_executor
     raise NotImplementedError
@@ -815,6 +818,10 @@ class Converger(MultiService):
 
 @attr.s
 class ConvergenceExecutor(object):
+    """
+    A bag of methods to enable configurable logic for different types of launch
+    configurations.
+    """
     gather = attr.ib()
     plan = attr.ib()
     get_desired_group_state = attr.ib()
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -90,6 +90,8 @@ from datetime import datetime
 from functools import partial
 from hashlib import sha1
 
+import attr
+
 from effect import Constant, Effect, Func, parallel
 from effect.do import do, do_return
 from effect.ref import Reference
@@ -130,6 +132,12 @@ from otter.util.timestamp import datetime_to_epoch
 from otter.util.zk import CreateOrSet, DeleteNode, GetChildren, GetStat
 
 
+def get_executor(launch_config):
+    if launch_config['type'] == 'launch_server':
+        return LaunchServerExecutor()
+    raise NotImplementedError
+
+
 def server_to_json(server):
     """
     Convert a NovaServer to a dict representation suitable for returning to the
@@ -162,7 +170,7 @@ def is_autoscale_active(server, lb_nodes):
                              if node.matches(server)]))
 
 
-def update_cache(group, servers, lb_nodes, now, include_deleted=True):
+def update_servers_cache(group, servers, lb_nodes, now, include_deleted=True):
     """
     Updates the cache, adding servers, with a flag if autoscale is active on
     each one.
@@ -220,7 +228,7 @@ def _execute_steps(steps):
 
 @do
 def convergence_exec_data(tenant_id, group_id, now,
-                          get_all_launch_server_data):
+                          get_executor):
     """
     Get data required while executing convergence
     """
@@ -232,7 +240,9 @@ def convergence_exec_data(tenant_id, group_id, now,
     group_state = manifest['state']
     launch_config = manifest['launchConfiguration']
 
-    gather_eff = get_all_launch_server_data(tenant_id, group_id, now)
+    executor = get_executor(launch_config)
+
+    gather_eff = executor.gather(tenant_id, group_id, now)
 
     (servers, lb_nodes) = yield gather_eff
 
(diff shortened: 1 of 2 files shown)


stretchservers:

stretchservers:

otter/convergence/service.py
```diff
@@ -799,3 +808,18 @@ def divergent_changed(self, children):
             # the return value is ignored, but we return this for testing
             eff = self._converge_all(my_buckets, children)
             return perform(self._dispatcher, self._with_conv_runid(eff))
+
+
+class ConvergenceExecutor(object):
+    gather = attr.ib()
+    plan = attr.ib()
+    get_desired_group_state = attr.ib()
+    convergence_succeeded = attr.ib()
```
manishtomar: Will `StackExecutor` not need `convergence_failed`?

radix: I think the thing here is that `convergence_failed` doesn't actually have any logic that's specific to either launch_server or heat stacks, whereas `convergence_succeeded` does -- by updating the servers cache.

I'm a little bit concerned about this factoring, and think there may be a better way to do it. For example, it looks like the *majority* of the code in `convergence_succeeded_servers` will need to be used in the heat case too, just not the servers-cache updating. So maybe it should be refactored so that `convergence_succeeded` is just a fixed function that's called (not dispatched in this `ConvergenceExecutor` object), and then maybe we could have an `after_success` method that, for launch_server, will update the cache and for heat will do nothing? (or maybe update the cache in a different way!)

dragorosson: Yes! I agree, and like I said in the original PR comment:
`The exact methods contained in ConvergenceExecutor may change. For example, just UpdateServersCache instead of the entirety of convergence_succeeded_servers since most of the code there isn't specific to launch_server.`

otter/convergence/service.py
```diff
@@ -799,3 +812,18 @@ def divergent_changed(self, children):
...
+
+class ConvergenceExecutor(object):
+    gather = attr.ib()
+    plan = attr.ib()
+    get_desired_group_state = attr.ib()
+    update_cache = attr.ib()
+
+
+@attr.s
+class LaunchServerExecutor(ConvergenceExecutor):
```
radix: So the way these objects are done strikes me as a biiiit gratuitous, but I'm not really that bothered by it.

However, the thing that does seem weird to me, is that if we are going to use an @attr.s object, that the specialized versions should really just be instances of that class instead of a subclass that re-declares all the attributes with defaults.

So I see two options that to me would be slightly preferable to what this currently does

1. just use plain classes, where ConvergenceExecutor is an ABC, that declares the methods with no implementation, and two subclasses (LaunchServerExecutor and HeatExecutor) that implement those methods differently
2. continue to use objects as bags of methods, but put the `@attr.s` decorator on `ConvergenceExecutor`, and declare the actual implementations as instances instead of subclasses, e.g.:

```
launch_server_executor = ConvergenceExecutor(
    gather=get_all_launch_server_data,
    plan=plan_launch_server,
    ...)
```

What do you think?

otter/test/convergence/test_service.py
```diff
@@ -848,12 +856,13 @@ def get_seq(self, with_cache=True):
         ]
 
     def _invoke(self, plan=None):
-        kwargs = {'plan_launch_server': plan} if plan is not None else {}
+        kwargs = {'plan': plan} if plan is not None else {}
+        executor = LaunchServerExecutor(gather=intent_func("gacd"), **kwargs)
```
radix: and if you go with option 2 above this would be something like `attr.assoc(launch_server_executor, gather=my_gather, ...)`

otter/convergence/service.py
```diff
@@ -340,11 +352,7 @@ def convergence_succeeded_servers(scaling_group, group_state, servers, now):
         yield cf_msg('group-status-active',
                      status=ScalingGroupStatus.ACTIVE.name)
     # update servers cache with latest servers
-    yield Effect(
-        UpdateServersCache(
-            scaling_group.tenant_id, scaling_group.uuid, now,
-            [thaw(s.json.set("_is_as_active", True))
-             for s in servers if s.state != ServerState.DELETED]))
+    yield executor.update_cache(scaling_group, servers, None, now)
```
manishtomar: The interface of `update_cache` changes with this. It will be nice to document that mentioning clearly the difference between passing `None` vs `[]` (empty list) in `lb_nodes`.

dragorosson:
@manishtomar, @radix: Ready for review!

otter/test/convergence/test_service.py
```diff
@@ -1417,3 +1427,28 @@ def test_multiple_lb_pending(self):
...
+        self.fake_attrs = {
+            'gather': 'g',
+            'plan': 'p',
+            'get_desired_group_state': 'gdgs',
+            'update_cache': 'uc',
+        }
+        self.lse = launch_server_executor
+
+    def test_launch_server_executor(self):
+        attrs = {
```
radix: Could you add docstrings to these test cases? We try to have docstrings for all tests

otter/convergence/service.py
```diff
@@ -802,3 +811,18 @@ def divergent_changed(self, children):
             # the return value is ignored, but we return this for testing
             eff = self._converge_all(my_buckets, children)
             return perform(self._dispatcher, self._with_conv_runid(eff))
+
+
+@attr.s
+class ConvergenceExecutor(object):
```
radix: A docstring would also be nice for this class

radix:
This is looking pretty good. Just needs some docstrings and I'm happy with merging it :)

otter/test/convergence/test_service.py
dragorosson: Done.

otter/convergence/service.py
dragorosson: Done.

PR #1754, opened 2015-10-03
Convergence throttling is optional

manishtomar:
It will be turned off when settings are not provided. This should allow high parallelism and reduce [integration tests times](https://as-jenkins.rax.io/job/_ci-integration-convergence-cafe/buildTimeTrend) from 15min to 5-6mins. Needs https://github.com/rackerlabs/autoscaling-chef/pull/781 merged before merging this.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -269,13 +269,12 @@ def _default_throttler(clock, stype, method):
     # no more than that are executed by a node, plus serialization of requests
     # will make it quite a bit lower than that.
 
-    cloud_client_config = config_value('cloud_client')
-    if cloud_client_config is None:
-        cloud_client_config = {}
-    throttling_config = cloud_client_config.get('throttling', {})
+    throttling_config = config_value('cloud_client.throttling')
+    if throttling_config is None:
+        return None
+
     create_server_delay = throttling_config.get('create_server_delay', 1)
     delete_server_delay = throttling_config.get('delete_server_delay', 0.4)
-
     policy = {
         (ServiceType.CLOUD_SERVERS, 'post'):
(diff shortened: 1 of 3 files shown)


otter/cloud_client.py
```diff
@@ -269,13 +269,12 @@ def _default_throttler(clock, stype, method):
...
 
-    cloud_client_config = config_value('cloud_client')
-    if cloud_client_config is None:
-        cloud_client_config = {}
-    throttling_config = cloud_client_config.get('throttling', {})
+    throttling_config = config_value('cloud_client.throttling')
+    if throttling_config is None:
+        return None
+
     create_server_delay = throttling_config.get('create_server_delay', 1)
```
radix: I would just get rid of these defaults here and rely on the config now. with your change I think the behavior is a bit confusing, in that if you put a `throttling: {}` in the config it will throttle with 1 and 0.4, I think it's better now to just not have any defaults and rely on the config to specify it.

radix:
LGTM, just one request. thanks!

otter/cloud_client.py
manishtomar: Sounds good! Will change it :)

radix:

radix:

radix:

manishtomar:

radix:
Definitely a big speedup on the builder. the cafe tests went from ~16min to ~7min! The worker tests are the longest at 9 minutes, but for some reason the whole PR builder takes ~15 minutes now (down from 22-25min). I wonder what that extra overhead is coming from, I guess it's the deployment of the code to preprod?

radix:
LGTM :)

PR #1755, opened 2015-10-06
removing last CLB node test

### LINKED ISSUE #1714: deleting node and server simultaneously
[CLB cafe test](https://github.com/rackerlabs/otter/blob/master/autoscale_cloudroast/test_repo/autoscale/system/integration/test_system_integration_lbaas.py#L244-L267) [failed](https://as-jenkins.rax.io/job/otter-cloudcafe-syd-convergence-system-tests/97/) because it expected nodes to get deleted _before_ server gets deleted as was the case with worker. However, convergence deletes server and corresponding nodes in same cycle. Mostly, nodes gets deleted first and the test passes but this time it deleted later than server and the test failed. 

Should the test expect the nodes to be gone after group state `activeCapacity` reflects that?

manishtomar:
This test was needed earlier when CLB had requirement of minimum 1 node. Now that requirement is not there and hence this test doesn't serve any purpose.

Fixes #1714.


radix:
LGTM :)

PR #1757, opened 2015-10-06
fix inter test dependency

manishtomar:
The test was expecting no nodes on clb3 which cannot be guaranteed since the CLB is being used by other tests and there is no per test setup and teardown. Now the test works with any existing nodes in it. This fixes strange failure occurred after merging #1755


manishtomar:

radix:
This LGTM, though it seems we should actually have a setUp/tearDown that cleans things up... Please feel free to merge after the build passes :)

radix:

PR #1759, opened 2015-10-07
worker throttling optional

manishtomar:
Needs `worker.create|delete_server_limit` for throttling to kick in. This allows higher number of tests to run concurrently increasing test run times. Requires https://github.com/rackerlabs/autoscaling-chef/pull/807 merged before this.

### DIFF
--- a/otter/test/worker/test_launch_server_v1.py
+++ b/otter/test/worker/test_launch_server_v1.py
@@ -907,6 +907,8 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
         self.treq = patch(self, 'otter.worker.launch_server_v1.treq')
         patch(self, 'otter.util.http.treq', new=self.treq)
 
+        launch_server_v1._create_server_sem = None
+
         self.generate_server_name = patch(
             self,
             'otter.worker.launch_server_v1.generate_server_name')
@@ -1143,11 +1145,6 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
         d = create_server('http://url/', 'my-auth-token', {'some': 'stuff'},
                           _treq=_treq, clock=self.clock)
 
-        # No result initially. Will be avail after 1 second due to
-        # injected delay
-        self.assertNoResult(d)
-        self.clock.advance(1)
-
         result = self.successResultOf(d)
         self.assertEqual(result, {"server": "created"})
         self.assertFalse(fs.called)
@@ -1155,7 +1152,7 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
     def test_create_server_limits(self):
         """
         create_server when called many times will post only 1 request at
-        a time
+        a time if configured with limit of 1
         """
         deferreds = [Deferred() for i in range(3)]
         post_ds = deferreds[:]
@@ -1166,6 +1163,7 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
             'imageRef': '1',
             'flavorRef': '3'
         }
+        set_config_data({"worker": {"create_server_limit": 1}})
 
         ret_ds = [create_server('http://url/', 'my-auth-token',
                                 server_config, clock=self.clock)
@@ -1194,11 +1192,6 @@ class ServerTests(RequestBagTestMixin, SynchronousTestCase):
         self.successResultOf(ret_ds[1])
         self.successResultOf(ret_ds[2])
 
-    def _create_server(self, url, token, conf, **kwargs):
-        d = create_server(url, token, conf, clock=self.clock, **kwargs)
-        self.clock.advance(1)
-        return d
-
     @mock.patch('otter.worker.launch_server_v1.find_server')
     def test_create_server_propagates_api_failure_from_create(self, fs):
         """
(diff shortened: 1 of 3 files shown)


radix:

radix:
LGTM :) running tests to see if we get a speedup

manishtomar:

manishtomar:

manishtomar:

manishtomar:

manishtomar:
Closing as I am don't feel like debugging why the one test is failing in such a strange way. Might revisit it later.

manishtomar:
Reopening to try again.

manishtomar:

manishtomar:

manishtomar:

manishtomar:
This fails because delete server also needs throttling.

manishtomar:

manishtomar:

manishtomar:

manishtomar:

manishtomar:
@glyph could you review this once you find some time?

otter/test/worker/test_launch_server_v1.py
```diff
@@ -912,9 +912,14 @@ def setUp(self):
             'otter.worker.launch_server_v1.generate_server_name')
         self.generate_server_name.return_value = 'as000000'
 
+        # reset it to empty in beginning of test as it is module-wise
+        # info retained in every test
+        launch_server_v1._semaphores = {}
```
glyph: This looks dodgy to me. Consider using [`patch`](https://twistedmatrix.com/documents/15.5.0/api/twisted.trial.unittest.SynchronousTestCase.html#patch)?

otter/test/worker/test_launch_server_v1.py
```diff
@@ -2224,6 +2220,10 @@ def setUp(self):
             self, 'otter.worker.launch_server_v1.remove_from_load_balancer')
         self.remove_from_load_balancer.return_value = succeed(None)
 
+        # reset it to empty in beginning of test as it is module-wise
+        # info retained in every test
+        launch_server_v1._semaphores = {}
```
glyph: Not only should this use `patch` (as above) but I feel like it should be in a method or something; repeating both the same code and the same comment is a code smell.

otter/test/worker/test_launch_server_v1.py
```diff
@@ -2423,6 +2421,9 @@ def test_delete_and_verify_limits(self):
         :func:`delete_and_verify` limits the number of delete server requests
         to 1. It also delays response by 1 second
         """
+        set_config_data({"worker": {"delete_server_limit": 1}})
+        self.addCleanup(set_config_data, None)
```
glyph: Should all calls to `set_config_data` work this way?  Maybe there should be a `set_config_data_for_test` function?

otter/worker/launch_server_v1.py
```diff
@@ -152,9 +152,23 @@ def check_status(server):
...
 
 
-# limit on 1 servers to be created simultaneously
-MAX_CREATE_SERVER = 1
-create_server_sem = DeferredSemaphore(MAX_CREATE_SERVER)
+# single global instance of semaphores
+_semaphores = {}
+
+
+def get_sempahore(operation, conf_name):
```
glyph: Not sure about Otter, but on Twisted I would definitely say I'd expect to see a `:param:` and perhaps a `:type:` for each parameter, and a `:return:` explaining the return type...

otter/worker/launch_server_v1.py
```diff
@@ -325,7 +340,11 @@ def _create_server():
 
         If not, and if no further errors occur, server creation can be retried.
         """
-        d = create_server_sem.run(_create_with_delay)
+        sem = get_sempahore("create_server", "worker.create_server_limit")
```
glyph: My preference would be to have a semaphore object in a namespace somewhere rather than using strings here.  The general idea is that I always like to find out about missing objects via a `NameError` or `AttributeError`, not a `KeyError` upon looking something up in a dictionary; reason being, if you have something that wishes to share the `"create_server"` semaphore, it shouldn't be able to accidentally and silently get a different semaphore if you have a typo.  But you may feel free to ignore this.

glyph:
Address to your taste and feel free to land thereafter; these are mostly aesthetic criticisms, and I'm not sure how strictly Otter likes to enforce documentation standards...

otter/test/worker/test_launch_server_v1.py
manishtomar: This is there in hundreds of places and I feel like hitting myself for not already doing it :angry: . Thanks for suggesting it :)

manishtomar: I just remembered. I actually started on a generic function that also preserves existing config 1-2 months but hit some roadblock. Let me try to revisit it and see.

otter/worker/launch_server_v1.py
manishtomar: I think a better a option is avoid strings altogether and use some kind of enums. However, this is old code supposed to go away. So I am feeling lazy to make that change :).

manishtomar:

manishtomar:
@glyph Addressed your comments. Thanks a lot for your review :) Merging.

PR #1760, opened 2015-10-08
rename `s` to `parent` in api.py

radix:
just had this lying around in my WC and didn't have a branch to attach it to, so I decided to make a PR just for this.

there was a local variable called `s` that was used across way too many lines. renamed it to `parent` (since it's a parent service).


manishtomar:
LGTM

PR #1761, opened 2015-10-08
allow specifying "all" as the value for "convergence-tenants"

radix:
super simple PR. When `convergence-tenants` is set to the string `"all"` (as opposed to a list of specific tenant IDs), then all tenants have convergence enabled.


manishtomar:
LGTM

PR #1762, opened 2015-10-08
less simple CLB throttling

radix:
This just uses the existing mechanism for throttling in cloud_client to throttle CLB usage.

### DIFF
--- a/otter/cloud_client.py
+++ b/otter/cloud_client.py
@@ -22,7 +22,6 @@ from toolz.dicttoolz import get_in
 from toolz.functoolz import identity
 from toolz.itertoolz import concat
 
-from twisted.internet.defer import DeferredLock
 from twisted.internet.task import deferLater
 
 from txeffect import deferred_performer, perform as twisted_perform
@@ -43,6 +42,7 @@ from otter.util.pure_http import (
     has_code,
     request,
 )
+from otter.util.weaklocks import WeakLocks
 
 
 def add_bind_service(catalog, service_name, region, log, request_func):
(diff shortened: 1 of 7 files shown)


radix:

otter/cloud_client.py
```diff
@@ -264,19 +264,20 @@ def _default_throttler(clock, stype, method):
...
-            clock, conf)
-
-    return policy.get((stype, method))
+    cfg_names = {
+        (ServiceType.CLOUD_SERVERS, 'post'): 'create_server_delay',
+        (ServiceType.CLOUD_SERVERS, 'delete'): 'delete_server_delay',
+        (ServiceType.CLOUD_LOAD_BALANCERS, 'get'): 'get_clb_delay',
+        (ServiceType.CLOUD_LOAD_BALANCERS, 'post'): 'post_clb_delay',
+        (ServiceType.CLOUD_LOAD_BALANCERS, 'put'): 'put_clb_delay',
+        (ServiceType.CLOUD_LOAD_BALANCERS, 'delete'): 'delete_clb_delay',
```
manishtomar: Hmm. I was hoping we can configure it on [per-tenant basis](http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/Rate_Limits-d1e821.html) since CLB doesn't really have scale issues. This may increase the call times and in effect convergence cycle times if many groups are converging. I am not sure if that is an issue. This is definitely a good start for thorttling :)

radix:
@manishtomar Oh, you make an excellent point. I'll see what I can do.

radix:
also, it turns out that the throttling code is totally broken and doesn't actually throttle calls -- the only thing that's actually working is the delay part, not the serialization part, because the lock objects we create aren't being shared between different calls to the throttler. This is entirely my fault since I wrote that buggy code in the first place :P

PR #1763, opened 2015-10-08
Trigger convergence script

manishtomar:
Script to trigger convergence on all groups or specific group(s). Tested this with single group.


radix:
LGTM :)

PR #1765, opened 2015-10-09
Fix throttling

radix:
Depends on #1764.

Previously, the lock object used to implement throttling in cloud_client was not shared between calls, so in essence there was absolutely no useful locking going on.

This change uses a `WeakLocks` object to keep track of the locks so that they're shared between calls. (this is a bit overkill for the current situation, since we have a fixed number of locks so we could just put them in a global data structure, but with upcoming changes to make locking be per-tenant, WeakLocks makes more sense).

### DIFF
--- a/otter/test/test_cloud_client.py
+++ b/otter/test/test_cloud_client.py
@@ -19,7 +19,6 @@ import six
 
 from toolz.dicttoolz import assoc
 
-from twisted.internet.defer import succeed
 from twisted.internet.task import Clock
 from twisted.trial.unittest import SynchronousTestCase
 
@@ -46,7 +45,6 @@ from otter.cloud_client import (
     _Throttle,
     _default_throttler,
     _perform_throttle,
-    _serialize_and_delay,
     add_bind_service,
     add_clb_nodes,
     change_clb_node,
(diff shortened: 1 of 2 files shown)


radix:

radix:
I think the integration tests might be failing because the now-working throttling is making the tests take too long, but I'm not certain of that

manishtomar:
There was problem in throttling config. I've reduced concurrency to earlier setting and tests should work now.

otter/test/test_cloud_client.py
```diff
@@ -414,36 +389,50 @@ def test_post_and_delete_not_the_same(self):
...
+        locks = WeakLocks()
         set_config_data(
-            {'cloud_client': {'throttling': {'create_server_delay': 500}}})
+            {'cloud_client': {'throttling': {cfg_name: 500}}})
         self.addCleanup(set_config_data, {})
         clock = Clock()
-        bracket = _default_throttler(clock, ServiceType.CLOUD_SERVERS, 'post')
+        bracket = _default_throttler(locks, clock, stype, method)
+        if bracket is None:
+            self.fail("No throttler for %s and %s" % (stype, method))
```
manishtomar: Nitpick. You could instead do `self.assertIsNotNone(bracket, "No throttler for {} and {}".format(stype, method))`

manishtomar:
LGTM modulo nitpick :). Can merge once build passes

manishtomar:

manishtomar:
Sorry this failure was because of me increasing concurrency in worker _without_ merging #1759. Let me run again with low concurrency.

manishtomar:

PR #1766, opened 2015-10-09
Implement per-tenant cloud_client throttling and use it for CLB requests

radix:
Depends on #1764 and #1765.

This adds support for cloud_client to throttle service requests on a per-tenant basis.

A number of CLB-related configuration options are added to allow specifying a request rate (within the scope of a tenant).

### DIFF
--- a/otter/test/test_cloud_client.py
+++ b/otter/test/test_cloud_client.py
@@ -393,34 +393,26 @@ class SerializeAndDelayTests(SynchronousTestCase):
 class DefaultThrottlerTests(SynchronousTestCase):
     """Tests for :func:`_default_throttler`."""
 
+    def tearDown(self):
+        set_config_data(None)
+
     def test_mismatch(self):
         """policy doesn't have a throttler for random junk."""
         bracket = _default_throttler(None, 'foo', 'get')
         self.assertIs(bracket, None)
 
-    def test_post_cloud_servers(self):
-        """POSTs to cloud servers get throttled by a second."""
-        clock = Clock()
-        bracket = _default_throttler(clock, ServiceType.CLOUD_SERVERS, 'post')
-        d = bracket(lambda: 'foo')
-        self.assertNoResult(d)
-        clock.advance(1)
-        self.assertEqual(self.successResultOf(d), 'foo')
-
-    def test_delete_cloud_servers(self):
-        """DELETEs to cloud servers get throttled by a second."""
-        clock = Clock()
-        bracket = _default_throttler(clock,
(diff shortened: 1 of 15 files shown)


radix:

radix:

manishtomar:

manishtomar:
I started again after reducing concurrency. Really sorry for the confusion @radix :(

otter/test/test_cloud_client.py
```diff
@@ -427,12 +429,56 @@ def _test_throttle(self, cfg_name, stype, method):
...
+        per-tenant.
+        """
+        locks = WeakLocks()
+        set_config_data(
+            {'cloud_client': {'throttling': {cfg_name: 500}}})
+        self.addCleanup(set_config_data, {})
+        clock = Clock()
+        bracket1 = _default_throttler(locks, clock, stype, method, 'tenant1')
+        if bracket1 is None:
+            self.fail("No throttler for %s and %s" % (stype, method))
```
manishtomar: Nitpick: Can use `self.assertIsNotNone` :)

otter/test/test_cloud_client.py
```diff
@@ -427,12 +429,56 @@ def _test_throttle(self, cfg_name, stype, method):
...
+        result1 = bracket1(lambda: 'bar1')
+        bracket2 = _default_throttler(locks, clock, stype, method, 'tenant2')
+        result2 = bracket2(lambda: 'bar2')
+        self.assertNoResult(result1)
+        self.assertNoResult(result2)
+        clock.advance(500)
+        self.assertEqual(self.successResultOf(result1), 'bar1')
+        self.assertEqual(self.successResultOf(result2), 'bar2')
```
manishtomar: Sorry I don't see how this tests that locks are not shared per-tenant. Maybe you can make 2 calls on tenant1 one after other and single call on tenant2 and observe that they don't depend on each other?

manishtomar:
LGTM modulo comments.

otter/test/test_cloud_client.py
radix: It creates two throttlers for two different tenants, and then *only* advances time by 500 seconds, and ensures that both requests complete. Since the delay is set to 500, this proves that the delays are independent. If the lock was shared, then after 500 seconds only one of the requests would have gone through, and it would require another 500 seconds for the second one to complete.

But I could be confused about your point, sorry if I'm missing something

manishtomar: Sorry you are right, please ignore my comment above :)

PR #1767, opened 2015-10-15
Consolidate cache updating into one method

dragorosson:


otter/test/convergence/test_service.py
```diff
@@ -933,9 +941,6 @@ def plan(dgs, servers, lb_nodes, now, build_timeout):
              noop)
         ]
 
-        # all the servers updated in cache in beginning
-        self.cache.append(thaw(deleted.json))
```
dragorosson: Must make sure removing this line is okay because it represents a change in behavior for updating the cache.

stretchservers:

stretchservers:

otter/convergence/service.py
```diff
@@ -172,7 +172,8 @@ def update_cache(group, servers, lb_nodes, now):
         sd = thaw(server.json)
         if is_autoscale_active(server, lb_nodes):
             sd["_is_as_active"] = True
-        server_dicts.append(sd)
+        if server.state != ServerState.DELETED:
```
manishtomar: I thought this wasn't required since `is_autoscale_active` is already checking for ACTIVE servers

dragorosson: Not quite. The logic in each method (before) did two things: One doesn't add the `_is_as_active` _key_ if that check doesn't pass. The other _doesn't add the server to the cache_ if it's deleted. That's not the same thing and that's why both are represented here. We want (I'm assuming) that deleted servers shouldn't be in the cache at all, not just have that key omitted.

dragorosson:
Updated.

otter/convergence/service.py
```diff
@@ -162,7 +162,7 @@ def all_met(server, current_lb_nodes):
                              if node.matches(server)]))
 
 
-def update_cache(group, servers, lb_nodes, now):
+def update_cache(group, servers, lb_nodes, now, include_deleted=True):
```
manishtomar: This function as a separate unit probably does not know the right default for `include_deleted`. I think it would be better to take it as a positional argument. Moreover, calling it explicitly with `True` or `False` when gathering and succeeding makes it clear when deleted servers are cached.

otter/convergence/service.py
```diff
@@ -172,7 +172,8 @@ def update_cache(group, servers, lb_nodes, now):
         sd = thaw(server.json)
         if is_autoscale_active(server, lb_nodes):
             sd["_is_as_active"] = True
-        server_dicts.append(sd)
+        if server.state != ServerState.DELETED or include_deleted:
```
manishtomar: Non-blocking: To me `if include_deleted and server.state == ServerState.DELETED` seems more readable :)

otter/test/convergence/test_service.py
```diff
@@ -817,7 +823,8 @@ def setUp(self):
                    links=freeze([{'href': 'link2', 'rel': 'self'}]))
         )
         self.state_active = {}
-        self.cache = [thaw(self.servers[0].json), thaw(self.servers[1].json)]
+        self.cache = [thaw(self.servers[0].json.set('_is_as_active', True)),
+                      thaw(self.servers[1].json.set('_is_as_active', True))]
```
manishtomar: I may be failing to understand but based on above `update_cache` changes I don't think servers need to become AS active.

otter/convergence/service.py
manishtomar: Also docstrings will really help :)

dragorosson: @manishtomar: The idea was that logically, whatever servers you give `update_cache` should be cached and if necessary, the deleted ones can be excluded. Therefore, I thought it made sense to give it a default. I am fine making it a positional argument, but what do you think about that?

And I will add a docstring.

otter/convergence/service.py
dragorosson: @manishtomar: The logic would have to be `if not (include_deleted and server.state == ServerState.DELETED)`, which seems more complicated. If you're just saying it'd be more readable if the conditions should be swapped, I can do that.

otter/test/convergence/test_service.py
dragorosson: Well, this comes down to how the tests should be updated. I had a choice between
- Changing `self.cache` and the lb_nodes returned by the fake gathering globally, and then removing the lb_nodes returned in certain test cases.
- Changing lb_nodes returned by gathering in the tests cases where convergence succeeds, which I think would require changing `self.cache` in those test cases.
- Some other combination of fiddling with tests.

I went with the first option. We can discuss this because I'm not going to pretend I know the best way to update the tests for this change.

dragorosson: And going with the first option requires `self.cache` to be changed because now the servers are actually on all of the load balancers they want to be, so `_is_as_active` should be set when they get cached.

otter/convergence/service.py
manishtomar: Hm.. I didn't from that angle. That sounds good. Its fine either way :) I'll leave it upto you.

otter/convergence/service.py
manishtomar: Duh. I didn't realize it will not add non-deleted servers. This is fine. If I was to write I would probably do
```python
if not include_deleted and server.state == ServerState.DELTED:
   continue
```
in beginning of the loop which probably increases the code size. But all this is bike-shedding :). This is fine as it is.

dragorosson: Ha, yes, I almost wrote that but then stopped myself :)

otter/test/convergence/test_service.py
manishtomar: Oh yes. You are right. Thanks for correcting! This is good :)

otter/convergence/service.py
dragorosson: I left it as is. Added docstring.

manishtomar:

manishtomar:
LGTM once build passes :)

PR #1770, opened 2015-10-23
rename cloudcafe package to avoid conflict

glyph:
opencafe uses `load_module` to inject namespaces, which means that the `autoscale` utility cloudcafe module conflicts with the `autoscale` sub-package `test_repo`.


glyph:
I can't even get the integration tests to start without this change. I realize that this is crazy, because everybody else is somehow running them, but from my perspective I can't understand how they ever possibly could have worked before.

stretchservers:

stretchservers:

radix:
@glyph - actually, this has always affected me, so I'm glad you figured out how to fix this :) It only happened to me under certain circumstances though -- like trying to run specific modules (though actually something may have changed recently to make it happen all the time...)

radix:

glyph:
Thanks a bunch @radix !

PR #1773, opened 2015-11-05
Add steps for stack operations

dragorosson:


stretchservers:

stretchservers:

otter/convergence/steps.py
```diff
@@ -539,3 +552,91 @@ def as_effect(self):
...
+class CreateStack(object):
+    """
+    A stack must be created.
+
+    :ivar pmap stack_config: Heat launch configuration.
+
+    """
+    stack_config = attr.ib(validator=instance_of(PMap))
+
+    def as_effect(self, append_stack_uuid=append_stack_uuid):
```
radix: given how simple of a transform this `append_stack_uuid` operation is, I personally wouldn't bother to parameterize it and override it in the tests -- it just saves you a tiny bit of data declaration in the test to create a config that has the necessary shape to be modified by this function.

radix: oh, hold on, I didn't notice the uuid4 before. lemme think about this a bit more.

radix: Okay, yeah, so this uuid4 call makes the function impure, which we don't want to do when it's a function that returns an Effect. (If it returns an Effect, the way it constructs that effect should be pure. The effect itself should represent all the side-effects to be performed). See `CreateServer.as_effect` for a very similar case.

So here's what I suggest:

    eff = Effect(Func(uuid4))
    def got_uid(uid):
        stack_config = append_stack_uuid(self.stack_config, uid)
        return create_stack(thaw(stack_config)).on(
            _success_reporter('Waiting for stack to create'))
    return eff.on(got_uid)

Or, if you want to avoid the explicit callback, you can use @do notation.

dragorosson: Hooray, good catch, @radix! I've updated this PR with that change.

otter/convergence/steps.py
```diff
@@ -73,6 +80,12 @@ def set_server_name(server_config_args, name_suffix):
     return set_in(server_config_args, ('server', 'name'), name)
 
 
+def append_stack_uuid(stack_config, uuid):
```
radix: If you want to just inline the body of this function into its only call-site I think that'd be fine, since it's so simple. But no big deal either way.

otter/test/convergence/test_steps.py
```diff
@@ -1177,3 +1188,125 @@ def test_returns_retry(self):
         self.assertEqual(
             sync_perform(base_dispatcher, eff),
             (StepResult.RETRY, ['building']))
+
+
+class AppendStackUUIDTests(SynchronousTestCase):
+    """Tests for :func:`append_stack_uuid`."""
+
+    def test_normal_use(self):
+        """Ensures correct behavior."""
```
radix: This docstring is a bit pointless ;) How about "The stack ID is inserted into the given stack configuration" or something.

otter/convergence/steps.py
dragorosson: I'll leave it because I like the small bit of clarity it adds and that it matches create server's `set_server_name`.

Also, added docstring.

otter/test/convergence/test_steps.py
dragorosson: Done.

radix:
Thank you :)

PR #1769, opened 2015-10-16
Return resources in dict instead of tuple

dragorosson:

PR #1774, opened 2015-11-05
Add launch_stack gathering methods

dragorosson:


stretchservers:

stretchservers:

otter/convergence/gathering.py
```diff
@@ -138,6 +141,22 @@ def get_scaling_group_servers(tenant_id, group_id, now,
...
+        query['tags'] = stack_tag
+
+    return list_stacks_all(query)
+
+
+@do
+def get_scaling_group_stacks(tenant_id, group_id, now,
+                             get_all_stacks=get_all_stacks):
+    stacks = yield get_all_stacks(stack_tag=get_stack_tag_for_group(group_id))
+    yield do_return(stacks)
```
manishtomar: Nitpick: Since this basically delegates call to another function, you don't need `do` generator here. This can just be `return get_all_stacks(stack_tag=get_stack_tag_for_group(group_id))`.

dragorosson: Nice! Done.

dragorosson:
Rebased.

otter/convergence/gathering.py
```diff
@@ -255,3 +272,20 @@ def get_all_launch_server_data(
...
+        get_scaling_group_stacks=get_scaling_group_stacks):
+    """
+    Gather all launch_stack data relevant for convergence w.r.t given time
+
+    Returns an Effect of ([HeatStack], None) to match
+    get_all_launch_server_data.
+    """
+    eff = (get_scaling_group_stacks(tenant_id, group_id, now)
+           .on(map(HeatStack.from_stack_details_json)).on(list)
+           .on(lambda stacks: {'stacks': stacks}))
```
manishtomar: If this returning stacks in a structured way (`HeatStack` objects) then why make it a dict?

dragorosson: I've updated the docstring for this method.

The reason is that it the gathering methods now return a dict of gathered resource types. `get_all_launch_server_data` returns `{'servers': [NovaServer], 'lb_nodes': [LBNode]}` and this method returns `{'stacks': [HeatStack]}`. This is new as of https://github.com/rackerlabs/otter/pull/1769

manishtomar: Oh ok cool. Sorry didn't realize that :)

otter/convergence/gathering.py
```diff
@@ -138,6 +141,20 @@ def get_scaling_group_servers(tenant_id, group_id, now,
...
+def get_all_stacks(stack_tag=None, batch_size=100):
+    query = {}
+
+    if stack_tag is not None:
+        query['tags'] = stack_tag
+
+    return list_stacks_all(query)
+
+
+def get_scaling_group_stacks(tenant_id, group_id, now,
```
radix: Is there a reason this takes `tenant_id` and `now` without using them?

otter/convergence/gathering.py
```diff
@@ -138,6 +141,20 @@ def get_scaling_group_servers(tenant_id, group_id, now,
     yield do_return(servers)
 
 
+def get_all_stacks(stack_tag=None, batch_size=100):
```
radix: likewise `batch_size` is an unused argument. I'd rather not give the impression that batching is supported when it's not

otter/test/convergence/test_gathering.py
```diff
@@ -702,3 +706,89 @@ def test_no_group_servers(self):
...
 
         self.assertEqual(resolve_stubs(eff), {'servers': [], 'lb_nodes': []})
+
+
+class GetAllStacksTests(SynchronousTestCase):
+    """Tests for :func:`get_all_stacks`."""
+
+    def test_default(self):
+        """
+        Tests for passing no arguments.
```
radix: In general we try to describe what should happen in a test docstring instead of what feature is being tested. Here's a nice short guideline on the subject: https://jml.io/pages/test-docstrings.html

OTOH, don't be afraid to keep them short. It's also okay to keep the triple-quotes on the same line as the content as long as it's short enough to fit on one line.

Here I would say something like "Passing no arguments causes all stacks to be requested."

radix:
Just a few minor nitpicks. LGTM after those are fixed :)

otter/convergence/gathering.py
dragorosson: An artifact from matching the interface of `get_all_launch_server_data` and then passing arguments through. Removed now.

Also, `get_scaling_group_servers` doesn't use `now` either. I would think that the linter would catch unused method arguments...

otter/convergence/gathering.py
dragorosson: Whoops. Done.

otter/test/convergence/test_gathering.py
dragorosson: Done, and thanks for the advice!

radix:
Thanks :)

PR #1775, opened 2015-11-05
Add launch_stack planning methods

dragorosson:


stretchservers:

stretchservers:

otter/convergence/planning.py
```diff
@@ -367,6 +373,103 @@ def drain_and_delete_a_server(server):
...
+    :param DesiredStackGroupState desired_state: The desired group state.
+    :param set stacks: a set of :obj:`HeatStack` instances.
+        This must only contain stacks that are being managed for the specified
+        group.
+    :rtype: :obj:`pbag` of `IStep`
+
+    """
+    config = desired_state.stack_config
+
+    group_states = [StackState.CREATE_UPDATE_COMPLETE,
```
radix: I think this code could be a bit more clear by just doing something like

```python
stacks_by_state = groupby(lambda stack: stack.get_state(), stacks)
stacks_complete = stacks_by_state.get(StackState.CREATE_UPDATE_COMPLETE, [])
stacks_failed = stacks_by_state.get(StackState.CREATE_UPDATE_FAILED, [])
```

This would be significantly shorter (line-wise), and I think more obvious to casual readers.

Anyway, just a nitpick, feel free to ignore :)

otter/convergence/planning.py
```diff
@@ -367,6 +373,103 @@ def drain_and_delete_a_server(server):
...
+
+    stacks_amiss = (stacks_failed +
+                    stacks_check_failed +
+                    stacks_in_progress +
+                    stacks_delete_in_progress)
+
+    if stacks_delete_failed:
+        return pbag([FailConvergence(["Stacks in DELETE_FAILED found."])])
+
+    # If there are no stacks in CHECK_* or other work to be done, we assume
```
radix: I think it'd be nice to have a more thorough/holistic description of the workflow we put stacks through somewhere, maybe at the top of this function, or the top of this module with a `Note` (like those in service.py). The comments you have here are good but I'd like something with a bit more context. Basically, something that distills the conversation we had way back when into some documentation. ASCII art diagrams may be useful ;)

radix:
Can we get some tests for various transitions of 0 and 1 stacks in a group? like, say:

- current: 0, desired: 1
- current: 1, desired: 0
- current: 0, desired: 0
- current: 1, desired: 1

otter/convergence/planning.py
dragorosson: I like it. Done.

otter/convergence/planning.py
dragorosson: Done.

dragorosson:
Added those tests.

radix:
Thank you @dragorosson :)

PR #1777, opened 2015-11-11
opt out convergence by default

manishtomar:
All tenants will be convergence enabled unless opted out via `"non-convergence-tenants"` config. "convergence-tenants" key is not used anymore. One has to put `none` in above key to enable convergence in all tenants instead of `all` earlier.

NOTE: This will be merged and deployed after getting a go-ahead from product. Requires https://github.com/rackerlabs/autoscaling-chef/pull/806 to be merged before this.


otter/convergence/composition.py
```diff
@@ -22,12 +22,12 @@ def tenant_is_enabled(tenant_id, get_config_value):
         "convergence-tenants" portion of the configuration file.
     :param callable get_config_value: config key -> config value.
     """
-    enabled_tenant_ids = get_config_value("convergence-tenants")
```
radix: I recommend making this an additive change so we can switch from a whitelist to a blacklist with only a config change.

Here's a suggestion for how: `convergence-enabled` could be a new config key, the value would either be `"blacklist"` or `"whitelist"`, and then there would also be `convergence-tenants` and `non-convergence-tenants` keys.

```python
if convergence_enabled == 'blacklist' and tenant_id not in non_convergence_tenants:
    return True
elif convergence_enabled == 'whitelist' and tenant_id in convergence_tenants:
    return True
else:
    return False
```

manishtomar: We can always have additive deploys by first deploying "non-convergence-tenants" config change followed by this PR change. Would those steps be comfortable?

radix: yeah, it makes sense to put off merging this until we want to make the change.

radix:
This is looking good, thanks for the change manish :)

manishtomar:

manishtomar:

manishtomar:

PR #1779, opened 2015-11-12
update_config_data function

manishtomar:
This updates any existing config and does not delete other configurations. Will be used in #1777


radix:
Maybe another test that ensures it creates a new dictionary if a parent key is also non-existent? e.g. `update_config_in("baz.newpath.newkey", "value")`?

otherwise LGTM :)

manishtomar:
@radix Sorry merging :) Please let me know if https://github.com/rackerlabs/otter/commit/baf19e7b0a4a4191c5b6b464eed3579f5ed3bf0b change was not enough. Thanks for the review!

PR #1781, opened 2015-11-16
Add migration for setting desired for groups based on CSV file

radix:
I added a migration to `load_cql.py` that reads a CSV file of `tenant_id,group_id,desired` and updates `desired` in the database.

I had to refactor the file a tiny little bit to give migrations access to arguments so it could load the CSV file based on a command line argument.


manishtomar:
This LGTM :) Let me just run this on sample data and see.

PR #1784, opened 2015-11-18
remove getting todays groups

manishtomar:
and all its related functions. It was done primarily to slowly increase load on Nova and monitor its performance. Its been long since all tenants are fetched and Nova has not had any issues.


radix:
LGTM :)

PR #1787, opened 2015-11-19
Add launch stack composition logic

dragorosson:


stretchservers:

stretchservers:

otter/convergence/composition.py
```diff
@@ -86,3 +106,20 @@ def prepare_server_launch_config(group_id, server_config, lb_descriptions):
...
+
+    :param str group_id: The group ID
+    :param PMap stack_config: The stack part of the Group's launch config,
+        as per :obj:`otter.json_schema.group_schemas.stack`.
+    """
+    # Set stack name and tag to the same thing
+    stack_config = set_in(stack_config, ('stack_name',),
+                          get_stack_tag_for_group(group_id))
+    stack_config = set_in(stack_config, ('tags',),
+                          get_stack_tag_for_group(group_id))
```
radix: we should probably append our tag to any existing 'tags' value if there is already one set in the launch config

dragorosson: Right now, there can't be because `stack_name` and `tags` are disallowed in the schema (see the JSON schema PR).

dragorosson: Added https://github.com/rackerlabs/otter/issues/1792 to track fixing this.

radix:
+1 :)

PR #1788, opened 2015-11-19
Add launch_stack JSON Schema

dragorosson:


stretchservers:

stretchservers:

radix:
As I mentioned on Slack, I think it would be a good idea to allow tags, but this LGTM.

PR #1791, opened 2015-11-19
Allow bfv server config with block_device_mapping_v2 setting

### LINKED ISSUE #1780: Launch config with block_device_mapping_v2
BFV servers can also be created with `block_device_mapping_v2` field in `POST /servers` call apart from `block_device_mapping` field. Please allow `block_device_mapping_v2` field in a group's launch config.

manishtomar:
All the [Nova docs](https://developer.rackspace.com/docs/cloud-servers/v2/developer-guide/#post-create-server-servers) point to _v2 setting. `block_device_mapping` is legacy. 

Closes #1780.

### DIFF
--- a/otter/json_schema/group_schemas.py
+++ b/otter/json_schema/group_schemas.py
@@ -87,6 +87,10 @@ _bfv_server = {
     }
 }
 
+_bfv2_server = deepcopy(_bfv_server)
+bdm = _bfv2_server["properties"].pop("block_device_mapping")
+_bfv2_server["properties"]["block_device_mapping_v2"] = bdm
+
 _non_bfv_server = {
     "type": "object",
     "properties": {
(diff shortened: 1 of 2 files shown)


otter/json_schema/group_schemas.py
```diff
@@ -87,6 +87,10 @@
     }
 }
 
+_bfv2_server = deepcopy(_bfv_server)
+bdm = _bfv2_server["properties"].pop("block_device_mapping")
```
radix: it seems to follow the convention this variable should also be prefixed with an `_`?

radix:
LGTM. Feel free to address my nitpick or not an then merge :)

manishtomar:
Thanks for the review @radix. Merging after making that change :).

PR #1793, opened 2015-11-24
Include deleted group's servers when collecting metrics

manishtomar:
Also, consider all non-"bad" servers under pending. "bad" servers are one's with status of SHUTOFF, ERROR  or DELETED.

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -60,7 +60,7 @@ def get_tenant_metrics(tenant_id, scaling_groups, grouped_servers,
     """
     if _print:
         print('processing tenant {} with groups {} and servers {}'.format(
-              tenant_id, len(scaling_groups), len(servers)))
+              tenant_id, len(scaling_groups), len(grouped_servers)))
 
     groups = {g['groupId']: g for g in scaling_groups}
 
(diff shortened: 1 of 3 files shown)


otter/metrics.py
```diff
@@ -57,19 +60,35 @@ def get_tenant_metrics(tenant_id, scaling_groups, servers, _print=False):
...
-    for group in scaling_groups:
-        group_id = group['groupId']
-        create_metrics = partial(GroupMetrics, tenant_id,
-                                 group_id, group['desired'])
-        if group_id not in servers:
-            metrics.append(create_metrics(0, 0))
+
+    groups = {g['groupId']: g for g in scaling_groups}
+
+    def get_metric(args):
```
radix: this code got pretty complicated and I think it's because it's using merge_with when that's not really a good solution (I think it's only really useful when the data is relatively homogenous -- i.e. it doesn't matter which dict it comes from).

I think a simpler solution would be to iterate over `set(groups.keys() + servers.keys())`.

manishtomar:
@radix had +1ed last week verbally before he left :cry:

PR #1794, opened 2015-11-24
Small fixes for launch_stack

dragorosson:


stretchservers:

stretchservers:

otter/test/test_cloud_client.py
```diff
@@ -1483,7 +1487,23 @@ def test_create_stack(self):
         result = perform_sequence(seq, eff)
         self.assertEqual(result, 'the_stack_json')
 
-    def test_check_stack(self):
+    def test_check_stack_200(self):
```
radix: so heat sometimes return _either_ 200 or 201?

radix:
LGTM, just have a question about 200/201

otter/test/test_cloud_client.py
dragorosson: Yes, it's kinda weird. Similarly, stack preview can error with 400, 404, or 409 depending on the problem. Stack preview is that way because there's a module that transforms a Heat exception into an HTTP exception (i.e. something like `ResourceNotFound` -> `HTTPNotFound`). I figure the 2xx responses for stack check are the same way because one has no response body, while the other tells you that one of the resource doesn't have stack check implemented (kind of like an FYI). Not sure why it's 201 though, seems like a weird one to pick for what it does.

radix: Cool API. Thanks for the explanation :)

PR #1795, opened 2015-12-01
trigger_convergence.py has multiple options to choose the groups to trigger

manishtomar:
Tested this on dev vm with all options


radix:
Looking good! :)

PR #1797, opened 2015-12-02
Add launch_stack integration test

dragorosson:
Can't merge this until the stack check endpoint is enabled in Heat.

### DIFF
--- a/otter/integration/tests/test_launch_stack.py
+++ b/otter/integration/tests/test_launch_stack.py
@@ -1,0 +1,143 @@
+"""
+Tests covering Otter's integration with Heat for the launch_stack launch
+configuration.
+"""
+
+from pyrsistent import pbag
+
+from twisted.internet import reactor
+from twisted.internet.defer import inlineCallbacks
+from twisted.python.log import msg
+from twisted.trial import unittest
+from twisted.web.client import HTTPConnectionPool
+
+from otter.convergence.model import get_stack_tag_for_group
+
+from otter.integration.lib.autoscale import ScalingGroup, ScalingPolicy
+from otter.integration.lib.resources import TestResources
+from otter.integration.lib.trial_tools import (
+    TestHelper,
+    get_identity,
+    get_resource_mapping,
+    region)
+
+from otter.log import log
+
+from otter.util.deferredutils import retry_and_timeout
+from otter.util.http import check_success, headers
+from otter.util.logging_treq import LoggingTreq
+from otter.util.retry import (
+    TransientRetryError,
+    repeating_interval,
+    terminal_errors_except
+)
+
+
+class TestLaunchStack(unittest.TestCase):
+    """Tests making sure Otter can talk to Heat properly."""
+
+    def setUp(self):
+        """
+        Establish an HTTP connection pool and commonly used resources for each
+        test. The HTTP connection pool is important for maintaining a clean
+        Twisted reactor.
+        """
+        self.helper = TestHelper(self)
+        self.rcs = TestResources()
+        self.identity = get_identity(self.helper.pool)
+        self.pool = HTTPConnectionPool(reactor, False)
+        self.treq = LoggingTreq(log=log, log_response=True)
+
+        scaling_group_config = {
+            'launchConfiguration': {
+                'args': {
+                    'stack': {
+                        'template': {
+                            'heat_template_version': '2015-04-30',
+                            'resources': {
+                                'rand': {'type': 'OS::Heat::RandomString'}
+                            }
+                        }
+                    }
+                },
+                'type': 'launch_stack'
+            },
+            'groupConfiguration': {
+                'name': 'test_launch_stack',
+                'cooldown': 0,
+                'minEntities': 0,
+                'maxEntities': 10
+            },
+            'scalingPolicies': [],
+        }
+
+        self.group = ScalingGroup(group_config=scaling_group_config,
+                                  treq=self.treq,
+                                  pool=self.pool)
+
+        return self.identity.authenticate_user(
+            self.rcs, resources=get_resource_mapping(), region=region)
+
+    def get_stack_list(self):
+        return (self.treq.get(
+                    '%s/stacks' % self.rcs.endpoints['heat'],
+                    headers=headers(str(self.rcs.token)),
+                    params={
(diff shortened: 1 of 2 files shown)


stretchservers:

stretchservers:

otter/integration/tests/test_launch_stack.py
```diff
@@ -0,0 +1,143 @@
...
+from otter.util.logging_treq import LoggingTreq
+from otter.util.retry import (
+    TransientRetryError,
+    repeating_interval,
+    terminal_errors_except
+)
+
+
+class TestLaunchStack(unittest.TestCase):
+    """Tests making sure Otter can talk to Heat properly."""
```
manishtomar: How about "tests that ensure launch_stack config can be used"?

otter/integration/tests/test_launch_stack.py
```diff
@@ -0,0 +1,143 @@
...
+        yield self.group.start(self.rcs, self)
+
+        yield p.start(self.rcs, self)
+        yield p.execute(self.rcs)
+        yield self.wait_for_stack_list([u'UPDATE_COMPLETE'] * 5)
+
+        yield scale_up.start(self.rcs, self)
+        yield scale_up.execute(self.rcs)
+        yield self.wait_for_stack_list(
+            [u'UPDATE_COMPLETE'] * 5 + [u'CREATE_COMPLETE'] * 2)
```
manishtomar: Non-blocking curiosity: Why are initial stacks created result in `UPDATE_COMPLETE` and later one in `CREATE_COMPLETE`?

manishtomar:
LGTM :)

otter/integration/tests/test_launch_stack.py
dragorosson: Only the final states of the stacks are captured here. In reality, the states from the first policy are:
```
1. [ ]
2. [CREATE_COMPLETE] * 5
3. [CHECK_COMPLETE] * 5
4. [UPDATE_COMPLETE] * 5
```
Going from 0 to N stacks is unusual because there's no `CHECK_` stacks to signal that it's the middle of a set of convergence cycles, so a check/update round is done at the end.

During the second policy execution, these are the states:
```
1. [UPDATE_COMPLETE] * 5
2. [CHECK_COMPLETE] * 5
3. [CHECK_COMPLETE] * 5 + [CREATE_COMPLETE] * 2
4. [UPDATE_COMPLETE] * 5 + [CREATE_COMPLETE] * 2
```
This is a more ordinary scenario (stacks already exist) so the stack checking is done at the beginning. Since there are no stacks in bad states in (3), Otter does stack update to clear the `CHECK_` status and convergence ends.

Finally, it is assured that scaling down to 1 stack will end up with one `UPDATE_COMPLETE` stack because stack checking is done at the beginning and all of the stacks will be in `CHECK_COMPLETE`. All but one are deleted and the remaining one is bumped to `UPDATE_COMPLETE`.

otter/integration/tests/test_launch_stack.py
dragorosson: Done.

dragorosson:
Merge after https://github.com/rackerlabs/mimic/pull/506 merges.

dragorosson:
Done. It seems I can remove `str()` wrapper, so I did.

manishtomar:

otter/integration/tests/test_launch_stack.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        Establish an HTTP connection pool and commonly used resources for each
+        test. The HTTP connection pool is important for maintaining a clean
+        Twisted reactor.
+        """
+        self.helper = TestHelper(self)
+        self.rcs = TestResources()
+        self.identity = get_identity(self.helper.pool)
+        self.pool = HTTPConnectionPool(reactor, False)
+        self.treq = LoggingTreq(log=log, log_response=True)
```
manishtomar: Sorry for the late re-review. I just realized you don't need `self.pool` and `self.treq` as they are available in `self.helper`.

otter/integration/tests/test_launch_stack.py
```diff
@@ -0,0 +1,143 @@
...
+                'minEntities': 0,
+                'maxEntities': 10
+            },
+            'scalingPolicies': [],
+        }
+
+        self.group = ScalingGroup(group_config=scaling_group_config,
+                                  treq=self.treq,
+                                  pool=self.pool)
+
```
manishtomar: Non-blocking: Would it be possible to use [`TestHelper.create_group`](https://github.com/rackerlabs/otter/blob/master/otter/integration/lib/trial_tools.py#L208) instead? It might be complicated to modify that for `launch_stack` config? How about `create_stack_group` function that takes `stack_args` as argument?

otter/integration/tests/test_launch_stack.py
```diff
@@ -0,0 +1,143 @@
...
+            poll, timeout,
+            can_retry=terminal_errors_except(TransientRetryError),
+            next_interval=repeating_interval(period),
+            clock=reactor,
+            deferred_description=(
+                "Waiting for group {} to reach state {}".format(
+                    self.group.group_id, str(expected_states))))
+
+    @inlineCallbacks
+    def test_create(self):
```
manishtomar: Unfortunately this test won't run with latest mimic due to https://github.com/rackerlabs/mimic/issues/513. Until that is resolved, would you mind [skipping](https://github.com/rackerlabs/otter/blob/master/otter/integration/lib/trial_tools.py#L345) this? This way this PR can land and it can be unskipped once that issue is resolved.

otter/integration/tests/test_launch_stack.py
dragorosson: I think that this would be a good idea when there are more tests that involve `launch_stack`.

otter/integration/tests/test_launch_stack.py
dragorosson: Done

otter/integration/tests/test_launch_stack.py
dragorosson: Thanks! Done

manishtomar:
Thanks @dragorosson :)

PR #1800, opened 2015-12-03
fix a bug where the converge endpoint wouldn't use bound log data for…

radix:
… logging


radix:
I've tested this on my dev VM and it now includes the scaling group ID

radix:
ok and now it has a unit test

manishtomar:
LGTM!

PR #1801, opened 2015-12-04
converge?on_error query arg

manishtomar:
to trigger convergence or not when group is in ERROR. Useful in `trigger_convergence.py` to trigger convergence only on non-ERROR group


radix:
LGTM :) Want me to merge now? I guess this hasn't been tested on the dev vm because the trigger_convergence.py script hasn't been updated yet?

radix:
I've tested this on my dev vm with the updated `trigger_convergence.py`. All good!

PR #1802, opened 2015-12-04
option to trigger convergence or not on ERROR group

manishtomar:
Have not tested it yet. My dev vm is broken. Will try to keep testing.


radix:
I've tested this (after hacking the `endpoint` to be hardcoded). Works well! Only triggered convergence for non-error groups.

PR #1810, opened 2015-12-09
Remove unwarranted step to delete a group

manishtomar:
This step made sense before `force` delete was implemented. It is unnecessary complication now.


manishtomar:
@stanzikratel Review please?

PR #1811, opened 2015-12-10
converge endpoint corrections

manishtomar:
also renamed the method filename and reference in other places. This removes reference to converge endpoint so that it is not yet visible. When we finish deploying for all users, it can be relinked.


manishtomar:
@stanzikratel review please. If possible, can you run it against `deconst` to see if the links are fine?

stanzikratel:
@manishtomar lgtm!

PR #1814, opened 2015-12-11
option to list steps executed

manishtomar:
if convergence is triggered on the group when desired is set to actual. This should be run before enabling convergence for tenants to ensure no changes is made in their infra.

### DIFF
--- a/scripts/trigger_convergence.py
+++ b/scripts/trigger_convergence.py
@@ -7,12 +7,18 @@ config file containing region, identity and cassandra info
 Examples:
 `python trigger_convergence -c config.json -g "tenantid:groupId"`
 will trigger convergence on given group(s)
-`python trigger_convergence -c config.json`
+`python trigger_convergence -c config.json --all`
 will trigger convergence on all groups got from cassandra
 """
 
+from __future__ import print_function
+
 import json
 from argparse import ArgumentParser
+from datetime import datetime
+
+from effect import Effect, Func, parallel
+from effect.do import do, do_return
 
 from toolz.curried import keyfilter
 from toolz.itertoolz import concat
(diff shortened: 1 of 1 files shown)


scripts/trigger_convergence.py
```diff
@@ -123,6 +136,35 @@ def get_groups(parsed, store, conf):
...
+    Return Effect of list of steps that would be performed on the group
+    if convergence is triggered on it with desired=actual
+    """
+    now_dt = yield Effect(Func(datetime.utcnow))
+    all_data_eff = convergence_exec_data(
+        group["tenantId"], group["groupId"], now_dt, get_executor)
+    all_data = yield Effect(TenantScope(all_data_eff, group["tenantId"]))
+    (executor, scaling_group, group_state, desired_group_state,
+     resources) = all_data
+    desired_group_state.desired = len(resources['servers'])
```
dragorosson: What if this is a launch_stack group?

manishtomar: It might raise `KeyError` from `resources["servers"]`. However, I am not concerned with that since this script is only used to transition _existing_ tenants with `launch_server` groups to convergence.

scripts/trigger_convergence.py
```diff
@@ -7,12 +7,18 @@
 Examples:
 `python trigger_convergence -c config.json -g "tenantid:groupId"`
 will trigger convergence on given group(s)
-`python trigger_convergence -c config.json`
+`python trigger_convergence -c config.json --all`
```
dragorosson: Is this an unrelated change? --all vs --steps

scripts/trigger_convergence.py
dragorosson: Gotcha.

dragorosson:
LGTM :)

scripts/trigger_convergence.py
dragorosson: (That's what I figured this was for.)

scripts/trigger_convergence.py
manishtomar: Yes. Sorry this is correct example after implementing `--all`. Didn't change it then :)

manishtomar:
@dragorosson Thank you soo much for the review!

Issue #1804, opened 2015-12-07
Destiny.WAIT error reason should mention server id


manishtomar:
`Waiting for temporarily unavailable server to become ACTIVE` error should mention server ID to be clear which server is temporarily unavailable along with their status. It would be awesome if it can mention the amount of time waited. However, that may be difficult with current code structure.

manishtomar:
This has been deployed with #1817.

PR #1817, opened 2015-12-15
Specify ids of temp unavailable servers

manishtomar:
Implements #1804. When a group is put in ERROR due to servers being temporarily unavailable then server ids along with their current status is mentioned in reasons


dragorosson:
LGTM, no comments :)

manishtomar:
Thanks! :smile:

PR #1824, opened 2016-01-13
update kazoo requirements

manishtomar:
to take advantage of certain lock recipe bugfixes


manishtomar:
@glyph This PR pins kazoo that is dependency of txkazoo in otter requirements.txt. txkazoo has loose version dependency on kazoo in its setup.py as it should be to allow broader package combinations. Is this usage of requirements.txt vs setup.py correct?

glyph:
Yes, although @dstufft is the only one who can give you ecclesiastically valid absolution.

manishtomar:

PR #1826, opened 2016-01-15
converging to desired lb state does not remove nodes

### LINKED ISSUE #1803: Use CLB Node Metadata to limit which nodes we clean up
Currently, otter deletes CLB nodes that it "doesn't know about" as a part of convergence. This is done to clean up CLB nodes in the case where the user has changed the configuration from using one Load Balancer to a different one (or just removing the load balancer association entirely). The problem with this is that if a user manually adds auto-scaled servers to a load balancer, we'll immediately delete that association.

We can avoid this by putting a little flag on CLB nodes that indicates that they are otter-managed. I think all it needs to be is a boolean, but we may want to make it point to server ID and group ID to make debugging easier. The behavior in change will be: if this flag exists, then deletion will proceed as it currently does. If the flag is *not* there, we will not delete the node. This way manually added nodes will not be deleted.

This will require:

- `cloud_client.py` function for adding node metadata
- `cloud_client.py` function for *retrieving* node metadata, since apparently it's not returned with the normal list nodes or node details calls. Which means we're more likely to hit rate limits! :-(
- step in otter.convergence.steps that invokes this function
- ILBNode and CLBNode to grow a `metadata` attribute
- `gathering.py` changes to fetch metadata and populate it in the CLBNode
- some changes to `planning.py` to
  - use this step to add metadata after it sees that the node exists
  - check for the metadata flag on the node before deleting the node

manishtomar:
it keeps existing nodes as such and only adds and changes nodes that do match the desired lbs. Fixes #1803.

### DIFF
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -195,20 +195,13 @@ def _converge_lb_state(server, current_lb_nodes):
         for desired in server.desired_lbs - set(met_desireds)
     ]
 
-    # Removes could be replaced with _remove_from_lb_with_draining if
-    # we wanted to support draining for moving load balancers too
-    removes = [
-        remove_node_from_lb(node=node)
-        for node in set(current_lb_nodes) - set(good_nodes)
-    ]
-
     changes = [
         change_lb_node(node=node, description=desired)
         for desired, node in desired_matching_existing
         if node.description != desired
     ]
 
-    return [step for step in (adds + removes + changes) if step is not None]
+    return [step for step in (adds + changes) if step is not None]
 
 
 def _drain_and_delete(server, timeout, current_lb_nodes, now):
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -186,7 +186,7 @@ def _converge_lb_state(server, current_lb_nodes):
         if desired.equivalent_definition(node.description)]
 
     if desired_matching_existing:
-        met_desireds, _ = zip(*desired_matching_existing)
+        met_desireds = next(iter(zip(*desired_matching_existing)))
     else:
         met_desireds = ()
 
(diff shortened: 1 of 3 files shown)


manishtomar:

manishtomar:
jenkins tests failed as expected. Will update convergence tests.

manishtomar:

manishtomar:

manishtomar:

manishtomar:
@glyph This is CLB change where otter does not remove nodes of CLBs it is not configured with we discussed few weeks ago. Could you review this?

glyph:
Sure. Will try to get to it before Friday.

otter/convergence/planning.py
```diff
@@ -186,29 +186,22 @@ def _converge_lb_state(server, current_lb_nodes):
         if desired.equivalent_definition(node.description)]
 
     if desired_matching_existing:
-        met_desireds, good_nodes = zip(*desired_matching_existing)
```
glyph: Non-blocking, and not even really about this change: Personally I dislike the use of `_` as a variable name. I think it's an unnecessary and confusing idiom.  This is not really a comment on this change, but using `zip` in this way when you expect a fixed number of results is weird.

Rather than make up a name for this variable though, you could just compress this whole conditional suite into:

```python
met_desireds = next(iter(zip(*desired_matching_existing), ()))
```

glyph:
This looks good to me. Love the diffstat :).

otter/convergence/planning.py
manishtomar: Honestly I don't like the unzipping functionality in `zip` function. There should've been `unzip` for that. But I can understand the discomfort with `_`. Will change it :)

manishtomar:

manishtomar:

PR #1829, opened 2016-01-20
create server steps limit is configurable

manishtomar:
This along with https://github.com/rackerlabs/autoscaling-chef/pull/812 will allow 100 servers to be created in one convergence cycle.

PR #1830, opened 2016-01-21
actual step limits configurable

manishtomar:
steps limit is taken from config when creating `Converger` and it is passed all the way down to planner which limits the number of steps executed. 

This fixes for previous not working #1829 PR. https://github.com/rackerlabs/autoscaling-chef/pull/813 sets the respective config.

### DIFF
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -533,7 +533,8 @@ def plan_launch_server(desired_group_state, now, build_timeout, step_limits,
     """
     Get an optimized convergence plan.
 
-    Takes the same arguments as :func:`converge_launch_server`.
+    Takes the same arguments as :func:`converge_launch_server`
+    except `step_limits` which is dict of step class -> limit
     """
     steps = converge_launch_server(desired_group_state, servers, lb_nodes, now,
                                    timeout=build_timeout)
@@ -548,7 +549,8 @@ def plan_launch_stack(desired_group_state, now, build_timeout, step_limits,
 
     The arguments `now` and `build_timeout` are ignored and only necessary to
     match those of `plan_launch_server`. The arguments `desired_group_state`
-    and `stacks` are the same as in `converge_launch_stack`.
+    and `stacks` are the same as in `converge_launch_stack`. `step_limits` is
+    dict of step class -> limit
     """
     steps = converge_launch_stack(desired_group_state, stacks)
     steps = limit_steps_by_count(steps, step_limits)
--- a/otter/convergence/planning.py
+++ b/otter/convergence/planning.py
@@ -528,8 +528,8 @@ def converge_launch_stack(desired_state, stacks):
                 converge_later)
 
 
-def plan_launch_server(desired_group_state, now, build_timeout, servers,
-                       lb_nodes):
+def plan_launch_server(desired_group_state, now, build_timeout, step_limits,
+                       servers, lb_nodes):
     """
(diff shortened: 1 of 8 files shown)


manishtomar:

manishtomar:
@dragorosson Yesterday's #1829 didn't work because config is not available until `makeService` is called. So the code change didn't have any impact. This is correct fix. I tested it in dev vm. If you don't mind, can you have a look when you've time? Thanks!

otter/tap/api.py
```diff
@@ -317,11 +318,10 @@ def setup_converger(parent, kz_client, dispatcher, interval, build_timeout,
         kz_client=kz_client,
         interval=interval,
         partitioner_path=CONVERGENCE_PARTITIONER_PATH,
-        time_boundary=15,  # time boundary
+        time_boundary=1,  # time boundary
```
dragorosson: What is this? I see in a docstring "time to wait for partitioner to stabilize". Partly, I'm curious, and I want to make sure this is an intentional change as well.

otter/convergence/service.py
```diff
@@ -550,7 +552,7 @@ def eff_finally(eff, after_eff):
 @do
 def converge_one_group(currently_converging, recently_converged, waiting,
                        tenant_id, group_id, version,
-                       build_timeout, limited_retry_iterations,
+                       build_timeout, limited_retry_iterations, step_limits,
```
dragorosson: Needs corresponding docstring.

otter/convergence/service.py
```diff
@@ -272,7 +273,8 @@ def _clean_waiting(waiting, group_id):
 
 @do
 def execute_convergence(tenant_id, group_id, build_timeout, waiting,
-                        limited_retry_iterations, get_executor=get_executor):
+                        limited_retry_iterations, step_limits,
```
dragorosson: Needs corresponding docstring.

otter/tap/api.py
manishtomar: Oh good catch. That was not meant to be committed. I was just testing it. Will change it back. It is the time partitioner waits before it allocates the buckets. Say if it is 15 seconds and no nodes must join or leave the partition for 15 seconds before it says allocated.

otter/convergence/planning.py
```diff
@@ -537,11 +537,12 @@ def plan_launch_server(desired_group_state, now, build_timeout, servers,
     """
```
dragorosson: I can't comment on the same line (536), but I suppose this is not true anymore.

otter/convergence/planning.py
```diff
@@ -550,7 +551,7 @@ def plan_launch_stack(desired_group_state, now, build_timeout, stacks):
     and `stacks` are the same as in `converge_launch_stack`.
```
dragorosson: Not true anymore, either.

otter/convergence/service.py
```diff
@@ -734,7 +738,7 @@ class Converger(MultiService):
 
     def __init__(self, log, dispatcher, num_buckets, partitioner_factory,
                  build_timeout, interval,
-                 limited_retry_iterations,
+                 limited_retry_iterations, step_limits,
```
dragorosson: Docstring

dragorosson:
LGTM besides the time_boundary thing and the docstring stuff. Fix whatever you feel should be fixed with the docstring stuff. :)

PR #1832, opened 2016-01-27
updating testtools and pyrsistent

manishtomar:
since installed versions of testtools and pyrsistent were incompatible

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ treq==0.2.1
 silverberg==0.1.12
 pyOpenSSL==0.14
 jsonfig==0.1.1
-testtools==0.9.32
+testtools==1.9.0
 croniter==0.3.5
 txkazoo==0.0.6b2
 kazoo==2.2.1
@@ -17,7 +17,7 @@ txeffect==0.9
 characteristic==14.3.0
 attrs==15.0.0
 toolz==0.7.4
-pyrsistent==0.10.3
+pyrsistent==0.11.10
 singledispatch==3.4.0.3
 six==1.9.0
 sumtypes==0.1a4


manishtomar:

manishtomar:

manishtomar:
@glyph review please?

glyph:
@manishtomar - LGTM, although I suspect I am missing some nuances of such a subtle change :).  How did these versions become incompatible?

manishtomar:
@glyph https://as-jenkins.rax.io/job/otter/lastBuild/console

glyph:
but... they were pinned

manishtomar:
The pinning didn't help when [dev_requirements](https://github.com/rackerlabs/otter/blob/master/dev_requirements.txt) was installed on the same virtualenv. `python-subunit` in dev_requirements required `testtools>=0.9.34` which resulted in uninstallation of pinned version and installation of _latest_ version of testtools. I was not sure if pip did the right thing of removing a pinned version. This leads me to confusion about maintaining `dev_requirements.txt`. How to describe the packages required for developing if it results in unexpected package installations? Moreover, I think otter's use of requirements.txt is incorrect. All the packages which it is _explicitly_ using should goto setup.py. This is based on my understanding of what we discussed last. Please let me know your thoughts.

glyph:
`setup.py` should generally not pin exact versions whereas `requirements.txt` generally should.  Mimic is facing this same issue regarding unexpected package installations, though; one way we're mitigating it currently is just installing the dev tools into a different virtualenv and always testing with `tox`; in the docker image we do not install dev stuff at all.

When Mimic comes up with a way to ensure spurious version upgrades are not installed, I'll let you know :).

PR #1835, opened 2016-02-02
Tests get imageRef from nova

manishtomar:
instead of configured one. This overcomes https://github.com/rackerlabs/mimic/issues/513 and allows us to use latest mimic.


manishtomar:
@glyph review please? Meanwhile, let me see how to deploy pinned mimic version in this repo instead of constantly changing the git revision in https://github.com/rackerlabs/autoscaling-chef

glyph:
Goin' in the queue…

manishtomar:

manishtomar:

manishtomar:

autoscale_cloudcafe/autoscale_fixtures/config.py
```diff
@@ -10,6 +10,9 @@ class AutoscaleConfig(ConfigSectionInterface):
     """
     SECTION_NAME = 'autoscale'
 
+    lc_image_ref = None
```
glyph: Mutable staaaaate :-(

I do like the fact that this makes the data structure more regular-python-ish and less of a wrapper around a weird dictionary.  But I don't like the fact that this change results in even less documentation for these two properties.  It took me a minute to understand what they were supposed to be, and why this was in a "config section" at all.  Can you add some `:ivar:` documentation to the class docstring?

Also, these are default-NULL, which I am coming to believe is always an antipattern.  I wrote some about this in Python: https://glyph.twistedmatrix.com/2015/09/python-option-types.html and of course there's the option of `sumtypes`.  When and why are these values `None`?

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -132,7 +141,24 @@ def _set_up_clients():
 
 _rcv3_cloud_network = autoscale_config.rcv3_cloud_network
 
-autoscale_client, server_client, lbaas_client, rcv3_client = _set_up_clients()
+(autoscale_client, server_client, images_client,
+ lbaas_client, rcv3_client) = _set_up_clients()
+
+
+def fetch_image_ids(images_client):
```
glyph: given that this hard-codes `Ubuntu`, either:

1. it should be called `fetch_ubuntu_image_ids`, or
2. it should be `def fetch_image_ids(images_client, name="ubuntu"):`

I guess this is a blocking call?

autoscale_cloudroast/test_repo/autoscale/fixtures.py
```diff
@@ -214,6 +238,10 @@ def setUpClass(cls):
         cls.lbaas_client = lbaas_client
         cls.rcv3_client = rcv3_client
 
+        cls.lc_image_ref, cls.lc_image_ref_alt = fetch_image_ids(images_client)
```
glyph: OK, I am beginning to see why this was a pain to update. This is a _lot_ of global mutable state. Is this what *all* cloudcafe code looks like?

otter/integration/lib/nova.py
```diff
@@ -141,6 +141,34 @@ def list_servers(rcs, pool, _treq=treq):
...
+        params=params,
+        headers=headers(str(rcs.token)),
+        pool=pool
+    ).addCallback(check_success, [200]).addCallback(_treq.json_content)
+
+
+@inlineCallbacks
+def fetch_image_id(rcs, pool):
+    """
+    Get image ID from nova that can be used in creating servers or
```
glyph: Including a cross-reference link to the cloudcafe version, since these are really the same function, would be handy.  Might it be a good idea to make the signature the same, and `returnValue` a 2-tuple of ubuntu, non-ubuntu?

glyph:
Once again sorry for the delay.

This mostly seems fine; :+1: . I have a lot of concerns about the cloudcafe code, but most of that stuff is not really related to this PR.  My general comment is maybe take a little more care in documenting parameter types and return values since not all of this stuff is very straightforward and I am betting that you will be cursing yourself in 6 months if you come back to it and are trying to remember why the twisted and blocking image-ID fetchers work differently :-).  So maybe refine the names or add a little bit more docs, but otherwise it looks like it's doing the right thing for mimc.

autoscale_cloudroast/test_repo/autoscale/fixtures.py
manishtomar: Yeah. All of cloudcafe stuff is blocking.

autoscale_cloudroast/test_repo/autoscale/fixtures.py
manishtomar: Pretty much. Honestly, I wasn't even thinking about mutable state / better patterns when doing this. I just wanted to get this done to move forward.

otter/integration/lib/nova.py
manishtomar: Cross-referencing is a great idea. Will do, however I am not sure if this function needs to return non-Ubuntu image as it is not used in tests anywhere.

manishtomar:
> Once again sorry for the delay.

No problem at all :smile: . I should say a big thanks for reviewing such a messy code. I can try to understand how confusing and horrible this can look from external eyes :confused:

manishtomar:

PR #1838, opened 2016-02-03
logging rcv3 bulk response

manishtomar:
to debug why it has been retrying forever

### DIFF
--- a/otter/convergence/steps.py
+++ b/otter/convergence/steps.py
@@ -429,11 +429,14 @@ def _rcv3_check_bulk_add(attempted_pairs, result):
     Checks if the RCv3 bulk add command was successful.
     """
     response, body = result
+    success_return = (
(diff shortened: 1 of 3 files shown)


manishtomar:
Moving RCv3 code to `cloud_client`. Will add logging with that. Hence closing this.

manishtomar:
Implemented a quick fix on this branch that does not continously retry. Instead returns step RETRY. This should hopefully behave well with RCv3 and not cause issues.

manishtomar:

manishtomar:

manishtomar:

manishtomar:
Since RCv3 steps are being used by worker also, removing internal retry will cause worker to fail on partial add success. Hence, the real solution is to limit number of internal retries. Will open that in another PR.

PR #1848, opened 2016-02-29
convergence tenants metrics

manishtomar:
adding `reg.conv_desired|actual|divergence` metric that is total desired and actual of convergence tenants groups only. Graphing this show true reliability of convergence.


manishtomar:
@dragorosson review please

dragorosson:
LGTM

PR #1849, opened 2016-03-01
Filter launch stack groups

manishtomar:
While testing #1848 I found that `launch_stack` groups skewing the metric. Hence filtering it out. Might need to implement separate metric that involves fetching stacks instead of servers for these kind of groups.


manishtomar:
@dragorosson review?

dragorosson:
LGTM. The nit is a non-blocker, but I just wanted to bring it to your attention. Also, I can't help but wonder why `get_all_groups` isn't named `get_all_valid_groups` if that's what it does unconditionally (still, non-blocker). :+1:

manishtomar:
> why get_all_groups isn't named get_all_valid_groups

This is a brilliant question and the only answer I can think is "I am stupid". If caller wants _all_ groups, they can call `get_all_group_rows`. I'll change it. Thanks!

PR #1850, opened 2016-03-01
Convergence tenants metrics fix

manishtomar:
Turns out I didn't set the global config data via `set_config_data` to use `config_value` in #1848 :confused: . However, instead of doing that I am passing config to the function that is done throughout `metrics.py` which is much better than managing global state. Functional programming yo.. (i miss @radix). Have tested this.

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -42,7 +42,6 @@ from otter.effect_dispatcher import get_legacy_dispatcher, get_log_dispatcher
 from otter.log import log as otter_log
 from otter.models.cass import CassScalingGroupCollection
 from otter.models.intents import GetAllGroups, get_model_dispatcher
-from otter.util.config import config_value
 from otter.util.fp import partition_bool
 
 
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -164,7 +164,7 @@ def calc_total(group_metrics):
 
 
 @do
-def add_to_cloud_metrics(ttl, region, group_metrics, no_tenants, config,
+def add_to_cloud_metrics(ttl, region, group_metrics, num_tenants, config,
                          log=None, _print=False):
     """
     Add total number of desired, actual and pending servers of a region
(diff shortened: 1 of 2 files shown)


manishtomar:
@dragorosson sorry review again please

radix:
:heart:

glyph:
:heart_decoration:

wirehead:
`<3`

otter/test/test_metrics.py
```diff
@@ -222,7 +221,8 @@ def test_added(self):
                 ServiceType.CLOUD_METRICS_INGEST, "POST", "ingest",
                 data=req_data, log=log).intent, noop)
         ]
-        eff = add_to_cloud_metrics(m['ttlInSeconds'], 'ord', metrics, 3, log)
+        eff = add_to_cloud_metrics(m['ttlInSeconds'], 'ord', metrics, 3,
```
dragorosson: `3` feels very magical. Had to look at the docstring for `add_to_cloud_metrics` to know that `3` == `no_tenants=3` == `number of tenants = 3`. I'd prefer this to be a kwarg.

otter/metrics.py
```diff
@@ -165,8 +164,8 @@ def calc_total(group_metrics):
 
 
 @do
-def add_to_cloud_metrics(ttl, region, group_metrics, no_tenants, log=None,
-                         _print=False):
+def add_to_cloud_metrics(ttl, region, group_metrics, no_tenants, config,
```
dragorosson: It'd also be nice if `no_tenants` could be renamed `num_tenants` or something, since "no" is a confusing abbreviation for "number".

dragorosson:
LGTM, save for the minor nit. Hi @radix and @wirehead!

otter/test/test_metrics.py
manishtomar: I agree with `3` seeming magical which is why I've added comment here. However, I am not so sure about making this kwarg due to this reason. It does help in increasing the readability but it also raises question about using positional args in general. When is it ok to have positional arg and when does it make the code unreadable. I think I'll keep this as it is unless you have a _very_ strong feeling against it. Thanks for the suggestion though :)

dragorosson:
LGTM

wirehead:
Otter

As I've long said: Death will not release you from otter-dom. :)

manishtomar:
Thanks @dragorosson :)

PR #1853, opened 2016-03-02
pip install in one pass

manishtomar:
This will disallow different versions of common dependent packages between requirements.txt and dev_requirements.txt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,7 @@ cache:
 
 install:
     - pip install -U pip wheel
-    - pip install -r requirements.txt -r dev_requirements.txt
-    - pip install coveralls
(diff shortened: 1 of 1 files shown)


manishtomar:
@glyph review?

glyph:
This won't catch *any* broken dependencies, but it's a step in the right direction :).

manishtomar:
Wow. This is the fastest you've responded. Looks like this PR didn't go through your task queue :). What would be the right description for the PR? "This will disallow different versions of common dependent packages between requirements.txt and dev_requirements.txt"?

glyph:
Yes, that sounds right.

manishtomar:
> but it's a step in the right direction :).

Assuming this is LGTM and merging :)

Issue #1855, opened 2016-03-07
Self-healing in otter


manishtomar:
Currently self-healing is done via calling `trigger_convergence.py` every 5 mins on metrics node via cron. This has couple of disadvantages:

1. Script doesn't have tests
2. No monitoring: 
   - cron MAILTO needs mailx and sendmail setup on metrics node
3. Deployment is a two step process - deploy otter and metrics
4. Breaks if metrics node is unavailable

A better approach is to add self-healing as a Twisted service in otter where it solves all the above problem.

PR #1856, opened 2016-03-07
trigger_convergence.py get_all_groups fix

manishtomar:
This script broke with #1849 since it used the older `get_all_groups` function. Now using renamed function `get_all_valid_groups`. This shows lack of regression tests and monitoring for this script. I think its better to move away from the script and add self-healing in otter-api itself via simple kazoo leader election. See #1855.


dragorosson:
Oopsies :)

PR #1858, opened 2016-03-08
Give launch_stack integration test more time

dragorosson:
The integration test is timing out because convergence runs approximately every 30 seconds.


manishtomar:

PR #1869, opened 2016-03-30
Filter error group servers

### LINKED ISSUE #1867: Filter ERROR group servers in metrics
ERROR groups tend to leave servers behind which get added to metrics as servers of deleted groups. This is incorrect and should be filtered out just like ERROR groups are filtered their servers should also be filtered.

manishtomar:
Fixes #1867. Note that even though behavior of `get_all_valid_groups` is changed, it should not have impact on `trigger_convergence.py` since it might want to trigger convergence on ERROR groups. In fact it fixes the script's behavior when `no_error=true` is passed.

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -70,6 +70,8 @@ def get_tenant_metrics(tenant_id, scaling_groups, grouped_servers,
         servers = grouped_servers.get(group_id, [])
         if group_id in groups:
             group = groups[group_id]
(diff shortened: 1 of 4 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/build-6053daa2a6/docs/autoscale/v1/developer-guide/)

manishtomar:
@dragorosson review please :)

dragorosson:
LGTM

PR #1873, opened 2016-04-12
use get_destiny to calculate active and pending servers

manishtomar:
since that is used by convergence. This way metrics handles all the server statues. For example, it now handles SUSPENDED state through `get_destiny`.

### DIFF
--- a/otter/metrics.py
+++ b/otter/metrics.py
@@ -22,6 +22,7 @@ from silverberg.cluster import RoundRobinCassandraCluster
 from toolz.curried import filter, get_in
 from toolz.dicttoolz import keyfilter, merge
 from toolz.itertoolz import groupby
+from toolz.recipes import countby
 
 from twisted.application.internet import TimerService
 from twisted.application.service import Service
@@ -75,12 +76,12 @@ def get_tenant_metrics(tenant_id, scaling_groups, grouped_servers,
             group = {'groupId': group_id_from_metadata(servers[0]['metadata']),
                      'desired': 0}
         servers = map(NovaServer.from_server_details_json, servers)
-        dservers = groupby(get_destiny, servers)
-        active = len(dservers.get(Destiny.CONSIDER_AVAILABLE, [])) + \
-            len(dservers.get(Destiny.AVOID_REPLACING, []))
-        ignore = len(dservers.get(Destiny.DELETE, [])) + \
-            len(dservers.get(Destiny.CLEANUP, [])) + \
-            len(dservers.get(Destiny.IGNORE, []))
+        counts = defaultdict(lambda: 0)
+        counts.update(countby(get_destiny, servers))
+        active = counts[Destiny.CONSIDER_AVAILABLE] + \
+            counts[Destiny.AVOID_REPLACING]
+        ignore = counts[Destiny.DELETE] + counts[Destiny.CLEANUP] + \
+            counts[Destiny.IGNORE]
         yield GroupMetrics(tenant_id, group['groupId'], group['desired'],
                            active, len(servers) - ignore - active)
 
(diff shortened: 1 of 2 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/build-09f48159fe/docs/autoscale/v1/developer-guide/)

manishtomar:
@glyph review please.

glyph:
@manishtomar Can you describe this PR a bit better?  You say "active and pending" but the actual code says "suspended".  What's the difference?  I _think_ I can infer this from the issue review I did yesterday but I can't really be sure.

manishtomar:
Metrics service calculates active and pending servers on each tenant and ingests it in Rackspace Metrics from where the dashboard is created. Currently it calculates that by explicitly looking at server status which duplicates work already done in planner under DESTINY. Hence, using that.

glyph:
If the metrics are the same, just using a different gathering mechanism, then why is the expected data in the test changing?

manishtomar:
> If the metrics are the same, just using a different gathering mechanism, then why is the expected data in the test changing?

To test new status SUSPENDED which was not checked in earlier code but is expected to be checked by `get_destiny`. Does this help?

glyph:
> To test new status SUSPENDED which was not checked in earlier code but is expected to be checked by get_destiny. Does this help?

Aah, I see.  So there's a new _status_ (i.e. an input to this algorithm), but there's no new group _metrIc_ (i.e. an output of it).  I was misreading the use of the `servers` variable in the test.

So then here are two pieces of feedback:

1. you're adding a new status, but you're just adjusting the existing test data to include some of every kind of status.  There is a whole potential future bug-surface for input values that are missing certain statuses, and as we add new statuses there should be new input/output test pairs.  For this change, you should add at least one additional test method to the `GetTenantMetricsTests` suite, and perhaps file a follow-up to add more simpler sets of servers (for example, a test case with just one server in each state).
2. update the description of this PR to explain that the `SUSPENDED` state was previously unhandled, and this change adds handling for it.

Do both of these make sense?

manishtomar:
> or this change, you should add at least one additional test method to the GetTenantMetricsTests suite, and perhaps file a follow-up to add more simpler sets of servers (for example, a test case with just one server in each state).

I think its ok to have multiple statues in same test. Its just less code. I'll add comments next to each status to clarify what it is checking. If that doesn't seem to clarify, then we can revisit this.

> update the description of this PR to explain that the SUSPENDED state was previously unhandled, and this change adds handling for it.

Sure. Will do :)

glyph:
> I think its ok to have multiple statues in same test. Its just less code. I'll add comments next to each status to clarify what it is checking. If that doesn't seem to clarify, then we can revisit this.

My point is that it's a different code path to have or not have a given status, so there should be test coverage of different points in that matrix.  For example, let's say just _one_ of the invocations of `.get()` accidentally failed to pass `[]` as its second arg; this would cause an exception, but right now, there is no test for that possible exception.

This is a minor issue but this sort of thing tends to build up over time.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/build-e843ff8a5d/docs/autoscale/v1/developer-guide/)

manishtomar:
@glyph I am sorry I still don't see the necessity of new tests as the current test covers the case `.get` not being called with `[]`. [Group g2](https://github.com/rackerlabs/otter/blob/metrics-svr-destiny/otter/test/test_metrics.py#L81) has no servers which handles case where either of the destiny doesn't return any servers. If it helps I've changed the implementation to use `defaultdict` which reduces chances of error. Not that it should affect coverage. Ideally, we wouldn't have this conversation if a proper coverage tool reported 100%. I've "use codecov.io" among list of things to do.

I am going to merge this. I hope this does not come as me not valuing your opinion which I always do :).

glyph:
Unfortunately, the second coverage code path is in the C code for `dict.get`, so it would be very hard for coverage to report this (it isn't even branch coverage, it's coverage-in-combination, which no Python tool I know of does).  But I can submit a PR for the tests later :).

PR #1880, opened 2016-04-21
trigger_convergence.py waits for all groups to converge

manishtomar:
Earlier it returned immediately even if one group errored. Some groups were not converged due to this. Also prints the errored groups.


rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/build-ba063789f8/docs/autoscale/v1/developer-guide/).

PR #1885, opened 2016-05-03
step_limits is a dict - not an int

manishtomar:
This was causing failure to print steps of some groups that interacted with step limits.


rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/build-ee185f0011/docs/autoscale/v1/developer-guide/).

PR #1887, opened 2016-05-09
Self heal service

### LINKED ISSUE #1855: Self-healing in otter

manishtomar:
Fixes #1855.


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/build-b45b4b6eaf/docs/autoscale/v1/developer-guide/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/build-9fe5a6116a/docs/autoscale/v1/developer-guide/).

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/build-1500ed8b60/docs/autoscale/v1/developer-guide/).

manishtomar:

manishtomar:
@glyph review please.

manishtomar:
Found a small problem while testing. Will update shortly.

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/build-b1ea38c9a7/docs/autoscale/v1/developer-guide/)

manishtomar:
@glyph Fixed the issue. This is working now. Kindly review it when you find time.

manishtomar:
Sorry another issue. Will update.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/build-03e64c5ace/docs/autoscale/v1/developer-guide/)

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/build-057211cdbb/docs/autoscale/v1/developer-guide/). :bow:

manishtomar:

manishtomar:
@glyph Finally ready for review :)

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,169 @@
+"""
+Self heal service
```
glyph: This could do with a few additional words as to what "self heal" is :)

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,169 @@
...
+from otter.log.intents import with_log
+from otter.models.intents import GetAllValidGroups, GetScalingGroupInfo
+from otter.models.interface import ScalingGroupStatus
+from otter.util.zk import GetChildren
+
+
+class SelfHeal(MultiService, object):
+    """
+    A service that triggers convergence on all the groups on interval basis.
+    Only one node is allowed to do this.
```
glyph: Non-blocking here but I really feel like a few `:ivar:`s here would not be misplaced.  There's a lot of state on this object and none of it is explained.

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        d = is_lock_acquired(self.disp, self.lock)
+        return d.addCallback(lambda b: (True, {"has_lock": b}))
+
+    def _perform(self):
+        d = perform(self.disp, get_groups_to_converge(self.config_func))
+        d.addCallback(self._setup_converges)
+        return d.addErrback(self.log.err, "self-heal-err")
+
+    def _setup_converges(self, groups):
```
glyph: stylistic, non-blocking:

This method is used only once, as a callback.  I generally prefer to avoid polluting the namespace of an object, even with private methods, if they're used only once as callbacks.  I'd define this as nested within `_perform` or even maybe make `_perform` an `inlineCallbacks` function.

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,169 @@
...
+        Stop service by cancelling any remaining scheduled calls and releasing
+        the lock
+        """
+        super(SelfHeal, self).stopService()
+        self._cancel_scheduled_calls()
+        return self.lock.release()
+
+    def _cancel_scheduled_calls(self):
+        """
+        Cancel any remaining scheduled calls
```
glyph: This is missing a `:return:`; it cancels calls but it also has a return value.

otter/test/convergence/test_selfheal.py
```diff
@@ -0,0 +1,327 @@
...
+        self.s._perform = mock.Mock(return_value=fail(ValueError("h")))
+        self.s.startService()
+        self.s._perform.assert_called_once_with()
+        self.log.err.assert_called_once_with(
+            CheckFailure(ValueError), otter_service="selfheal")
+        self.s._perform.return_value = succeed(None)
+        self.clock.advance(300)
+        self.assertEqual(self.s._perform.call_count, 2)
+
+    @mock.patch("otter.convergence.selfheal.check_and_trigger")
```
glyph: Why do you need to `mock.patch` a function that returns an `Effect`?  Isn't the whole _point_ of `Effect` that you don't have to do this?

otter/tap/api.py
```diff
@@ -278,17 +279,22 @@ def on_client_ready(_):
                                              get_service_configs(config),
                                              kz_client, store, supervisor,
                                              cassandra_cluster)
+
             # Setup scheduler service after starting
             scheduler = setup_scheduler(parent, dispatcher, store, kz_client)
-            health_checker.checks['scheduler'] = scheduler.health_check
-            otter.scheduler = scheduler
+            if scheduler is not None:
```
glyph: This is making me deeply suspicious - I don't fully understand what's going on, but all I see in this PR right now is
```
+        mock_setup_scheduler.return_value = None
```
I don't see anywhere that `setup_scheduler` itself has been modified to return a different value.

If it really needs to return a new type, could you consider an Active Nothing here?

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,169 @@
...
+    def health_check(self):
+        """
+        Return about whether this object has lock
+        """
+        d = is_lock_acquired(self.disp, self.lock)
+        return d.addCallback(lambda b: (True, {"has_lock": b}))
+
+    def _perform(self):
+        d = perform(self.disp, get_groups_to_converge(self.config_func))
+        d.addCallback(self._setup_converges)
```
glyph: non-blocking, but:

I have some general feedback about the algorithm you're using to do the timed work here.

It involves managing a lot of state, so it's somewhat error prone.  I don't spot any _actual_ errors but I suspect it may be a bit tricky to maintain.

Rather than having one looping call which sets up another set of looping calls, which, if the network slows down, might fail to apply backpressure and therefore end up backing up, I think you should switch to a service which has only a single `LoopingCall` that just figures out what the "next thing to do" is.

For example, initially, the first thing to do is to get the groups to converge.  Then you can return that `Deferred` from your `LoopingCall` so you won't be called again until it's done.  Once you've got the groups, then the next "thing to do" is to converge all the groups.  This batch-convergence should probably be wrapped in a `gatherResults`.  When you run out of groups and they're all converged, the next thing to do is to get groups to converge again.  This little state machine can completely eliminate the management of outstanding calls, since you'd only have one outstanding delayed call at a time.  This would involve having a much higher frequency.

Does this make sense as a follow-up change?  ideally most of the tests would be similar...

glyph:
@manishtomar Sorry for the lag time, but trying to land this right before PyCon was bad timing :).  Please have a look at my comments and let me know if you need another review.

otter/convergence/selfheal.py
manishtomar: > if the network slows down, might fail to apply backpressure and therefore end up backing up,

Hmm. My understanding is each looping call is independent of other looping calls and slow down of one/some/all of them should not affect each other. I specifically chose the algorithm because of this. 

> This batch-convergence should probably be wrapped in a gatherResults

I don't want to use `gatherResults` as they will all be triggered at same time. I want to space out triggering evenly over the 5 min period. 

I understand the discomfort with extra state. In fact in the beginning I tried to do "get all groups and trigger convergence spaced out" as an effect function which would be called on each interval but it would result in one bad/slow call blocking other calls.

glyph: The problem with spreading convergence over a range of time like this is: let's say it takes 8 minutes to complete a run of convergence of all groups.  But the interval is set to 5 minutes.  Slowly, over time, the delta of 3 minutes worth of in-memory state and in-progress I/O builds up, eventually running your server out of memory.

Of course, *one* slow request shouldn't halt *all* progress, so instead, you probably want to use something like `DeferredSemaphore`, and have a maximum acceptable level of parallelism; just do work as fast as the API servers can keep up.

manishtomar: > The problem with spreading convergence over a range of time like this is: let's say it takes 8 minutes to complete a run of convergence of all groups.

Just to be clear, _running_ of convergence is distributed equally among the nodes using partitioner. Only _triggering_ is done by this one node. This involves making one call to CASS and ZK which should not take time (1-2 ms during normal operation). If there are network issues then health check script would alert and we would take care of the server long before it takes too much memory. In fact every IO (scheduler, API, convergence, etc) would block and take memory during network failure. 

> just do work as fast as the API servers can keep up.

My primary concern is upstream services instead of our API servers. I would prefer to keep reduced load on upstream services as much as possible. Memory/network will be monitored on our servers and self-healing will switch to another node when one node's network is not healthy.

Unless you have very strong opinions against this, I was hoping it is ok. We can continue to monitor the server memory with this and see how it goes.

otter/test/convergence/test_selfheal.py
manishtomar: Because the returned effect with have a `Func` intent due to `do` decorator that needs to be resolved to have its callback called to know what parameters `check_and_trigger` was called with. Mock is not used when we are completely in effect world (see tests for `check_and_trigger` and `is_lock_acquired`) but I find it difficult to write tests without mock in mixed case like this one. I mean its doable but gets complicated.

otter/convergence/selfheal.py
manishtomar: Sounds good. Thanks!

otter/convergence/selfheal.py
manishtomar: Will do.

otter/convergence/selfheal.py
manishtomar: Sure. Will do.

otter/convergence/selfheal.py
glyph: Thanks, this is a little clearer.  If the work is partitioned I guess the individual API nodes can be smart and can (and do?) take out a lock to avoid partitioning.

I do still think that having a single timer and a state machine would be simpler than a lot of timers and a bunch of ad-hoc state but I definitely do not have very strong feelings about that; please feel free to proceed with this implementation.

otter/tap/api.py
manishtomar: > I don't see anywhere that setup_scheduler itself has been modified to return a different value.

https://github.com/rackerlabs/otter/pull/1887/files#diff-17282da832855fb34e8e70ff0ec5ddd1R349. This is just explicit. Actually, `setup_scheduler` had always returned `None` when it didn't find scheduler config but that case was never handled by the caller (this code). This code now handles that case.

> If it really needs to return a new type, could you consider an Active Nothing here?

Sorry what is `Active Nothing`. To be clear, `setup_scheduler` returns scheduler service object based on the config. If config is not found, it returns `None`. I'll add this to the docstring.

glyph: Thanks.

Active Nothing is more popularly known as Null Object https://en.wikipedia.org/wiki/Null_Object_pattern but I like Active Nothing because Null Object sounds more like it might be just None :).

In this case it would mean that `setup_scheduler` would define a type of a given interface, and it would never return `None`, but it might return one of two different implementations of that interface, one of which does nothing.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/build-215843105b/docs/autoscale/v1/developer-guide/). :bow:

otter/convergence/selfheal.py
manishtomar: Does the description in the service object: `A service that triggers convergence on all the groups on interval basis. Only one node is allowed to do this.` serve the purpose?

manishtomar:

manishtomar:
@glyph Made review suggestions except https://github.com/rackerlabs/otter/pull/1887#discussion_r64288911. Would you like to review again? Thanks!

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/build-071b451ce1/docs/autoscale/v1/developer-guide/). :bow:

manishtomar:

manishtomar:

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    A service that triggers convergence on all the groups on interval basis.
+    Only one node is allowed to do this.
+
+    :ivar disp: Effect dispatcher to perform all effects (ZK, CASS, log, etc)
+    :type disp: Either :obj:`ComposedDispatcher` or :obj:`TypeDispatcher`
+
+    :ivar log: Twisted logger with .msg and .err methods
+    :ivar lock: Lock object from :obj:`TxKazooClient`
+    :ivar :obj:`TxKazooClient` kz_client: Twistified kazoo client object
```
glyph: Is this valid ReST syntax?  I am pretty sure the thing that goes after `:ivar ...` is implicitly an `:obj:` anyway.

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
+"""
+Self heal service.
```
glyph: It would be helpful to describe in a few brief sentences what the "self heal service" is.

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+        calling _perform
+        """
+        if self.kz_client.state != KazooState.CONNECTED:
+            self.log.err(RuntimeError("self-heal-kz-state"),
+                         "self-heal-kz-state", state=self.kz_client.state)
+            returnValue(None)
+        # The reason why it checks everytime by talking to ZK is because
+        # we could've lost the lock if there where connection changes between
+        # subsequent intervals and it is not clear how state change to
+        # SUSPENDED should be handled
```
glyph: It seems like this pattern really ought to be something like `ensure_lock_acquired`.  What you're trying to express, semantically, with the following block of code, is that `self.lock` governs mutually exclusive access to the resource represented by `self._perform`.  In order to achieve that, though, you have the actual application logic - `self._perform(config_func, time_range)` - repeated twice.  Can you factor out this logic so it can be more easily independently tested, and re-used for anything (probably any lock-protected work) which many need to repeat this pattern?

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+                self.log.msg("self-heal-lock-acquired")
+                yield self._perform(config_func, time_range)
+            except LockTimeout:
+                # expected. Nothing to do here. Will try on next interval
+                pass
+
+
+def get_groups_to_converge(config_func):
+    """
+    Get all tenant's all groups that needs convergence triggering
```
glyph: once again with the `:param:`s - what's `config_func`?  :)

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+
+    if state.status == ScalingGroupStatus.ACTIVE and (not state.paused):
+        yield with_log(
+            trigger_convergence(tenant_id, group_id),
+            tenant_id=tenant_id, scaling_group_id=group_id)
+
+
+def is_lock_acquired(dispatcher, lock):
+    """
+    Does given lock object has acquired the lock? It does this by getting
```
glyph: Grammar: this sentence is a mess 😄 .  I think you mean something like: "Is the given lock object currently acquired by this worker?"

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+
+    :return: `Deferred` of `bool`
+    """
+    return perform(dispatcher, is_lock_acquired_eff(lock))
+
+
+@do
+def is_lock_acquired_eff(lock):
+    """
+    Does given lock object has acquired the lock?
```
glyph: See above.

otter/convergence/selfheal.py
```diff
@@ -0,0 +1,193 @@
...
+
+    :return: `Effect` of `bool`
+    """
+    children = yield Effect(GetChildren(lock.path))
+    if not children:
+        yield do_return(False)
+    # The last 10 characters are sequence number as per
+    # https://zookeeper.apache.org/doc/current/zookeeperProgrammers.html\
```
glyph: You should not need to break lines with URLs in them this way (especially since this isn't really a valid way to break a URL up ;)).  I specifically reported a bug on pep8 to ensure that it would skip single-token long lines like this.

glyph: (See https://github.com/PyCQA/pycodestyle/issues/326 and https://github.com/PyCQA/pycodestyle/issues/224 )

otter/convergence/service.py
```diff
@@ -1,7 +1,7 @@
 """
 Converger service
 
-The top-level entry-points into this module are :obj:`ConvergenceStarter` and
+The top-level entry-points into this module are :func:`trigger_convergence` and
```
glyph: Niiiiice :)

otter/test/convergence/test_selfheal.py
```diff
@@ -0,0 +1,328 @@
...
+
+class SelfHealTests(SynchronousTestCase):
+    """
+    Tests for :obj:`SelfHeal`
+    """
+
+    def setUp(self):
+        self.lock = mock.Mock(specs=["acquire", "release"])
+        self.lock.release.return_value = succeed("released")
+        self.kzc = mock.Mock(specs=["Lock"])
```
glyph: General tech-debt note: a functional test-mock for txkazoo would be very helpful in ensuring that tests are doing the Right Thing when it comes to the distributed lock semantics, instead of having each test spin up its own Mocks and hope that they do the right thing...

glyph:
I added a few more minor notes.  I'm still pretty unhappy about the over-use of Mock in the test suite, but that is not a new problem and I would not want to block this on you having to go back and build out a complete set of verified fakes :).  Plus the mocks mostly seem to cover the intent.  So I think that this is OK to land - address the feedback that I gave to your satisfaction, and file some issues to track the general Mock tech-debt for future refactoring...

otter/convergence/selfheal.py
manishtomar: Where is the valid ReST syntax?? I am never able to find authoritative documentation of ReST. Maybe I am just too stupid to search with right keywords. Sorry just ranting out here :) Nothing with related to above comment.

glyph: This part of it is documented here: http://www.sphinx-doc.org/en/stable/domains.html#python-signatures

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/build-90dc6bdcce/docs/autoscale/v1/developer-guide/)

manishtomar:

Issue #1870, opened 2016-04-07
CLB node gets added before application in server is ready


manishtomar:
When a server becomes ACTIVE, it is immediately added to CLB. During this time, the application in server that is serving the port is mostly not yet ready and this results in various errors like 503s or 404s or "welcome to apache" depending on what situation the web server is in. This can be avoided by doing the following:

1. Add a node with a “DRAINING” condition. It should be pulled into rotation except that no new connections can be established.
2. Wait for the health checks to complete (depends on the health check parameters) at which point the status of the node will either be “ONLINE” or “OFFLINE”.
3. Change condition to “ENABLED”. If status is “ONLINE” the node will start to receive traffic. If status is “OFFLINE” the node should not serve traffic. If status is “OFFLINE” health check will eventually change status to “ONLINE” once it deems the node healthy. 

Following are possible changes to be done in convergence:

1. When server's node is not found, create new node in `DRAINING` if health check is enabled on the CLB. If not, add it as `ENABLED`.
2. When it notices that a node exists but is in `DRAINING`, it will check whether corresponding server's [`DESTINY`](https://github.com/rackerlabs/otter/blob/21cd1b150e67923a6b3db2d78f22d94565fa4dfd/otter/convergence/planning.py#L311-L316) is to drain. If so, it will continue to drain as usual, otherwise it will assume that it is new node that needs to be `ENABLE`ed. It will `ENABLE` the node if node's status is `ONLINE`. Otherwise, return RETRY. It is possible that node may never go `ONLINE` in which case we should put group in ERROR if it remains OFFLINE for > timeout (1 hour). This can be done by checking node's created time with current time. Created time can be got from node feed.

One of the problems with above changes is that it is CLB specific and the planner is generic using `ILBDescription` and `ILBNode` interfaces. Granted that all this will be done only if `ILBNode` implements `IDrainable` which currently is only CLB but is a problem if another LB type (RCv3) implements `IDrainable`. We could try to structure this part in seperate CLB specific functions.

manishtomar:
@glyph Not sure if you are following the as-dev mails but would like your opinion on above.

glyph:
I love the naming for `Destiny`, first time encountering that :).

The "health check" here is the LB's ["health monitor"](https://developer.rackspace.com/docs/cloud-load-balancers/v1/developer-guide/#show-health-monitor-configuration), yes?

It took a while for me to completely understand this, so I'm curious as to how you're going to document it.  Would something like this be accurate?

[...]

Issue #1443, opened 2015-05-12
Walk backwards to find draining time for CLB nodes


radix:
From gathering.py extract_CLB_drained_at:

TODO: This function temporarily only looks at last entry assuming that it was draining operation. May need to look at all entries in reverse order and check for draining operation. This could include paging to further entries

We're only looking at the last event in a CLB feed. We need to walk the events backwards until we find the relevant one, since other events may happen on CLB nodes.

manishtomar:
Closing as it has been fixed in #1443.

PR #1890, opened 2016-05-12
Getting CLB node draining and creation time

manishtomar:
Beginning of #1870.

- `extract_CLB_drained_at` now returns time when the node was updated with DRAINING or when it was created with DRAINING since technically the node was drained at the time it was created for a new node. This will be used to get new node's creation time for #1870.
- `get_clb_node_feed` now fetches more atom's next link until there are no entries effectively fixing #1443. This required changes in `atom` to parse `next_link`

### DIFF
--- a/otter/cloud_client/__init__.py
+++ b/otter/cloud_client/__init__.py
@@ -15,6 +15,7 @@ from effect import (
     catch,
     perform,
     sync_performer)
+from effect.do import do
 
 import six
 
@@ -705,21 +706,34 @@ def get_clbs():
         success=lambda (response, body): body['loadBalancers'])
 
 
-def get_clb_node_feed(lb_id, node_id):
-    """Get the atom feed associated with a CLB node. Returns feed as str."""
+def _node_feed_request(lb_id, node_id, params):
     return service_request(
-        ServiceType.CLOUD_LOAD_BALANCERS,
-        'GET',
+        ServiceType.CLOUD_LOAD_BALANCERS, 'GET',
         append_segments('loadbalancers', str(lb_id), 'nodes',
                         '{}.atom'.format(node_id)),
+        params=params,
         json_response=False
     ).on(
         error=_only_json_api_errors(
             lambda c, b: _process_clb_api_error(c, b, lb_id))
     ).on(
         log_success_response('request-get-clb-node-feed', identity)
-    ).on(
-        success=lambda (response, body): body)
+    )
+
+
+@do
+def get_clb_node_feed(lb_id, node_id):
+    """Get the atom feed associated with a CLB node. Returns feed as str."""
+    all_entries = []
+    params = {}
+    while True:
+        feed = yield _node_feed_request(lb_id, params)
+        entries = atom.entries(feed)
+        if entries == []:
+            yield do_return(all_entries)
+        all_entries.extend(entries)
+        next_link = atom.next_link(atom.parse(feed))
+        params = parse_qs(urlparse(next_link).query)
 
 
 def _expand_clb_matches(matches_tuples, lb_id, node_id=None):
--- a/otter/cloud_client/__init__.py
+++ b/otter/cloud_client/__init__.py
@@ -2,6 +2,7 @@
 import json
 import re
 from functools import partial, wraps
+from operator import itemgetter
 from urlparse import parse_qs, urlparse
 
 import attr
@@ -15,7 +16,7 @@ from effect import (
     catch,
     perform,
     sync_performer)
-from effect.do import do
+from effect.do import do, do_return
 
 import six
 
@@ -29,6 +30,7 @@ from txeffect import deferred_performer, perform as twisted_perform
 
 from otter.auth import Authenticate, InvalidateToken, public_endpoint_url
 from otter.constants import ServiceType
+from otter.indexer import atom
 from otter.log.intents import msg as msg_effect
 from otter.util.config import config_value
 from otter.util.http import APIError, append_segments, try_json_with_keys
@@ -718,22 +720,30 @@ def _node_feed_request(lb_id, node_id, params):
             lambda c, b: _process_clb_api_error(c, b, lb_id))
     ).on(
         log_success_response('request-get-clb-node-feed', identity)
-    )
+    ).on(itemgetter(1))
 
 
 @do
 def get_clb_node_feed(lb_id, node_id):
-    """Get the atom feed associated with a CLB node. Returns feed as str."""
+    """
+    Get the atom feed associated with a CLB node.
+
+    :param int lb_id: Cloud Load balancer ID
+    :param int node_id: Node ID of in loadbalancer node
+
+    :returns: ``list`` of atom entry :class:`Element`
+    :rtype: ``list``
+    """
     all_entries = []
     params = {}
     while True:
-        feed = yield _node_feed_request(lb_id, params)
+        feed_str = yield _node_feed_request(lb_id, node_id, params)
+        feed = atom.parse(feed_str)
         entries = atom.entries(feed)
         if entries == []:
             yield do_return(all_entries)
         all_entries.extend(entries)
-        next_link = atom.next_link(atom.parse(feed))
-        params = parse_qs(urlparse(next_link).query)
+        params = parse_qs(urlparse(atom.next_link(feed)).query)
 
 
 def _expand_clb_matches(matches_tuples, lb_id, node_id=None):
--- a/otter/cloud_client/__init__.py
+++ b/otter/cloud_client/__init__.py
@@ -708,7 +708,17 @@ def get_clbs():
         success=lambda (response, body): body['loadBalancers'])
 
 
-def _node_feed_request(lb_id, node_id, params):
+def _node_feed_page(lb_id, node_id, params):
+    """
+    Return page of CLB node feed
+
+    :param int lb_id: Cloud Load balancer ID
+    :param int node_id: Node ID of in loadbalancer node
+    :param dict params: Request query parameters
+
+    :returns: Unparsed response body as string
+    :rtype: `str`
+    """
     return service_request(
         ServiceType.CLOUD_LOAD_BALANCERS, 'GET',
         append_segments('loadbalancers', str(lb_id), 'nodes',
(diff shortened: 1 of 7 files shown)


rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/build-ccc8bf3e05/docs/autoscale/v1/developer-guide/). :bow:

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/build-1e5fa48cb0/docs/autoscale/v1/developer-guide/)

manishtomar:
Jenkins please run integration tests

manishtomar:

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/build-089ff072e7/docs/autoscale/v1/developer-guide/)

manishtomar:

manishtomar:
@glyph review please.

otter/cloud_client/__init__.py
```diff
@@ -705,21 +708,45 @@ def get_clbs():
         success=lambda (response, body): body['loadBalancers'])
 
 
-def get_clb_node_feed(lb_id, node_id):
-    """Get the atom feed associated with a CLB node. Returns feed as str."""
+def _node_feed_request(lb_id, node_id, params):
```
glyph: A docstring would be nice here :)

otter/cloud_client/__init__.py
```diff
@@ -705,21 +708,45 @@ def get_clbs():
...
+
+@do
+def get_clb_node_feed(lb_id, node_id):
+    """
+    Get the atom feed associated with a CLB node.
+
+    :param int lb_id: Cloud Load balancer ID
+    :param int node_id: Node ID of in loadbalancer node
+
+    :returns: ``list`` of atom entry :class:`Element`
```
glyph: Is this accurate?  Doesn't a `@do`-decorated function always return an `Effect`?  Later, the effect will _resolve_ to some `Element` instances, but the distinction in the call signature is very important.

otter/cloud_client/__init__.py
```diff
@@ -705,21 +708,45 @@ def get_clbs():
...
+    :rtype: ``list``
+    """
+    all_entries = []
+    params = {}
+    while True:
+        feed_str = yield _node_feed_request(lb_id, node_id, params)
+        feed = atom.parse(feed_str)
+        entries = atom.entries(feed)
+        if entries == []:
+            yield do_return(all_entries)
```
glyph: A minor stylistic nitpick: I find it really hard to read a loop where it's `while True` and then no actual exiting elements except special exceptions; `do_return` here is a secret non-control-flow abstraction telling you to exit the loop via a return (without saying `return`).

Luckily, in this case, fixing the loop to have visible exits also eliminates some duplication: what do you think of replacing both `yield do_return(all_entries)` lines with a `break` statement, and putting `yield do_return(all_entries)` at the end of the function, outside the `while` loop?

otter/cloud_client/__init__.py
```diff
@@ -705,21 +708,45 @@ def get_clbs():
...
+        all_entries.extend(entries)
+        next_link = atom.next_link(feed)
+        if not next_link:
+            yield do_return(all_entries)
+        params = parse_qs(urlparse(next_link).query)
```
glyph: Technically speaking, this is subtly incorrect.  In a REST protocol like Atom, the "next link" may be any valid HTTPS URI; you shouldn't be parsing it or trying to understand it here, just issuing a `GET` to the exact URL that you found in the feed.  For example, it's valid for a feed to say `/foo/bar?page=2` or `/foo/bar/page/2` or even `https://example.com/e8e0e8ce795e3bee11f0`.  Nothing says that it's just the query string that changes.

I think this is fine for now given the way that cloud feeds' API works and is documented, but would you mind at least documenting the technical debt here with an issue?

otter/convergence/gathering.py
```diff
@@ -196,25 +197,31 @@ def update_drained_at(node):
...
+    "^Node successfully created with address: '.+', port: '\d+', "
+    "condition: 'DRAINING', weight: '\d+'$")
+_DRAINING_UPDATED_RE = (
+    "^Node successfully updated with address: '.+', port: '\d+', "
+    "weight: '\d+', condition: 'DRAINING'$")
+_DRAINING_RE = re.compile(
+    "({})|({})".format(_DRAINING_UPDATED_RE, _DRAINING_CREATED_RE))
+
+
 def extract_CLB_drained_at(feed):
```
glyph: I know this isn't new, but "CLB" shouldn't be capitalized here.

otter/convergence/gathering.py
```diff
@@ -196,25 +197,31 @@ def update_drained_at(node):
     yield do_return(list(filter(bool, nodes)))
 
 
+_DRAINING_CREATED_RE = (
```
glyph: Seriously? :(

glyph: To be clear: it's just sad that this is the only way to do this.  I wonder if it would be appropriate to open a ticket on CLBs to get this information in a more structured way, sans regexes (if you haven't already).

otter/test/cloud_client/test_init.py
```diff
@@ -600,6 +602,94 @@ def test_perform_srvreq_nested(self):
              self.throttler, 1, ereq.intent))
 
 
+def assert_parses_common_clb_errors(testcase, intent, eff, lb_id="123456"):
+    """
+    Assert that the effect produced performs the common CLB error parsing:
+    :class:`CLBImmutableError`, :class:`CLBDescription`,
+    :class:`NoSuchCLBError`, :class:`CLBRateLimitError`,
+    :class:`APIError`
```
glyph: Could you add some `:param:` values here?

otter/test/cloud_client/test_init.py
```diff
@@ -600,6 +602,94 @@ def test_perform_srvreq_nested(self):
              self.throttler, 1, ereq.intent))
 
 
+def assert_parses_common_clb_errors(testcase, intent, eff, lb_id="123456"):
```
glyph: Why is there a default `lb_id`?

otter/test/convergence/test_gathering.py
```diff
@@ -320,31 +321,54 @@ class ExtractDrainedTests(SynchronousTestCase):
...
-        '</feed>')
+    updated_summary = ("Node successfully updated with address: "
+                       "'10.23.45.6', port: '8080', weight: '1', "
+                       "condition: 'DRAINING'")
+    created_summary = ("Node successfully created with address: "
+                       "'10.23.45.6', port: '8080', condition: 'DRAINING', "
+                       "weight: '1'")
+    updated1 = '2014-10-23T18:10:48.001Z'
+    updated2 = '2015-09-23T08:00:40Z'
+    entry = '<entry><summary>{}</summary><updated>{}</updated></entry>'
```
glyph: FWIW: `twisted.web.template` can produce valid XML if you find yourself needing to do a lot of this.

otter/test/convergence/test_gathering.py
```diff
@@ -381,9 +405,19 @@ def nodes_req(lb_id, nodes):
 
 
 def node_feed_req(lb_id, node_id, response):
-    return lb_req(
-        'loadbalancers/{}/nodes/{}.atom'.format(lb_id, node_id),
-        False, response)
+    if isinstance(response, Exception):
```
glyph: It looks like the type of the parameters got more important here - can you add a docstring?

glyph:
Mostly stylistic nitpicking aside, this looks pretty good to me.  Address to taste and land. 👍

otter/cloud_client/__init__.py
manishtomar: Sounds good. Coming from C background this is the only time I miss goto statement :). Break is a good compromise.

otter/cloud_client/__init__.py
manishtomar: I agree with you completely. However, this function works/returns with `Effect of ServiceRequest` which is tied to a specific service whose URI gets expanded during performing. It will get complicated to support any URI _with_ authentication. 

I had similar confusion when implementing getting all nova servers. But settled with a similar solution. But you are right, this should be documented as a tech-debt. Will do. Thanks!

otter/cloud_client/__init__.py
manishtomar: > Is this accurate?

No.

> Doesn't a @do-decorated function always return an Effect?

Yes.

Thanks for pointing this out. I start thinking in terms of returning value wrapped in the Effect when I am working with a lot of effect code which results in above docstring similar to how it happens when working with a lot of `Deferred` code :). I'll change it.

otter/convergence/gathering.py
manishtomar: > wonder if it would be appropriate to open a ticket on CLBs to get this information in a more structured way, sans regexes (if you haven't already).

It would be appropriate. However, I really doubt they would even look at it considering the fact that CLB is EOLing (for past 2 years 😉). Moreover, this is an unintentional API that according to them no one uses :). I think I'll create a ticket anyway just so that they are aware about it.

otter/test/cloud_client/test_init.py
manishtomar: This function is factored out from `CLBClientTests` so that it can be used in newer `GetCLBNodeFeedTests`. It had much more calls from `CLBClientTests` which used "123456" as lb_id as compared to one call from `GetCLBNodeFeedTests` which used different lb_id. Hence, the default value. But this sounds like a stupid reason to have default value especially since this function is outside both the classes. I'll remove it. Thanks for catching this :)

otter/test/convergence/test_gathering.py
manishtomar: Will do!

manishtomar: Will do!

otter/test/cloud_client/test_init.py
manishtomar: This function is factored out from `CLBClientTests` so that it can be used in newer `GetCLBNodeFeedTests`. It had much more calls from `CLBClientTests` which used "123456" as lb_id as compared to one call from `GetCLBNodeFeedTests` which used different lb_id. Hence, the default value. But this sounds like a stupid reason to have default value especially since this function is outside both the classes. I'll remove it. Thanks for catching this :)

otter/cloud_client/__init__.py
glyph: The terminology in `Deferred` land is `Deferred firing with...` or `Deferred failing with...` - so it should be getting captured by your documented return type :)

otter/convergence/gathering.py
glyph: Well, CLB V2 lives on, yes? So it would be good for them to know :)

PR #1892, opened 2016-06-07
Group becomes ACTIVE in beginning of convergence cycle

### LINKED ISSUE #1809: ERROR Group should become ACTIVE at beginning of convergence
When user finds ERROR group with information about invalid launch config, he/she will fix the launch config and call `converge`. At that time, the group status is still shown as ERROR and is updated to ACTIVE only after convergence succeeds which can be multiple cycles resulting in a update potentially after 1-2 minutes depending on the group and number of resources created/deleted. During that time the group is technically _active_ and working to bring desired state but shows ERROR. This is incorrect and results in user assuming that `converge` didn't work. The group should move to ACTIVE at beginning of convergence cycle. This way it is accurate with its definition.

manishtomar:
Closes #1809

### DIFF
--- a/otter/models/intents.py
+++ b/otter/models/intents.py
@@ -78,11 +78,11 @@ def perform_update_group_status(dispatcher, ugs_intent):
 
 
 @attr.s
-class UpdateScalingGroupStatus(object):
+class LoadAndUpdateGroupStatus(object):
     """
-    Intent to update scaling group status. This is different from
-    :obj:`UpdateGroupStatus` by taking tenant_id and group_id instead
-    of group object
+    Intent to load the scaling group object and update its status. This is
+    different from :obj:`UpdateGroupStatus` by taking tenant_id and group_id
+    instead of group object
     """
     tenant_id = attr.ib()
     group_id = attr.ib()
@@ -90,8 +90,8 @@ class UpdateScalingGroupStatus(object):
 
 
 @deferred_performer
-def perform_update_scaling_group_status(log, store, dispatcher, intent):
-    """Perform an :obj:`UpdateScalingGroupStatus`."""
+def perform_load_and_update_group_status(log, store, dispatcher, intent):
+    """Perform an :obj:`LoadAndUpdateGroupStatus`."""
     group = store.get_scaling_group(log, intent.tenant_id, intent.group_id)
     return group.update_status(intent.status)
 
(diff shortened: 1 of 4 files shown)


rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/build-3f7004b413/docs/autoscale/v1/developer-guide/). :bow:

manishtomar:

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/build-6881c9610e/docs/autoscale/v1/developer-guide/)

manishtomar:

manishtomar:
@glyph review please.

otter/models/intents.py
```diff
@@ -78,6 +78,25 @@ def perform_update_group_status(dispatcher, ugs_intent):
...
+    of group object
+    """
+    tenant_id = attr.ib()
+    group_id = attr.ib()
+    status = attr.ib()
+
+
+@deferred_performer
+def perform_update_scaling_group_status(log, store, dispatcher, intent):
+    """Perform an :obj:`UpdateScalingGroupStatus`."""
```
glyph: Can you add some `:param`s here?  https://txeffect.readthedocs.io/en/latest/ is not much help in deciphering the expected signature of a `@deferred_performer`, unfortunately.

otter/test/convergence/test_service.py
```diff
@@ -857,7 +858,7 @@ def setUp(self):
         self.now = datetime(1970, 1, 1)
         self.waiting = Reference(pmap())
 
-    def get_seq(self, with_cache=True):
+    def get_seq(self, with_cache=True, upd_status_nogroup_err=False):
```
glyph: That parameter's role is far from obvious to the casual reader: this needs a docstring :)

otter/models/intents.py
```diff
@@ -78,6 +78,25 @@ def perform_update_group_status(dispatcher, ugs_intent):
 
 
 @attr.s
+class UpdateScalingGroupStatus(object):
```
glyph: This class name confused me at first.  The salient property here is not "scaling", as "group" and "scaling group" are synonyms, but rather the fact that one is *loaded already* and one isn't.  Do you think a name-change here to `LoadAndUpdateGroup` would make sense?  Given that there's an "and" in the name, is there any way to more gracefully compose the group-loading and existing group-updating effects?  (If not, that's fine, but if we could get away with fewer effects that need distinct performers, that would be great.)

glyph:
All in all this one looks like a slam dunk to me - I'd like more docs in a few places (as noted in inline comments), and I think the naming is unduly confusing for new maintainers -  but overall it seems pretty clear what is going on.  After addressing please land.

otter/models/intents.py
manishtomar: Perfect! Thanks for coming out with much better name! Will change it.

otter/models/intents.py
manishtomar: Hmm. This is not good as lot of code has limited documentation. It leads me to believe that we need much more documentation for effect related code :-|

otter/test/convergence/test_service.py
manishtomar: Oh it needs more than a docstring. Its one of those messy tests that I would like to rewrite. However, I'll add a docstring to explain.

PR #1906, opened 2016-06-30
Add node as draining

### LINKED ISSUE #1870: CLB node gets added before application in server is ready

manishtomar:
The final piece of #1870. Closes #1870. 

CLB health info is stored in a separate object and passed to planner which uses the info to decide whether to add the node as DRAINING.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -16,6 +16,7 @@ from otter.cloud_client import (
     get_clb_node_feed,
     get_clb_nodes,
     get_clbs,
+    is_clb_health_monitor_enabled,
     list_servers_details_all,
     list_stacks_all,
     service_request)
@@ -170,9 +171,16 @@ def get_clb_contents():
     lb_ids = [lb['id'] for lb in (yield _retry(get_clbs()))]
     node_reqs = [_retry(get_clb_nodes(lb_id).on(error=gone([])))
                  for lb_id in lb_ids]
-    all_nodes = yield parallel(node_reqs)
-    lb_nodes = {lb_id: [CLBNode.from_node_json(lb_id, node) for node in nodes]
-                for lb_id, nodes in zip(lb_ids, all_nodes)}
+    healthmon_reqs = [
+        _retry(is_clb_health_monitor_enabled(lb_id).on(error=gone(False)))
+        for lb_id in lb_ids]
+
+    all_nodes_hms = yield parallel(node_reqs + healthmon_reqs)
+    all_nodes, hms = all_nodes_hms[:len(lb_ids)], all_nodes_hms[len(lb_ids):]
+    lb_nodes = {
+        lb_id: [CLBNode.from_node_json(lb_id, node, health_mon)
+                for node in nodes]
+        for lb_id, nodes, health_mon in zip(lb_ids, all_nodes, hms)}
     draining = [n for n in concat(lb_nodes.values())
                 if n.description.condition == CLBNodeCondition.DRAINING]
     feeds = yield parallel(
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -22,6 +22,7 @@ from otter.cloud_client import (
     service_request)
 from otter.constants import ServiceType
 from otter.convergence.model import (
+    CLB,
     CLBNode,
     CLBNodeCondition,
     HeatStack,
@@ -168,19 +169,22 @@ def get_clb_contents():
 
     def gone(r):
         return catch(CLBNotFoundError, lambda exc: r)
+
     lb_ids = [lb['id'] for lb in (yield _retry(get_clbs()))]
     node_reqs = [_retry(get_clb_nodes(lb_id).on(error=gone([])))
                  for lb_id in lb_ids]
     healthmon_reqs = [
-        _retry(get_clb_health_monitor(lb_id).on(error=gone(False)))
+        _retry(get_clb_health_monitor(lb_id).on(error=gone(None)))
         for lb_id in lb_ids]
-
     all_nodes_hms = yield parallel(node_reqs + healthmon_reqs)
     all_nodes, hms = all_nodes_hms[:len(lb_ids)], all_nodes_hms[len(lb_ids):]
     lb_nodes = {
-        lb_id: [CLBNode.from_node_json(lb_id, node, bool(health_mon))
+        lb_id: [CLBNode.from_node_json(lb_id, node)
                 for node in nodes]
-        for lb_id, nodes, health_mon in zip(lb_ids, all_nodes, hms)}
+        for lb_id, nodes in zip(lb_ids, all_nodes)}
+    clbs = {
+        str(lb_id): CLB(bool(health_mon))
+        for lb_id, health_mon in zip(lb_ids, hms) if health_mon is not None}
     draining = [n for n in concat(lb_nodes.values())
                 if n.description.condition == CLBNodeCondition.DRAINING]
     feeds = yield parallel(
@@ -202,7 +206,9 @@ def get_clb_contents():
         else:
             return node
     nodes = map(update_drained_at, concat(lb_nodes.values()))
-    yield do_return(list(filter(bool, nodes)))
+    yield do_return((
+        list(filter(bool, nodes)),
+        keyfilter(lambda k: k not in deleted_lbs, clbs)))
 
 
 _DRAINING_CREATED_RE = (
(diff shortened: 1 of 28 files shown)


manishtomar:

manishtomar:

manishtomar:

manishtomar:
@glyph review please.

otter/convergence/gathering.py
```diff
@@ -273,9 +287,10 @@ def get_all_launch_server_data(
          .on(map(NovaServer.from_server_details_json)).on(list),
          get_clb_contents(),
          get_rcv3_contents()]
-    ).on(lambda (servers, clb, rcv3): {
+    ).on(lambda (servers, (clb_nodes, clbs), rcv3_nodes): {
```
glyph: FYI: argument-list tuple-unpacking is now illegal in Python 3.

```
>>> lambda a, (b, c): None
  File "<input>", line 1
    lambda a, (b, c): None
              ^
SyntaxError: invalid syntax
```

otter/convergence/model.py
```diff
@@ -563,11 +579,22 @@ def equivalent_definition(self, other_description):
...
+    # Is health monitor enabled?
+    health_monitor = attr.ib(default=False)
+
+
 @implementer(ILBNode, IDrainable)
 @attributes([Attribute("node_id", instance_of=basestring),
              Attribute("description", instance_of=CLBDescription),
              Attribute("address", instance_of=basestring),
              Attribute("drained_at", default_value=0.0, instance_of=float),
+             Attribute("status", instance_of=NamedConstant,
```
glyph: Question: are there any other possible future values here for node status?  My inclination would be to just call this `is_online` and make it a boolean if "online" and "offline" are really the only two; having a whole enumerated `NamedConstant` seems a bit precious if it's really just a boolean.

otter/convergence/model.py
```diff
@@ -622,6 +653,7 @@ def from_node_json(cls, lb_id, json):
         return cls(
             node_id=str(json['id']),
             address=json['address'],
+            status=CLBNodeStatus.lookupByName(json["status"]),
```
glyph: Question: Is this already expressed in a declarative schema somewhere?

otter/convergence/model.py
manishtomar: Can't tell about the future but the [documentation](https://developer.rackspace.com/docs/cloud-load-balancers/v1/api-reference/nodes/) only mentions `OFFLINE` and `ONLINE` as of now. I think it is a good idea to have boolean attribute `online` except that future reader will need to know that it is correlated with `status` field.

otter/convergence/planning.py
```diff
@@ -587,24 +619,54 @@ def remove_node_from_lb(node):
...
 
-    :ivar node: The node to be changed.
+    :param node: The node to be changed.
     :type node: :class:`ILBNode` provider
-
-    :ivar description: The description of the load balancer and how to add
+    :param description: The description of the load balancer and how to add
         the server to it.
     :type description: :class:`ILBDescription` provider
+    :param float now: Current time in EPOCH
```
glyph: Minor nitpicks about terminology:

- "EPOCH" is not an acronym; it should just be "Epoch"
- It's the current time in "seconds _since_ [the Epoch](https://en.wikipedia.org/wiki/Epoch_(reference_date)#Notable_epoch_dates_in_computing)"

otter/convergence/planning.py
```diff
@@ -587,24 +619,54 @@ def remove_node_from_lb(node):
...
+    :param float timeout: How long can node remain OFFLINE after adding
+        in seconds?
+
+    :return: :obj:`IStep` object or None
```
glyph: Just a thought for the future - would a "do-nothing step" object make more sense than None here?

otter/convergence/service.py
```diff
@@ -180,13 +180,15 @@ def all_met(server, current_lb_nodes):
                              if node.matches(server)]))
 
 
-def update_servers_cache(group, now, servers, lb_nodes, include_deleted=True):
+def update_servers_cache(group, now, servers, lb_nodes, lbs,
```
glyph: Eventually I figured out how `update_servers_cache` managed to change its signature without `update_servers_cache` showing up anywhere in the diff, so I don't actually have feedback about this change.  However, I will take this opportunity to say that this is kind of pointlessly confusing, and `ConvergenceExecutor` really wants to be an interface, and `launch_server_executor` and `launch_stack_executor` want to be providers of it.  This might be a place where `@provider` might be more appropriate than `@implementer` because they're guaranteed to be stateless, but nevertheless, idiomatic convention in Python is to group related functions in a namespace by making classes or modules, not making "bags" of functions that require said functions to be defined under misleadingly distinct names :)

otter/test/convergence/test_gathering.py
```diff
@@ -404,6 +405,11 @@ def nodes_req(lb_id, nodes):
                   True, {'nodes': nodes})
 
 
+def lb_hm_req(lb_id, health_mon):
```
glyph: docstring please - it's far from obvious what this is :)

otter/convergence/service.py
```diff
@@ -180,13 +180,15 @@ def all_met(server, current_lb_nodes):
...
+                         include_deleted=True):
     """
     Updates the cache, adding servers, with a flag if autoscale is active on
     each one.
     :param group: scaling group
     :param list servers: list of NovaServer objects
     :param list lb_nodes: list of CLBNode objects
+    :param dict lbs: load balancer objects keyed on ID (currently ignored)
```
glyph: I was also a little bit confused by "currently ignored" since this is supposed to be the _last_ bit of #1870.  So let me explain what I discovered to confirm my understanding:

`convergence_exec_data` gathers "resources", which are passed as keyword arguments to `executor.update_cache`.  For servers, `update_cache` is `get_all_launch_server_data`, which now returns a `lbs` key; this means that `convergence_exec_data` must accept this argument.  But `get_all_launch_server_data` is returning this for the benefit of `launch_server_executor.plan`, by which I mean of course `plan_launch_server`, which now requires `lbs` to do _its_ job.

For this change, I think the only thing to do is to note where `lbs` is actually coming from and why it's unused (since as far as I can tell the plan is not to use it in the future) in the docstring here.  In the future, I hope we can do a little refactoring to make it more structurally obvious how data moves around here.

glyph:
With some close reading of the changes, I think I understand pretty clearly what's going on here.

glyph:
@manishtomar - Most of my feedback is just general comments on the current architecture of `ConvergenceExecutor`, necessary mainly for me to come up to speed on this code.  This seems like a clean, isolated change, and a shining example of how to break up a large, complex change into comprehensible individual changes that each make forward progress.  Great work!  Please land after you've addressed to your own satisfaction, although if you make any major changes and want me to take another look I'm happy to.

otter/convergence/planning.py
manishtomar: "Active Nothing" pattern? :) Sounds good.

manishtomar:
Tested in prod and addressed most of the comments. Will deploy on Monday.

manishtomar:

PR #1919, opened 2016-07-23
node.drained_at is 0 instead of None

### LINKED ISSUE #1863: CLB drained_at error recovery
Right now when something is not right when extracting draining time otter errors and tries to converge foreover. It should instead set `node.drained_at` to `None` and continue converging. If planner finds out that it needs this info and is not there, can ERROR the group. 

Note that this is slightly related to #1443 where that change should also set `node.drained_at=None` if keeps going back forever.

manishtomar:
when it cannot find draining info. This fixes #1863 because node will look to be OFFLINE for a very long time resulting in ERRORing the group. The error log might be confusing though.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -224,7 +224,7 @@ _DRAINING_RE = re.compile(
 def extract_clb_drained_at(feed):
     """
     Extract time when node was changed to DRAINING from a CLB atom feed. Will
-    return node's creation time if node was created with DRAINING. Return None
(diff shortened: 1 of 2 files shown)


manishtomar:

manishtomar:
This will really confuse any future maintainer (or future me) when a group ERRORs with very large node OFFLINE time. Moreover it will require unnecessary digging to find out why such an error occurred; it occurred because gathering couldn't find `node.drained_at`. The right way to fix this is to not be lazy and write proper error reason.

PR #1920, opened 2016-07-26
CLB objects returned as pmap instead of dict

manishtomar:
primarily because JSON output of `execute-convergence` log contains `lbs` with `CLB` objects which gets indexed in ES as `lbs.lbid`. This is unnecessary and complicates ES index.

Also good idea to use pmap instead of dict due to immutability.


manishtomar:

Issue #1915, opened 2016-07-18
LockTimeouts :(


manishtomar:
One tenant is constantly getting locktimeouts when calling webhook 3 times simultaneously. Need to debug and see where is the problem in Kazoo. I've been failing to reproduce this on my dev env.

manishtomar:
After extensive debugging, I found at least one problem. Lock release deletes child ephemeral znode which triggers watch event on the previous caller who is waiting to acquire the lock. On receiving the event, that caller must wake up and try the acquiring process again: i.e. get the children and see if's created child is the smallest. THIS does not happen. After receiving the watch event the caller wakes up after some extra time (in my case 22 seconds!) to check if it has the lock. 

[This](https://github.com/python-zk/kazoo/blob/2.2.1/kazoo/recipe/lock.py#L236) is where the watch is set and next line should unblock on getting the event continuing the while loop which doesn't happen. 

I am not sure if this could be due to twisted threadpool. @glyph What do you think? Any remote chances? Please note that `self._watch_predecessor` is called from events callback thread that is created [here](https://github.com/python-zk/kazoo/blob/2.2.1/kazoo/handlers/threading.py#L116-L120) in beginning of the application.

glyph:
I've been looking at this ticket for the last few days without coming to any specific conclusion.  This does indeed look complex.

The only contribution I could think of that the Twisted threadpool *might* have is that it sets a fairly conservative maximum number of threads.  In some cases this can provoke deadlocks.  For example:

Let's say you have a thread-pool with a limit of 4 threads maximum.  Some other part of your application starts 2 long-running tasks.  Your code, knowing that the limit is 4, starts 3 parallel tasks which all wait on a lock.  Then it fires up a 4th task to release the lock.  That 4th task will wait in line forever, never being run; your threadpool is now wedged forever.

I think that having a fixed maximum size for a threadpool is actually a design error.  This is a ticket in Twisted for addressing this problem: https://twistedmatrix.com/trac/ticket/5298

manishtomar:
Not sure if it helps but I've seen the number of threads be much less than max threads in production. Max is set to 110 and I've seen sometimes 23 sometimes 53 threads when this problem is occurring.

glyph:
Hmm.  OK, that seems unlikely then.  If the Twisted threadpool is not wedged on thread-creation, there are not a lot of other places it can go wrong…

glyph:
[...]

PR #1925, opened 2016-08-10
Zookeeper PollingLock recipe 

manishtomar:
Zookeeper lock recipe that polls lock's children instead of leaving watch on the previous child. This is to avoid https://github.com/rackerlabs/otter/issues/1915#issuecomment-237145764. This is not yet used. I'll put another PR that uses this recipe in `modify_state`. I've tested this on dev vm by removing local lock on worker and converger code but it needs to be tested with multiple otter instances. I'll test that and update the result. Meanwhile, will be nice to review this code.

### DIFF
--- a/otter/test/util/test_zk.py
+++ b/otter/test/util/test_zk.py
@@ -1,17 +1,26 @@
 """Tests for otter.util.zk"""
 
+import time
+import uuid
+
 from functools import partial
 
 from characteristic import attributes
 
-from effect import ComposedDispatcher, Effect, TypeDispatcher, sync_perform
+from effect import (
+    ComposedDispatcher, Constant, Delay, Effect, Func, TypeDispatcher,
+    sync_perform)
+from effect.testing import SequenceDispatcher, perform_sequence
 
-from kazoo.exceptions import BadVersionError, NoNodeError, NodeExistsError
+from kazoo.exceptions import (
+    BadVersionError, LockTimeout, NoNodeError, NodeExistsError,
+    SessionExpiredError)
 
 from twisted.internet.defer import fail, succeed
 from twisted.trial.unittest import SynchronousTestCase
 
-from otter.test.utils import test_dispatcher
+from otter.test.utils import const, conste, intent_func, noop, test_dispatcher
+from otter.util import zk
 from otter.util.zk import (
     AcquireLock, CreateOrSet, CreateOrSetLoopLimitReachedError,
     DeleteNode, GetChildren, GetChildrenWithStats,
@@ -36,13 +45,15 @@ class ZKCrudModel(object):
     """
     def __init__(self):
         self.nodes = {}
+        self.create_makepath = True
 
-    def create(self, path, content, makepath=False):
+    def create(self, path, value="", acl=None, ephemeral=False, sequence=False,
+               makepath=False):
         """Create a node."""
-        assert makepath is True, "makepath must be True"
+        assert makepath == self.create_makepath
         if path in self.nodes:
             return fail(NodeExistsError("{} already exists".format(path)))
-        self.nodes[path] = (content, 0)
+        self.nodes[path] = (value, 0)
         return succeed(path)
 
     def get(self, path):
@@ -246,7 +257,7 @@ class GetStatTests(SynchronousTestCase):
 
     def test_get_stat(self):
         """Returns the ZnodeStat when the node exists."""
-        self.model.create('/foo/bar', content='foo', makepath=True)
+        self.model.create('/foo/bar', value='foo', makepath=True)
         result = self._gs('/foo/bar')
         self.assertEqual(result, ZNodeStatStub(version=0))
 
@@ -270,6 +281,18 @@ class DeleteTests(SynchronousTestCase):
         self.assertEqual(result, 'delete return value')
 
 
+class CreateTests(SynchronousTestCase):
+    """Tests for :obj:`CreateNode`."""
+    def test_create(self):
+        model = ZKCrudModel()
+        model.create_makepath = False
+        eff = Effect(zk.CreateNode(path='/foo', value="v"))
+        dispatcher = get_zk_dispatcher(model)
+        result = sync_perform(dispatcher, eff)
+        self.assertEqual(model.nodes, {"/foo": ("v", 0)})
+        self.assertEqual(result, '/foo')
+
+
 class AcquireLockTests(SynchronousTestCase):
     """Tests for :obj:`AcquireLock`."""
     def test_success(self):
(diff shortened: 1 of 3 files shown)


manishtomar:
@glyph review please.

otter/test/util/test_zk.py
```diff
@@ -36,13 +45,15 @@ class ZKCrudModel(object):
...
     def __init__(self):
         self.nodes = {}
+        self.create_makepath = True
 
-    def create(self, path, content, makepath=False):
+    def create(self, path, value="", acl=None, ephemeral=False, sequence=False,
+               makepath=False):
         """Create a node."""
-        assert makepath is True, "makepath must be True"
+        assert makepath == self.create_makepath
```
glyph: What is the purpose of making the caller pass this argument?  `assert`s make me nervous, because they're basically untested error conditions :)

otter/test/util/test_zk.py
```diff
@@ -279,3 +302,280 @@ def test_success(self):
...
+             const("/testlock/prefix0000000000")),
+            (GetChildren("/testlock"), const(["prefix0000000000"]))
+        ]
+        self.assertTrue(
+            perform_sequence(seq, self.lock.acquire_eff(False, None)))
+
+    def test_acquire_blocking_success(self):
+        """
+        acquire_eff creates child, realizes its not the smallest. Tries again
+        every 0.01 seconds until it succeeds
```
glyph: 0.01 is a magic number here; it should really be a constant.

otter/util/zk.py
```diff
@@ -170,3 +199,113 @@ def get_zk_dispatcher(kz_client):
...
+    of leaving a watch on previous child. It's supposed to replace
+    :obj:`kazoo.recipe.lock.Lock` and has same signature and _similar_ behavior
+    for ``acquire`` and ``release`` methods.
+    """
+
+    def __init__(self, dispatcher, path, identifier=""):
+        self.dispatcher = dispatcher
+        self.path = path
+        self.identifier = identifier
+        self._node = None
```
glyph: Side effect?  Sigh, defect :).

This is a mostly beautiful and apparently correct implementation of the locking logic, which I *think* I understand.  Except for this piece.  `_node` is set when the lock is *going* to be acquired… but then it's cleared when the lock is released.  And then a bunch of tests set it directly in order to test things about that internal state, rather than driving it through the external interface.

I think the main thing that needs to be done here, for now, is to explicitly document the meaning and purpose of `_node`, since it's effectively part of the public interface, at least as far as the tests are concerned, and this is _very_ important for any future maintainer looking at this.

I struggled to think of how to deploy Active Nothing here, although I'm sure if we could figure it out the code would be the better for it.  It's almost like we need [something](https://github.com/glyph/automat/) that could track changes in state and automatically respond... :o)

glyph:
The one other broad comment that I have is that having separate acquire/release methods is an antipattern.  You want to use something like a contextmanager or a higher-order function to ensure that calls to acquire and release are always paired.  Does this already exist somewhere else in the codebase?  I realize that's probably a bit out of scope for this PR though.

Modulo a bit more documentation to help future maintainers clearly apprehend the purpose of the `_node` attribute, I think that this is good to land.  I wish I could understand it better - I need to educate myself more as to the specifics of Zookeeper - but from what I can read of the tests description it's a good improvement.  Looking forward to the follow-up PR :)

otter/util/zk.py
manishtomar: > Side effect? Sigh, defect :).

This is bare minimal internal state required to function. Please compare with [kazoo](https://github.com/python-zk/kazoo/blob/master/kazoo/recipe/lock.py#L74-L107) :)

> This is a mostly beautiful and apparently correct implementation of the locking logic

Thanks!

> which I think I understand

https://zookeeper.apache.org/doc/trunk/recipes.html#sc_recipes_Locks without the watch part.

> Except for this piece. _node is set when the lock is going to be acquired…

I am sorry this is confusing. I am assuming you mean "node is released when locking is going to be acquired". This is done to allow leader election behavior (long-held lock) being used in selfheal service along with immediate acquire/release lock behavior used when modifying state. The idea is one worker will acquire the lock and check if it remains acquired by calling `is_lock_acquired` at some interval. If it is not acquired which can happen if the session was re-established it will call `acquire` again. Now, what should the function do if it already finds `_node` in such a situation? I believe the sanest thing to do is delete it (by calling `release_eff`) before beginning to acquire the lock. This way that child node (if it still remains for whatever reason) is not hanging around. Child node remaining is the original problem this code is trying to solve. I'll add docs about this like you suggested. 

> And then a bunch of tests set it directly in order to test things about that internal state, rather than driving it through the external interface.

I could change the tests to drive through external interface but I believe that is being tested in `test_acquire_*` functions. I am trying to think if it is a bad idea to test internal state without driving it from public interface. Not sure.

> I struggled to think of how to deploy Active Nothing here, although I'm sure if we could figure it out the code would be the better for it.

Hmm. I don't know how that pattern will help here. Moreover, it looks you want to use it because you want to use it :) ? Did you have any reasons for this?

glyph: > Hmm. I don't know how that pattern will help here. Moreover, it looks you want to use it because you want to use it :) ? Did you have any reasons for this?

This talk is a good motivation for pretty much *always* using this pattern: https://www.youtube.com/watch?v=29MAL8pJImQ

If this code is being changed in the future, with a `None`-or-not-`None` variable, you have to be *careful* about how you touch it.  You have to account for both cases at every call site.  You have to carefully consider the significance of `None` and re-derive it if you're not quite sure.  (I had to think pretty hard about what states `None` reflected, since it actually reflects 2 out of 3.)

With an Active Nothing, instead, you can blindly trust that you can send the same set of messages to the same attribute at any time.  It might *do something different*, but you don't have to be *careful*.  This is why Automat is a possible alternative to Active Nothing - with Automat you can "blindly" touch variables because the state-machine constraints will prevent you from entering a state where it isn't set.  It's still not quite as good for this sort of thing, but it also lets you deal with more complex state hierarchies where the "messages being responded to" is less clear (and I do think it's less clear here, or I'd just suggest an interface).

In any case - this is just something I will bring up *every* time I see a `None` now, because `None` is Python's manifestation of [the billion dollar mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare) and one day I'd like to stop making it.

manishtomar:
> Does this already exist somewhere else in the codebase?

[Yes](https://github.com/rackerlabs/otter/blob/4b5d9fe1ab6cce41a1e6273e21a54eeae6c86ccc/otter/util/deferredutils.py#L236-L269). This is being used when modifying state.

> I wish I could understand it better - I need to educate myself more as to the specifics of Zookeeper

https://zookeeper.apache.org/doc/trunk/recipes.html#sc_recipes_Locks without the watch :) 

While responding to these comments, I realized 2 things that need to be done: `release_eff` should ignore `NoNodeError` and [handle node creation failure by comparing it with children from getChildren](https://zookeeper.apache.org/doc/trunk/recipes.html#sc_recipes_errorHandlingNote). I'll probably leave the second one as tech-debt ticket.

otter/test/util/test_zk.py
manishtomar: I struggled with this myself. I am not sure why other tests use this.

otter/test/util/test_zk.py
manishtomar: I'll take it as an argument to object.

otter/util/zk.py
glyph: > I could change the tests to drive through external interface but I believe that is being tested in test_acquire_* functions. I am trying to think if it is a bad idea to test internal state without driving it from public interface. Not sure.

This is another hint that Active Nothing would be a good idea here.  _Ideally_, you would never drive any behavior through a private interface.  If you find yourself needing to break encapsulation to talk to a private attribute, you should break the code into two layers, and test the behavior of the inner layer.  The main benefit of doing this is that it increases the value of the tests as _regression_ tests, because maintenance of the code's implementation details (for example replacing the `_node` attribute with some state-machine goop!) will not cause them to need to change.  The tension is that if you have to set up a _lot_ of state, it decreases their value as _unit_ tests, so sometimes you have to just accept some ugliness in the name of isolation.  In this case the state-setup seems pretty minimal and could be accomplished with a tiny function, so I think it should be done, as it would be a higher-level expression of intent: you want a `PollingLock` in a particular _state_, not one with an attribute set to a specific _value_.

manishtomar: > In any case - this is just something I will bring up every time I see a None now, because None is Python's manifestation of the billion dollar mistake and one day I'd like to stop making it.

I think the best way to do this by using a language that does not allow it: Haskell :trollface:

glyph: Haskell makes the compiler yell at you you way more for using NULLs, but ultimately Just and Maybe still put the onus on the caller to figure out what to do with the stateful value.  I actually vastly prefer the "OO" solution to this problem, where the responsibility rests squarely with the callee.

Issue #1904, opened 2016-06-30
CASS and ZK script during deploy


manishtomar:
Rarely but surely there are deploys that require updates in Cassandra or ZK. The deploy pipeline should have means to run some script/change with possibly automated tests. This way it is automated and auditable.

PR #1927, opened 2016-08-16
Use polling lock when modifying state

### LINKED ISSUE #1915: LockTimeouts :(

manishtomar:
Using the `PollingLock` recipe implemented in #1925 when modifying group state and deleting group. This fixes #1915. Refactored tests to use `ZKLock` instead of fragile mocked lock object (thanks @glyph!; tracked in #1904). 

This was tested on dev vm with 3 otter instances connected to single zk on the same vm load balanced via nginx talking. I had removed [local lock](https://github.com/rackerlabs/otter/blob/d15e6659436f6822bf6952d4e1751b31c3d66759/otter/models/cass.py#L827) in `modify_state` and directly used kazoo lock as I could not reproduce the issue with local lock. This is still a mystery to me. Following are test results:

|Lock type|acquire timeout|later acquire timeout|webhook calls|locktimeouts|total errors|child nodes left|
| ------ | ------ | ------ | ------ | ------ | ------ | ------ |
||||||||
|kazoo|20|30|100|22|100|11|
|kazoo|20|30|200|18|200|24|
|kazoo|20|300|200|167|200|1|
|polling with 0.1 delay|20|25|100|0|0|0|
|polling with 0.1 delay|20|25|200|93|93|0|
|polling with 0.1 delay|20|25|200|17|17|0|
|polling with 0.01 delay|20|25|200|108|108|0|
|polling with 0.01 delay|20|25|200|109|109|0|

Above, *webhook calls* represent number of concurrent calls made. You will notice that `LockTimeout`s increases as the number of concurrent webhook calls increase. However, "child nodes left" is 0 when using `PollingLock` in any combination. This is important as this causes one lock failure to sustain in subsequent calls and is what `PollingLock` is trying to solve.

### DIFF
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -65,6 +65,15 @@ DEFAULT_CONSISTENCY = ConsistencyLevel.QUORUM
 
 QUERY_LIMIT = 10000
 
+# Max number of seconds to wait to acquire group kazoo lock. This should
+# not be held for more than 10-15ms in normal circumstances but we keep it high
+# for safety sake. We don't want it to be < 30 since any request will be
+# terminated by CLB by then.
+ACQUIRE_TIMEOUT = 10
+# Ideally release should not even timeout but keep this to capture any unwanted
+# error
+RELEASE_TIMEOUT = 10
+
 
 @attributes(['query', 'params', 'consistency_level'])
 class CQLQueryExecute(object):
@@ -824,12 +833,13 @@ class CassScalingGroup(object):
             return d.addCallback(_write_state)
 
         lock = zk.PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=10)
+        lock.acquire = functools.partial(lock.acquire, timeout=ACQUIRE_TIMEOUT)
         local_lock = self.local_locks.get_lock(self.uuid)
         return local_lock.run(
             with_lock, self.reactor, lock, _modify_state,
             log.bind(category='locking', lock_reason='modify_state'),
-            acquire_timeout=15, release_timeout=10)
+            acquire_timeout=ACQUIRE_TIMEOUT + 5,
+            release_timeout=RELEASE_TIMEOUT)
 
     def update_status(self, status):
         """
@@ -1319,11 +1329,11 @@ class CassScalingGroup(object):
                     otter_msg_type="ignore-delete-lock-error"))
 
         lock = zk.PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=10)
+        lock.acquire = functools.partial(lock.acquire, timeout=ACQUIRE_TIMEOUT)
         d = with_lock(self.reactor, lock, _delete_group,
                       log.bind(category='locking', lock_reason='delete_group'),
-                      acquire_timeout=15,
-                      release_timeout=10)
+                      acquire_timeout=ACQUIRE_TIMEOUT + 5,
+                      release_timeout=RELEASE_TIMEOUT)
         # Cleanup /locks/<groupID> znode as it will not be required anymore
         d.addCallback(_delete_lock_znode)
         d.addCallback(lambda _: None)
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -50,14 +50,13 @@ from otter.models.interface import (
     UnrecognizedCapabilityError,
     WebhooksOverLimitError,
     next_cron_occurrence)
-from otter.util import timestamp
+from otter.util import timestamp, zk
 from otter.util.config import config_value
 from otter.util.cqlbatch import Batch, batch
 from otter.util.deferredutils import with_lock
 from otter.util.hashkey import generate_capability, generate_key_str
 from otter.util.retry import repeating_interval, retry, retry_times
 from otter.util.weaklocks import WeakLocks
-from otter.util.zk import PollingLock
 
 
 LOCK_PATH = '/locks'
@@ -824,17 +823,13 @@ class CassScalingGroup(object):
                 self, state, *args, **kwargs))
             return d.addCallback(_write_state)
 
-        lock = PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=20)
+        lock = zk.PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
+        lock.acquire = functools.partial(lock.acquire, timeout=10)
         local_lock = self.local_locks.get_lock(self.uuid)
         return local_lock.run(
             with_lock, self.reactor, lock, _modify_state,
             log.bind(category='locking', lock_reason='modify_state'),
-            acquire_timeout=25, release_timeout=10)
-        #return with_lock(
-        #    self.reactor, lock, _modify_state,
-        #    log.bind(category='locking', lock_reason='modify_state'),
-        #    acquire_timeout=25, release_timeout=10)
+            acquire_timeout=15, release_timeout=10)
 
     def update_status(self, status):
         """
@@ -1323,12 +1318,12 @@ class CassScalingGroup(object):
                     exc=f.value,
                     otter_msg_type="ignore-delete-lock-error"))
 
-        lock = self.kz_client.Lock(LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=120)
+        lock = zk.PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
+        lock.acquire = functools.partial(lock.acquire, timeout=10)
         d = with_lock(self.reactor, lock, _delete_group,
                       log.bind(category='locking', lock_reason='delete_group'),
-                      acquire_timeout=150,
-                      release_timeout=30)
+                      acquire_timeout=15,
+                      release_timeout=10)
         # Cleanup /locks/<groupID> znode as it will not be required anymore
         d.addCallback(_delete_lock_znode)
         d.addCallback(lambda _: None)
--- a/otter/models/cass.py
+++ b/otter/models/cass.py
@@ -825,12 +825,16 @@ class CassScalingGroup(object):
             return d.addCallback(_write_state)
 
         lock = PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=10)
-        #local_lock = self.local_locks.get_lock(self.uuid)
-        return with_lock(
-            self.reactor, lock, _modify_state,
+        lock.acquire = functools.partial(lock.acquire, timeout=20)
+        local_lock = self.local_locks.get_lock(self.uuid)
+        return local_lock.run(
+            with_lock, self.reactor, lock, _modify_state,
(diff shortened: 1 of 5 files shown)


manishtomar:

manishtomar:
@glyph review please.

otter/tap/api.py
```diff
@@ -289,11 +289,12 @@ def on_client_ready(_):
             # Give dispatcher to Otter REST object
             otter.dispatcher = dispatcher
 
-            # Set the client after starting
+            # Set the client and dispatcher after starting
             # NOTE: There is small amount of time when the start is
             # not finished and the kz_client is not set in which case
             # policy execution and group delete will fail
             store.kz_client = kz_client
+            store.dispatcher = dispatcher
```
glyph: Obviously it would be nicer if this were a discrete event on `Otter` (use automat!), or even better constructor arguments.  Clearly this has been previously considered though, given the comment.

otter/test/models/test_cass_models.py
```diff
@@ -411,10 +413,23 @@ def _responses(*args):
         self.policies = []
         self.mock_log = mock.MagicMock()
 
-        self.kz_client = mock.Mock()
-        self.lock = self.mock_lock()
-        self.kz_client.Lock.return_value = self.lock
-        self.kz_client.delete.return_value = defer.succeed('something else')
+        self.kz_client = ZKCrudModel()
```
glyph: I wish there were some kind of fake verification here; a check for the consistency between `kz_client` here and what it is in real life.  It looks right to me I just wish I didn't have to manually check :)

otter/test/models/test_cass_models.py
```diff
@@ -411,10 +413,23 @@ def _responses(*args):
...
+        self.kz_client.nodes = {"/locks/" + self.group_id: ("", 0)}
+
+        self.acquire_call = (True, 10, True)
```
glyph: 10 is a magic number here and should be a constant.

otter/test/models/test_cass_models.py
```diff
@@ -411,10 +413,23 @@ def _responses(*args):
...
+
+        def create_ZKLock(disp, path):
+            self.assertEqual(disp, "dispatcher")
+            self.lock = ZKLock("client", path)
+            self.lock.acquire_call = self.acquire_call
+            self.lock.release_call = self.release_call
+            return self.lock
+
+        from otter.models.cass import zk
+        self.patch(zk, "PollingLock", create_ZKLock)
```
glyph: Speaking of checking correspondence between different things :).  The docstring for `ZKLock` hasn't even been updated to indicate that it's a mock of `PollingLock` yet, which makes me wonder if this `patch` is really correct.

Do you really need to mock out `PollingLock` like this?  Isn't all I/O already routed through this test's `dispatcher`, which won't do any Zookeeper I/O?

otter/test/models/test_cass_models.py
```diff
@@ -411,10 +413,23 @@ def _responses(*args):
...
+        self.addCleanup(setattr, self, "acquire_call", (True, 10, True))
```
glyph: 10 magic number again :)

otter/models/cass.py
```diff
@@ -1318,12 +1318,12 @@ def _delete_lock_znode(result):
                     exc=f.value,
                     otter_msg_type="ignore-delete-lock-error"))
 
-        lock = self.kz_client.Lock(LOCK_PATH + '/' + self.uuid)
-        lock.acquire = functools.partial(lock.acquire, timeout=120)
+        lock = zk.PollingLock(self.dispatcher, LOCK_PATH + '/' + self.uuid)
+        lock.acquire = functools.partial(lock.acquire, timeout=10)
```
glyph: Magic number - also - why is it going down by an order of magnitude?

Does this really need to be 10 for specific correctness reasons or should it be configurable?

otter/models/cass.py
```diff
@@ -1318,12 +1318,12 @@ def _delete_lock_znode(result):
...
         d = with_lock(self.reactor, lock, _delete_group,
                       log.bind(category='locking', lock_reason='delete_group'),
-                      acquire_timeout=150,
-                      release_timeout=30)
+                      acquire_timeout=15,
```
glyph: Magic number :)

otter/models/cass.py
```diff
@@ -1318,12 +1318,12 @@ def _delete_lock_znode(result):
...
+                      release_timeout=10)
```
glyph: Magic number :)

otter/test/models/test_cass_models.py
```diff
@@ -796,58 +812,60 @@ def modifier(_group, _state):
...
+            return modifier_d
 
         self.group.view_state = mock.Mock(return_value=defer.succeed('state'))
-        # setup local lock
+
         llock = defer.DeferredLock()
-        self.group.local_locks = mock.Mock(
-            get_lock=mock.Mock(return_value=llock))
+
+        class Locks(object):
```
glyph: What is this a mock for?

otter/test/models/test_cass_models.py
```diff
@@ -857,22 +875,16 @@ def test_modify_state_lock_not_acquired(self, mock_serial):
         ``modify_state`` raises error if lock is not acquired and does not
         do anything else
         """
-        self.lock.acquire.side_effect = \
-            lambda timeout: defer.fail(ValueError('a'))
+        self.acquire_call = (True, 10, Failure(ValueError("eh")))
```
glyph: Magic number.

otter/test/models/test_cass_models.py
```diff
@@ -857,22 +875,16 @@ def test_modify_state_lock_not_acquired(self, mock_serial):
...
-
         d = self.group.modify_state(modifier)
         self.failureResultOf(d, ValueError)
-
         self.assertEqual(self.connection.execute.call_count, 0)
-        self.kz_client.Lock.assert_called_once_with(
-            '/locks/' + self.group.uuid)
-        self.lock._acquire.assert_called_once_with(timeout=10)
-        self.assertEqual(self.lock.release.call_count, 0)
+        self.assertFalse(self.lock.acquired)
```
glyph: `.acquired` is not a public attribute on the real implementation, so making this test more realistic would also break it.  https://glyph.twistedmatrix.com/2015/05/separate-your-fakes-and-your-inspectors.html is relevant here, I think.

otter/test/models/test_cass_models.py
```diff
@@ -2069,20 +2079,15 @@ def test_delete_empty_scaling_group_with_zero_policies(self, mock_naive,
...
+        self.assertEqual(self.kz_client.nodes, {})
 
     @mock.patch('otter.models.cass.CassScalingGroup.view_state')
     def test_delete_lock_not_acquired(self, mock_view_state):
         """
         If the lock is not acquired, do not delete the group.
         """
-        self.lock.acquire.side_effect = \
-            lambda timeout: defer.fail(ValueError('a'))
+        self.acquire_call = (True, 10, Failure(ValueError('a')))
```
glyph: Magic number.

glyph:
My main concern here is really increases in technical debt in the test suite; the implementation itself looks solid and that table that you built *really* indicates you've done your homework to make sure this works and it should just land.

So it seems to me that the coverage is adequate, but could be more expressive / even less mocky; could you land this and then do some follow-up testing cleanups?

otter/tap/api.py
manishtomar: Oh yes. `Otter` object can be in multiple states. `automat` would be a good fit here!

otter/test/models/test_cass_models.py
manishtomar: > The docstring for ZKLock hasn't even been updated to indicate that it's a mock of PollingLock yet, which makes me wonder if this patch is really correct.

`PollingLock` is supposed to provide same interface as `KazooLock` and `ZKLock` is stub for `KazooLock`. Hence, `ZKLock` can be used as stub for `PollingLock` too. I'll update the docs to reflect this. 

> Do you really need to mock out PollingLock like this? Isn't all I/O already routed through this test's dispatcher, which won't do any Zookeeper I/O?

I don't _really_ need to mock `PollingLock`. And you are right: all IO is routed through test dispatcher. However, to make `PollingLock.acquire` return a particular value the test will need to know the internal implementation (effects) of `acquire` method which I don't prefer.

otter/models/cass.py
manishtomar: > why is it going down by an order of magnitude?

Good question. Actually, the diff is not up-to-date. It should not show 120. I changed it from 120 to 10 in #1923 mainly to see if `LockTimeouts` reduce. 120 was not good because 1 orphan child node will keep many pending requests running for 120 seconds. They will all be timed out by CLB in 30 seconds but unnecessarily remain in the system. When analyzing sample requests I found that the time the lock is actually held is for 10-15 ms. Hence, I wanted to change it to some number < 30s. I'll add some documentation and make this a constant. It might be complicated to make this configurable for now.

otter/test/models/test_cass_models.py
manishtomar: It's a stub object to remove above
```python
self.group.local_locks = mock.get_lock=mock.Mock(return_value=llock))
```
mocking code.

otter/test/models/test_cass_models.py
manishtomar: Hmm.. interesting. I hadn't thought of it like this. My immediate reaction is that this seems to be over-engineering mainly because the public attribute is being used in test-facing code only. But I respect your experience to realize that you might have _actually_ faced this problem. I'll try to refactor this in this PR but if it seems to be taking more time than I would like, I'll create a tech-debt ticket instead. Thanks!

otter/test/models/test_cass_models.py
manishtomar: It's a stub for [`WeakLocks`](https://github.com/rackerlabs/otter/blob/master/otter/util/weaklocks.py)

manishtomar:
> My main concern here is really increases in technical debt in the test suite

😞  I intended it to be exact opposite by removing some of the mocks. 

> could you land this and then do some follow-up testing cleanups?

I'll try cleaning up whatever I can here and leave others as tech-debt tickets. Although the only change I am planning to make _as of now_ is https://github.com/rackerlabs/otter/pull/1927#discussion_r75567737

glyph:
> > My main concern here is really increases in technical debt in the test suite
> 😞 My intention was exact opposite by removing some of the mocks.

Sorry @manishtomar - this was an unfair criticism.  It does indeed remove mocking and move in a positive direction.  What I meant was not really "increases" in technical debt; rather, the continued dependence upon things like `patch` and mocks.  This change itself is an overall improvement!  It's just not an optimal improvement.

manishtomar:
> Sorry @manishtomar - this was an unfair criticism.

No problem at all :) I "try" to take criticism constructively. 

> rather, the continued dependence upon things like patch and mocks

I would like to remove them wherever possible but unless all the code moves to Effect I find it difficult to do that. Moreover, one can remove this dependency slowly in small PRs. Even with Effect sometimes its till difficult to test without patching. We can talk about it face-to-face next time. 

> This change itself is an overall improvement!

Thanks!

glyph:
> No problem at all :) I "try" to take criticism constructively.

All the more reason it should be accurate :)

otter/tap/api.py
glyph: Rad, it would be nice to get some real usage out of Automat :)

PR #1932, opened 2016-09-10
ERROR group without drained_at

### LINKED ISSUE #1863: CLB drained_at error recovery

manishtomar:
Fixes #1863.


manishtomar:

manishtomar:

manishtomar:
@glyph review please

otter/convergence/model.py
```diff
@@ -572,11 +573,20 @@ class CLB(object):
...
+    lb_id = attr.ib()
+    node_id = attr.ib()
+
+
 @implementer(ILBNode, IDrainable)
 @attributes([Attribute("node_id", instance_of=basestring),
              Attribute("description", instance_of=CLBDescription),
              Attribute("address", instance_of=basestring),
-             Attribute("drained_at", default_value=0.0, instance_of=float),
+             Attribute("drained_at", default_value=None),
```
glyph: Active nothing! Adding `None` to `drained_at` means needing to check every single access point in the codebase.  Perhaps this should be a private `attr` with a public `@property` that raises `DrainingUnavailable`?

requirements/mimic.txt
```diff
@@ -1 +1 @@
-git+https://github.com/rackerlabs/mimic.git@f317ee5d322aa7cb2b117e2984dfe8ef10558cd5
+git+https://github.com/rackerlabs/mimic.git@594dd5c8e80b670fa2d0c42f5eec9645e14aa54e
```
glyph: Hrm.  Perhaps we're a bit behind on doing a new Mimic release?

otter/convergence/model.py
```diff
@@ -521,6 +521,7 @@ def is_done_draining(now, timeout):  # pragma: no cover
 
         :return: Whether the node is done draining.
         :rtype: `bool`
+        :raises: ``DrainingUnavailable`` if draining info is not available
```
glyph: Under what conditions is draining info not available?

otter/convergence/model.py
manishtomar: Good idea!

PR #1934, opened 2016-09-26
SelfHeal service refactoring

manishtomar:
Separate out locking from SelfHeal service and put it as functions in `otter.util.zk`.  This way it can be used to implement https://github.com/rackerlabs/autoscaling-chef/issues/833 which has similar pattern as SelfHeal service: call a function if it has the lock on timely basis.


manishtomar:

manishtomar:

manishtomar:

manishtomar:
@glyph review please.

otter/convergence/selfheal.py
```diff
@@ -78,103 +72,36 @@ def _cancel_scheduled_calls(self):
...
         if active:
             # This should never happen
-            self.log.err(RuntimeError("self-heal-calls-err"),
-                         "self-heal-calls-err", active=active)
+            self.log.err(RuntimeError("selfheal-calls-err"),
+                         "selfheal-calls-err", active=active)
         if not groups:
             returnValue(None)
-        wait_time = float(time_range) / len(groups)
+        wait_time = float(self.time_range) / len(groups)
```
glyph: This `float` call is somewhat redundant; `self.time_range` has a validator *and* a converter of `float`.

otter/convergence/selfheal.py
```diff
@@ -78,103 +72,36 @@ def _cancel_scheduled_calls(self):
...
-        groups = yield perform(self.disp,
-                               get_groups_to_converge(config_func))
+        groups = yield perform(self.dispatcher,
+                               get_groups_to_converge(self.config_func))
         active = self._cancel_scheduled_calls()
         if active:
             # This should never happen
-            self.log.err(RuntimeError("self-heal-calls-err"),
-                         "self-heal-calls-err", active=active)
+            self.log.err(RuntimeError("selfheal-calls-err"),
```
glyph: Why the change to the log ID?

otter/tap/api.py
```diff
@@ -309,18 +313,40 @@ def on_client_ready(_):
...
 
         d.addCallback(on_client_ready)
         d.addErrback(log.err, 'Could not start TxKazooClient')
 
     return parent
 
 
+def setup_selfheal_service(clock, config, dispatcher, health_checker, log):
+    """
+    Setup selfheal timer service and add it to ``parent``
```
glyph: What `parent`?

otter/test/convergence/test_selfheal.py
```diff
@@ -30,72 +30,54 @@ class SelfHealTests(SynchronousTestCase):
...
+        self.patch(sh, "check_and_trigger", lambda t, g: t + g)
+        self.s = sh.SelfHeal(self.clock, base_dispatcher, "cf", 300.0,
+                             self.log)
+        self.groups = [
+            {"tenantId": "t{}".format(i), "groupId": "g{}".format(i)}
+            for i in range(5)]
+
+    def test_call(self):
+        """
+        ``self.s()`` will setup convergences to be triggered over specified
```
glyph: This should probably be documented as ``:obj:`SelfHeal` `` rather than ``self.s()``, since it's a reference to the behavior of the `SelfHeal` class.

otter/test/convergence/test_selfheal.py
```diff
@@ -30,72 +30,54 @@ class SelfHealTests(SynchronousTestCase):
...
+            self.assertEqual(c.getTime(), i * 60)
+            self.assertEqual(c.func, sh.perform)
+            self.assertEqual(c.args,
+                             (self.s.dispatcher, "t{}g{}".format(i, i)))
 
-    def test_setup_again_on_err(self):
+    def test_call_err(self):
         """
-        Calls _setup_if_locked at every interval even it it fails
+        ``self.s()`` will log any error and return success
```
glyph: Same as above.

otter/convergence/selfheal.py
```diff
@@ -4,72 +4,66 @@
...
 from otter.convergence.service import trigger_convergence
+from otter.log import BoundLog
 from otter.log.intents import msg, with_log
 from otter.models.intents import GetAllValidGroups, GetScalingGroupInfo
 from otter.models.interface import NoSuchScalingGroupError, ScalingGroupStatus
-from otter.util import zk
 
 
-class SelfHeal(MultiService, object):
+@attr.s
```
glyph: The switch to `attr.s` is welcome, thanks :)

otter/convergence/selfheal.py
manishtomar: Good catch. Thanks!

otter/convergence/selfheal.py
manishtomar: I went through lot of iterations of this code as you can see in commits and decided to use this ID probably because I was generating this ID instead of hard-coding it at one time. In any case, this seemed more appealing due to its one less character :). However, thanks to you I just realized that this to updated in `spec.py` where it is expanded.

otter/tap/api.py
manishtomar: Good catch. I was passing `parent` earlier and adding this service to that but didn't feel comfortable with mutability aspect of the function. So moved the adding to caller but forgot to update the docs.

PR #1941, opened 2016-11-09
unpin lxml so we can use the latest version

shawnashlee:
pinning lxml to 3.4.1 was causing the devvm  to jump through a lot of hoops to build it. Functional tests pass with the latest version so we should unpin it (or we can pin it to 3.6.4).

https://gist.github.com/shawnashlee/dedc78e73341fadf1e1c19673a9e1be7

```
Ran 176 tests in 82.525s

OK (skipped=8)
```


mburns:
lgtm, fwiw

PR #1943, opened 2016-11-21
Nosuchendpoint error is terminal

manishtomar:
When an account is suspended or closed, its groups error with `NoSuchEndpoint` probably because its service catalog does not contain compute or CLB services. These group's convergence fails and keeps trying again every 30s because convergence considers any unknown error as  transient. This PR considers it as terminal and stops further convergence. This is temporary until https://github.com/rackerlabs/autoscaling-chef/issues/833 is implemented.


manishtomar:
@glyph review please. Minor change.

glyph:
Nice compact fix. Merge away.

PR #1944, opened 2016-11-22
remove temporary part of the comment - thanks @glyph

manishtomar:
Since it is possible to execute convergence on a suspended account even after implementing https://github.com/rackerlabs/autoscaling-chef/issues/833 as convergence could've been triggered just after account got suspended but before otter received the event.


glyph:
LGTM.

PR #1946, opened 2016-11-30
One CLB mutating type step per CLB

### LINKED ISSUE #1939: One CLB write call per convergence cycle
If planner returns multiple types of steps w.r.t a CLB (like add AND delete nodes) then otter calls them concurrently causing CLB to go into ERROR sometimes. The chances increase if there are many nodes involved in the call. 

This can be fixed if steps transformer allows only one mutating call per CLB since it any mutating call results in PENDING_UPDATE state which stays for a few seconds. Next convergence call can take care of other mutating calls.

manishtomar:
Fixes #1939.

### DIFF
--- a/otter/convergence/transforming.py
+++ b/otter/convergence/transforming.py
@@ -92,28 +92,24 @@ _register_bulk_rcv3_optimizer(BulkAddToRCv3)
 _register_bulk_rcv3_optimizer(BulkRemoveFromRCv3)
 
 
-def filter_clb_mutating_types(steps):
+def one_clb_step(steps):
     """
-    Allow only one CLB mutating steps per CLB
+    Allow only one CLB mutating step per CLB since calling only one will be
+    allowed in CLB while others will get PENDING_UPDATE.
 
     :param steps: Iterable of :obj:`IStep` instances
     :return: Iterable of :obj:`IStep` instances such that only one mutating
-        type is returned per CLB
+        step is returned per CLB
     """
-    mutating_clb_types = (AddNodesToCLB, RemoveNodesFromCLB, ChangeCLBNode)
-    lb_step_type = {}
+    # set of loadbalancer IDs whose step has been returned
+    seen_lbids = set()
     for step in steps:
-        stype = type(step)
-        if stype not in mutating_clb_types:
-            yield step
-            continue
-        lb_stype = lb_step_type.get(step.lb_id)
-        if lb_stype is None:
-            lb_step_type[step.lb_id] = stype
-            yield step
-        else:
-            if stype is lb_stype:
-                yield step
+        if isinstance(step, (AddNodesToCLB, RemoveNodesFromCLB,
+                             ChangeCLBNode)):
+            if step.lb_id in seen_lbids:
+                continue
+            seen_lbids.add(step.lb_id)
+        yield step
 
 
 def optimize_steps(steps):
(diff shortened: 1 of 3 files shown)


manishtomar:

manishtomar:

manishtomar:
@glyph review please

otter/convergence/transforming.py
```diff
@@ -91,6 +92,30 @@ def optimize_steps(rcv3_steps):
...
+    mutating_clb_types = (AddNodesToCLB, RemoveNodesFromCLB, ChangeCLBNode)
+    lb_step_type = {}
+    for step in steps:
+        stype = type(step)
+        if stype not in mutating_clb_types:
+            yield step
+            continue
+        lb_stype = lb_step_type.get(step.lb_id)
+        if lb_stype is None:
+            lb_step_type[step.lb_id] = stype
```
glyph: This algorithm will allow multiple concurrent modifications _of the same type_, which I think is not what's intended.  As you explained to me on slack; `AddNodesToCLB` and `RemoveNodesFromCLB` will be coalesced into single batch steps, but `ChangeCLBNode` won't (can't) be, which means this could still allow through multiple concurrent `ChangeCLBNode` calls, and that would leave us in the same problematic scenario we now find ourselves.

The good news is that the algorithm actually gets quite a bit simpler if you implement it to simply base it on LB ID rather than LB ID plus type; something like this should work:

```python
def filter_clb_mutating_types(steps):
    lbs_with_mutations = set()  # set of loadbalancer IDs
    for step in steps:
        if isinstance(step, (AddNodesToCLB, RemoveNodesFromCLB,
                             ChangeCLBNode)):
            if step.lb_id in lbs_with_mutations:
                continue
            lbs_with_mutations.add(step.lb_id)
        yield step
```

manishtomar: > This algorithm will allow multiple concurrent modifications of the same type, which I think is not what's intended.

This function is called _before_ the steps are combined. Hence it returns multiple steps of the same type. However, things become drastically simple if this function is called _after_ combining the steps whose algorithm is given by you above. I'll make that change. Thanks! 

(How this didn't strike me will hurt me for a long time)

otter/test/convergence/test_transforming.py
```diff
@@ -87,9 +96,129 @@ def test_get_limits_conf(self):
...
+        clb_types = (AddNodesToCLB, RemoveNodesFromCLB, ChangeCLBNode)
+        lb_steps, no_lb_steps = partition_bool(lambda s: type(s) in clb_types,
+                                               steps)
+        steps_by_id = groupby(lambda s: s.lb_id, lb_steps)
+
+        filtered = filter_clb_mutating_types(steps)
+
+        filtered_steps_by_id = groupby(lambda s: getattr(s, "lb_id", "no"),
+                                       filtered)
+        for lb_id, lb_steps in steps_by_id.items():
```
manishtomar: I agree. When I started writing regular unit tests there seemed to be too many cases which tested the same property: "Are the steps returned of one type per CLB?". So I thought of using hypothesis instead. However, testing that property for any kind of input turned out to be more work than expected. In any case, now that the function will be very simple it can have regular unit tests. 

(I was hoping to sneak in hypothesis with this PR but maybe some other time)

manishtomar:

manishtomar:
@glyph could you approve the changes?

manishtomar:
@glyph Let me fix the tests. I thought they passed.

manishtomar:

manishtomar:
@glyph Now could you approve changes? :)

glyph:
In the queue, although working with @mburns today.

glyph:
Yep, looks good!

manishtomar:
@glyph Thanks!

PR #1948, opened 2016-12-03
Group ERRORs on NoSuchEndpoint exception

manishtomar:
This is the better version of #1943 which was in fact buggy fix. It should catch `FirstError` and check for `NoSuchEndpoint`. This PR reverts that change. It catches the error during convergence execution and ERRORs the group with the appropriate reason. This is better because selfheal does not trigger convergence once the group is in ERROR. This was not the case with #1943 where it temporarily stopped convergence but selfheal would've triggered it every 5 mins. Tested this on dev by pointing to prod identity and using a suspended account.

### DIFF
--- a/otter/test/convergence/test_errors.py
+++ b/otter/test/convergence/test_errors.py
@@ -2,6 +2,7 @@ import traceback
 
 from twisted.trial.unittest import SynchronousTestCase
 
+from otter.auth import NoSuchEndpoint
 from otter.cloud_client import (
     CLBDeletedError,
     CLBNodeLimitError,
@@ -49,7 +50,10 @@ class PresentReasonsTests(SynchronousTestCase):
             CreateServerConfigurationError("Your server is wrong"):
                 'Server launch configuration is invalid: Your server is wrong',
             CreateServerOverQuoteError("You are over quota"):
-                'Servers cannot be created: You are over quota'
+                'Servers cannot be created: You are over quota',
+            NoSuchEndpoint(service_name="nova", region="ord"):
+                "Could not locate service nova in the service catalog. "
+                "Please check if your account is still active."
         }
         excs = excs.items()
         self.assertEqual(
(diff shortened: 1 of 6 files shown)


manishtomar:

manishtomar:

manishtomar:
@glyph review please.

otter/test/convergence/test_errors.py
```diff
@@ -49,7 +50,10 @@ def test_present_exceptions(self):
             CreateServerConfigurationError("Your server is wrong"):
                 'Server launch configuration is invalid: Your server is wrong',
             CreateServerOverQuoteError("You are over quota"):
-                'Servers cannot be created: You are over quota'
+                'Servers cannot be created: You are over quota',
+            NoSuchEndpoint(service_name="nova", region="ord"):
+                "Could not locate service nova in the service catalog. "
+                "Please check if your account is still active."
```
glyph: Pro Tip: put a `,` at the end of this line to avoid it showing up in the next diff that modifies this dict (as the 'servers cannot be created' line shows up in this one because of the new comma...).

manishtomar:
> If we can't detect the disabling of an account via an event, then it doesn't seem there's a way to detect resumption either.

This doesn't replace [terminator](https://github.com/rackerlabs/autoscaling-chef/issues/833)  which will track suspend and activate event and change the group accordingly. That is being worked on separately. This PR handles the cases before that fix is deployed. Even after implementing that, the event could be delayed and otter could be converging a suspended account. 

> What happens if there's a temporary billing glitch, or other error, which causes an account to be briefly suspended? Will self-healing eventually pick it up somehow?

The group will go into ERROR with proper reason. Self-healing will not pick it up. It will need to be re-enabled which support can do if and when such situation occurs. 

> Has this been communicated to support somehow?

Support doesn't need to be informed because group's error reason will be accessible by API. Support has fixed error groups on many occasions when its error reason was self-explanatory.

glyph:
Thanks for the explanation. In that case - looks great to me; merge away :)

PR #1949, opened 2016-12-07
Dev env using docker

manishtomar:
This is complete otter development environment using `docker-compose`. One needs to have docker installed on a machine and running `docker-compose up -d` should run otter with integration tests. No need for dev vm! 🍾


manishtomar:
@glyph and @shawnashlee could you try this out and let me know if this works? Also, review please :)

glyph:
Looks good.  It works. But:

- Please file an issue upstream with both docker-compose and Twisted for the 'blocking stdout' issue.  Both should be fixed, I think.
- Updating the (private) autoscale chef repo install instructions with docker for mac might be a good idea as well.

manishtomar:
Thanks @glyph. I'll address the above comments later. I'll merge this as I really need it for other development :)

PR #1951, opened 2016-12-14
Fixes error introduced in commit 5dcd716

manishtomar:
`assoc_obj` was the real culprit in commit 5dcd716 which returned `CLBNode` with `_drained_at=None` all the time. (I never liked it from the beginning). This made the unit tests pass as code and test got objects with `None` `drained_at`. Fixed by switching to attr and using its assoc method which ensures the attribute exists before accepting.

### DIFF
--- a/otter/convergence/gathering.py
+++ b/otter/convergence/gathering.py
@@ -35,7 +35,6 @@ from otter.convergence.model import (
     group_id_from_metadata)
 from otter.indexer import atom
 from otter.models.cass import CassScalingGroupServersCache
-from otter.util.fp import assoc_obj
 from otter.util.http import append_segments
 from otter.util.retry import (
     exponential_backoff_interval, retry_effect, retry_times)
@@ -211,9 +210,9 @@ def get_clb_contents():
         if node.description.lb_id in deleted_lbs:
             return None
         if feed is not None:
-            return assoc_obj(node, drained_at=extract_clb_drained_at(feed))
-        else:
-            return node
+            node.drained_at = extract_clb_drained_at(feed)
+        return node
+
     nodes = map(update_drained_at, concat(lb_nodes.values()))
     yield do_return((
         list(filter(bool, nodes)),
--- a/otter/test/convergence/test_gathering.py
+++ b/otter/test/convergence/test_gathering.py
@@ -4,6 +4,8 @@ from copy import deepcopy
 from datetime import datetime
 from functools import partial
 
+import attr
+
 from effect import (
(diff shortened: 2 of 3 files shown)


manishtomar:

manishtomar:

manishtomar:
@glyph review please

manishtomar:
@glyph Please note that https://github.com/rackerlabs/otter/commit/5dcd716adbf9865062a6f86f72e6ac01d5f4b799 has not been deployed due to failed integration tests.

otter/test/convergence/test_gathering.py
```diff
@@ -605,7 +604,7 @@ def test_lb_disappeared_during_feed_fetch(self):
         eff = get_clb_contents()
         self.assertEqual(
             perform_sequence(seq, eff),
-            ([assoc_obj(CLBNode.from_node_json(2, node21), drained_at=2.0)],
+            ([attr.assoc(CLBNode.from_node_json(2, node21), _drained_at=2.0)],
```
glyph: Huh. That's an interesting bug in `assoc`; I thought it would have behaved the same as `__init__` when it came to private variables.

otter/convergence/gathering.py
```diff
@@ -211,9 +210,9 @@ def update_drained_at(node):
         if node.description.lb_id in deleted_lbs:
             return None
         if feed is not None:
-            return assoc_obj(node, drained_at=extract_clb_drained_at(feed))
-        else:
-            return node
+            node.drained_at = extract_clb_drained_at(feed)
```
glyph: Why is this being changed to mutate its input rather than return a new value?  What is the input being shared with?

otter/test/convergence/test_gathering.py
manishtomar: https://github.com/hynek/attrs/issues/116#issuecomment-263853383

otter/convergence/gathering.py
manishtomar: > Why is this being changed to mutate its input rather than return a new value?

I didn't want to do call `assoc` with a private variable. 

> What is the input being shared with?

What input? `node` argument? Sorry, didn't understand.

glyph: Yes, the `node` argument.

glyph: As far as calling `assoc` with a private variable... that seems like a bug in `attrs` which this is just a workaround for (and you're not reaching across library boundaries, at least)

PR #1952, opened 2016-12-14
remove unnecessary undocumented on_error flag

manishtomar:
`POST ../groupId/converge` contains an undocumented flag `on_error=False` which when given will check if group is in ERROR before triggering convergence. The documented and right behavior is to always trigger convergence. This was a hack for cron based selfheal that was removed in #1887. This change should have been along that.

### DIFF
--- a/otter/rest/groups.py
+++ b/otter/rest/groups.py
@@ -670,20 +670,10 @@ class OtterGroup(object):
         Trigger convergence on given scaling group
         """
 
-        class ConvergeErrorGroup(Exception):
-            pass
-
         def can_converge(group, state):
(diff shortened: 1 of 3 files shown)


manishtomar:
@glyph review please

glyph:
What about https://github.com/rackerlabs/otter/blob/6df2b822fb8543727d22476d88d0a76b7dbee11c/scripts/trigger_convergence.py#L67-L70 ?

manishtomar:
@glyph ooh. Good catch. Will remove it there also. That script was called by cron job earlier.

PR #1953, opened 2016-12-15
GroupState class using attrs

manishtomar:
Creating `GroupState` class using attrs. Fixed many tests after adding types. This can be used to implement part of terminator PR that adds suspended attribute.


manishtomar:
@glyph Review please. I am guessing coverage decrease is due to delete tests that are no longer required due to attrs.

glyph:
"coverage decrease" is never what I look for, but rather "new uncovered lines" :)

otter/models/interface.py
```diff
@@ -3,89 +3,83 @@
...
 class GroupState(object):
     """
     Object that represents the state
 
     :ivar bytes tenant_id: the tenant ID of the scaling group whose state this
         object represents
-
     :ivar bytes group_id: the ID of the scaling group whose state this
         object represents
-    :ivar bytes group_name: the name of the scaling group whose state this
```
glyph: This should still have type information associated with it.  Given the intent of the rest of the PR, it should probably be`text_type` rather than `bytes`.

otter/models/interface.py
```diff
@@ -3,89 +3,83 @@
...
+    :ivar group_name: the name of the scaling group whose state this
         object represents
     :ivar dict active: the mapping of active server ids and their info
     :ivar dict pending: the list of pending job ids and their info
-    :ivar bytes group_touched: timezone-aware ISO860 formatted timestamp
```
glyph: Same here.

otter/test/test_util.py
```diff
@@ -892,3 +893,24 @@ def test_auth_error_propogates(self):
...
+    def test_unicode(self):
+        """ unicode is returned as such """
+        self.assertEqual(text(u"test"), u"test")
+
+    def test_string(self):
+        """ string is converted to unicode """
+        self.assertEqual(text("test"), u"test")
+
+    def test_py3(self):
+        """ Not implemented for PY3 """
```
glyph: This should be a link to an issue.  Patching `six.PY2` is a pretty gross way to test this :-\\

otter/util/http.py
```diff
@@ -543,3 +545,16 @@ def try_json_with_keys(maybe_json_error, keys):
         return None
     else:
         return get_in(keys, error_body, None)
+
+
+def text(thing):
```
glyph: The purpose of this function, as I understand it, is to provide a _temporary_ "lenient" text converter for code which is sloppily bandying about bytes objects.  This makes sense, but it should have a future-looking plan for how to deal with it.  My earlier comment was to suggest that it should be implemented for py3, but a better future-looking plan would be to just start emitting deprecation warnings for bytes to ensure that everyone using APIs marked as having a `text` converter are properly passing text.  (Consequently, this should have a more descriptive name for that role, like `lenient_ascii_text`.)

For what it's worth, the purpose of `six.u` is to be used only with string _literals_, not with string input from other code.  Early versions of python 3 did not support the `u''` prefix, which is what `u()` was designed to mimic; it was a way you could say `u("someval")` and get an ascii-only unicode string on both 2 and 3.  Since every version of py3 that you should even consider supporting these days is capable of the `u""` syntax, an explicit `thing.decode("ascii")` would be more explicit.

manishtomar:
> A good refactoring, mostly, but it disturbs me that created_at can be changed from a datetime object to a string with so little other code changing. Is Cassandra really OK with a string going into a date field like this?

Yeah. That is pretty bad. It works because that object is not stored in CASS. It is only used to return back to REST API which [ignores it when formatting it](https://github.com/rackerlabs/otter/blob/23b61d0e3789e8df8ae8971b5e41d9e9516cbf80/otter/rest/groups.py#L55-L91).

otter/util/http.py
manishtomar: Thanks for the details. While I completely agree with you I am not going to provide warning in this PR as there is no framework to provide that as of now. Its either info or err. I've created ticket #1959 to address this.

manishtomar:

manishtomar:

PR #1955, opened 2016-12-16
Servers cache bugfix

### LINKED ISSUE #1828: Group state returns empty active list
Since [updating cache](https://github.com/rackerlabs/otter/blob/cde0bbd9f0694a45a7ced6a9be4f3c4b4249a1d4/otter/models/cass.py#L1834-L1860) first deletes existing entries and inserts new one, the cache returned can be empty if it is called in between these calls. This can be more visible when updating cache with many 1000+ servers because inserting 1000 servers in one batch call takes time.

manishtomar:
Fixes #1828. Each update of cache is stored on a timestamp. The implementation expects it to be monotonically increasing which it currently almost does as each update call is always from the same node. Unfortunately, the timestamp knowledge of implementation has creeped into the interface and its callers which I am not proud of. IMHO a simple impl that doesn't store based on timestamp but just merges the servers during update operation will be nice but it requires DB schema changes that I don't want to do now.

### DIFF
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -348,7 +348,7 @@ def execute_convergence(tenant_id, group_id, build_timeout, waiting,
     # Handle the status from execution
     if worst_status == StepResult.SUCCESS:
         result = yield convergence_succeeded(
-            executor, scaling_group, group_state, resources, now_dt)
+            executor, scaling_group, group_state, resources)
     elif worst_status == StepResult.FAILURE:
         result = yield convergence_failed(tenant_id, group_id, reasons)
     elif worst_status is StepResult.LIMITED_RETRY:
@@ -376,8 +376,7 @@ def update_stacks_cache(scaling_group, now, stacks, include_deleted=True):
 
 
 @do
-def convergence_succeeded(executor, scaling_group, group_state, resources,
-                          now):
+def convergence_succeeded(executor, scaling_group, group_state, resources):
     """
     Handle convergence success
     """
@@ -392,6 +391,7 @@ def convergence_succeeded(executor, scaling_group, group_state, resources,
         yield cf_msg('group-status-active',
                      status=ScalingGroupStatus.ACTIVE.name)
     # update servers cache with latest servers
+    now = yield Effect(Func(datetime.utcnow))
     yield executor.update_cache(scaling_group, now, include_deleted=False,
                                 **resources)
     yield do_return(ConvergenceIterationStatus.Stop())
--- a/otter/convergence/service.py
+++ b/otter/convergence/service.py
@@ -83,6 +83,21 @@ The top-level entry-points into this module are :func:`trigger_convergence` and
 # or config has changed), only if there are _any_ outstanding requests for
 # convergence, since convergence always uses the most recent data.
 
+
+# # Note [Convergence servers cache]
+# Each convergence cycle runs 3 primary steps: gather, generate plan and
+# execute plan. For launch_server type convergence it keeps cache of servers
+# fetched during gather step to provide 2 functionalities: provide quick list
+# for `GET .../groupId/state` API and to keep track of deleted servers IPs to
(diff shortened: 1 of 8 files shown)


manishtomar:

manishtomar:
@glyph review please

otter/models/cass.py
```diff
@@ -1831,43 +1826,53 @@ def _dict(r): return json.loads(r['server_blob'])
...
-        else:
-            return cql_eff(batch(queries, get_client_ts(self.clock)), params)
+    @do
+    def update_servers(self, time, servers):
+        """
+        See :method:`IScalingGroupServersCache.update_servers`
+
+        This implementation makes two important assumptions which needs to be
+        taken care by caller:
+        - It is not re-entrant for a given object; i.e a call to this must
```
glyph: This is what locks are for.  Why isn't there a mutex of some kind around this instead of this (destined to be ignored) caution?

otter/models/cass.py
```diff
@@ -1831,43 +1826,53 @@ def _dict(r): return json.loads(r['server_blob'])
...
+          be returned before another can be made
+        - `time` must be higher for subsequent calls.
```
glyph: what happens when it isn't?  leap seconds are a thing...

otter/convergence/service.py
```diff
@@ -392,6 +391,7 @@ def convergence_succeeded(executor, scaling_group, group_state, resources,
         yield cf_msg('group-status-active',
                      status=ScalingGroupStatus.ACTIVE.name)
     # update servers cache with latest servers
+    now = yield Effect(Func(datetime.utcnow))
```
glyph: Presumably the timing of this function call is pretty important.  When does it happen? Should the docstring say anything about that, or a comment at least?

otter/models/cass.py
```diff
@@ -1831,43 +1826,53 @@ def _dict(r): return json.loads(r['server_blob'])
...
+        # Insert new ones
+        if servers:
+            query = ('INSERT INTO {cf} ("tenantId", "groupId", last_update, '
+                     'server_id, server_blob, server_as_active) '
+                     'VALUES(:tenantId, :groupId, :last_update, :server_id{i},'
+                     ' :server_blob{i}, :server_as_active{i});')
+            params = assoc(self.params, "last_update", time)
+            queries = []
+            for i, server in enumerate(servers):
+                params['server_id{}'.format(i)] = server['id']
```
glyph: multi-key updates like this I prefer to spell differently, i.e.:

```python
                params.update({
                    'server_id{}'.format(i): server['id'],
                    'server_as_active{}'.format(i): server.pop('_is_as_active', False)
                    'server_blob{}'.format(i): json.dumps(server)
                })
```

If you want to get really fancy, [`merge`](http://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.merge) will let you do this and return a new value, which will also let you avoid mutating `queries` as you build it; `cql_eff(batch([<big generator expression>]))` would be a point-free way of expressing this which means it's a lot easier to see there are no side-effects.

otter/convergence/service.py
manishtomar: Yes. There should be comment about this. Unfortunately there is dependency between multiple functions calls that should be documented. Will do. Thanks!

otter/models/cass.py
manishtomar: Using generator expression is a nice idea! Thanks!

otter/models/cass.py
manishtomar: Because this function is called serially by only one node guaranteed by convergence partitioner. The same is for `time` comment below. But yeah the leap seconds can cause issue. I'll create ticket to track this.

manishtomar: Sorry just realized that reentrancy can still be an issue with calls from same node. I'll add that to the ticket.

manishtomar: #1966

PR #1962, opened 2016-12-22
A group can be suspended

manishtomar:
This is first among the set of [terminator](https://github.com/rackerlabs/autoscaling-chef/issues/833) PRs. This adds "suspended" attribute to the group including CASS that will be set by terminator when it finds out that associated tenant has been suspended. Although repose will not allow any calls to suspended accounts this errors any modification calls on suspended group for mimic based integration tests.


manishtomar:

manishtomar:

manishtomar:

manishtomar:

manishtomar:
@glyph review please

manishtomar:
@glyph self-reviewing shows some unpleasant things. Let me fix them and come back.

manishtomar:

manishtomar:

manishtomar:
@glyph review again please.

otter/controller.py
```diff
@@ -98,6 +98,16 @@ def __init__(self, tenant_id, group_id, operation, extra=None):
             fmt.format(t=tenant_id, g=group_id, o=operation))
 
 
+class TenantSuspendedError(Exception):
```
glyph: Just a thought; you might want to have `tenant_id` as an actual attribute, in case anyone handling this error ever wants to inspect which tenant ID it is without regexing it out of the `str`.

otter/controller.py
```diff
@@ -306,17 +316,25 @@ def modify_state(group_info):
 @defer.inlineCallbacks
 def modify_and_trigger(dispatcher, group, logargs, modifier, *args, **kwargs):
     """
-    Modify group state and trigger convergence after that
+    Modify group state and trigger convergence after that. Checks if group is
```
glyph: The docstring should call out the failure mode - i.e. it no longer returns `Deferred with None`, it now returns "Deferred firing with None when the group state has been modified, and fails with TenantSuspendedError if the tenant containing the group is suspended".

glyph: In other words, just saying "checks" is generally bad - the caller doesn't care whether the implementation "checks" or not, they care about what happens when the check fails or succeeds.

otter/convergence/selfheal.py
```diff
@@ -126,7 +126,8 @@ def check_and_trigger(tenant_id, group_id):
                   tenant_id=tenant_id, scaling_group_id=group_id)
     else:
         state = info["state"]
-        if state.status == ScalingGroupStatus.ACTIVE and (not state.paused):
+        if (state.status == ScalingGroupStatus.ACTIVE and
+                (not state.paused) and (not state.suspended)):
```
glyph: Dumb nitpick, but this might read more nicely as:

`(state.status == ScalingGroupStatus.ACTIVE and not (state.paused or state.suspended))`

otter/models/cass.py
```diff
@@ -495,6 +496,9 @@ def _unmarshal_state(state_dict):
 
     status = _group_status(state_dict['status'],
                            state_dict.get('deleting', False))
+    suspended = state_dict["suspended"]
```
glyph: Should this just be `bool(state_dict["suspended"])`?

glyph: Looking at the `bool` validator below, it seems like "yes"...

otter/test/test_controller.py
```diff
@@ -1510,19 +1505,32 @@ class ModifyAndTriggerTests(SynchronousTestCase):
...
-             nested_sequence([(("tg", "t", "g"), noop)]))
+             nested_sequence([(("tg", "tid", "gid"), noop)]))
         ])
 
     def modify(self, group, state):
         return "newstate"
 
+    def test_tenant_suspended(self):
+        """
+        Raises :obj:`TenantSuspendedError` if associated tenant is suspended.
```
glyph: I prefer to say "fails with" rather than "raises", since `delete_group` doesn't raise anything synchronously.

docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
             - ./otter:/app/otter
             - ./scripts:/app/scripts
             - ./twisted:/app/twisted
+            - ./schema:/app/schema
```
glyph: This reminds me that there should really be a second `docker-compose.yml` that actually ships all the code in the image rather than depending on volumes.  This is better / quicker for development, but not a best practice for deployment.

manishtomar: Yeah. That is true. However, my (intial?) goal for docker was to primarily set up dev environment. Doing deploys via docker are not gonna happen anytime soon.

manishtomar:
> My one real open question here is - how are schema modifications applied / coordinated?

You are not going to like this. They aren't automated. I manually apply the modifications in each region :(

otter/controller.py
manishtomar: Why? Whats wrong with regexing the string? I think its a perfectly valid solution over storing the attribute :P (just kidding)

otter/convergence/selfheal.py
manishtomar: Not dumb at all. It definitely reads better.

PR #1967, opened 2017-01-05
Common function to read feed entries

manishtomar:
Part of [terminator](https://github.com/rackerlabs/autoscaling-chef/issues/833) PR. 

Common function to read feed entries is implemented in `otter.cloud_client.cloudfeeds` which is currently being used by `clb.get_node_feed` and will also be used by terminator to get customer access events.

### DIFF
--- a/otter/constants.py
+++ b/otter/constants.py
@@ -34,22 +34,23 @@ def get_service_configs(config):
     :param dict config: Config from file containing service names that will be
         there in service catalog
     """
+    region = config['region']
     configs = {
         ServiceType.CLOUD_SERVERS: {
             'name': config['cloudServersOpenStack'],
-            'region': config['region'],
+            'region': region,
         },
         ServiceType.CLOUD_LOAD_BALANCERS: {
             'name': config['cloudLoadBalancers'],
-            'region': config['region'],
+            'region': region,
         },
         ServiceType.CLOUD_ORCHESTRATION: {
             'name': config['cloudOrchestration'],
-            'region': config['region'],
+            'region': region,
         },
         ServiceType.RACKCONNECT_V3: {
             'name': config['rackconnect'],
-            'region': config['region'],
+            'region': region,
         }
     }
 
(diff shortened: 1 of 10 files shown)


manishtomar:

otter/constants.py
```diff
@@ -52,10 +59,22 @@ def get_service_configs(config):
         configs[ServiceType.CLOUD_METRICS_INGEST] = {
             'name': metrics['service'], 'region': metrics['region']}
 
+    # {"cloudfeeds"...} contains config for 2 feeds services: One where otter
+    # pushes scaling group events `otter.log.cloudfeeds` as {"url": ...}
+    # and other where it fetches customer access events for terminator which
+    # is {"customer_access_events": {"url": "http://url"} or
```
glyph: again, `https` I hope.  Honestly HTTP URLs should really be disallowed for anything containing a token.

otter/test/cloud_client/test_cloudfeeds.py
```diff
@@ -43,3 +47,147 @@ def test_publish_autoscale_event(self):
...
+                  "next": cf.Direction.NEXT}
+    service_type = ServiceType.CLOUD_FEEDS
+
+    def svc_intent(self, params={}):
+        return service_request(
+            self.service_type, "GET",
+            self.url, params=params, json_response=False).intent
+
+    def feed(self, rel, link, summaries):
+        return self.feed_fmt.format(
```
glyph: This strategy is going to quote all kinds of stuff improperly.  It might be good enough for this use-case, but, pro tip:

```
>>> from twisted.web.template import Tag, flattenString
>>> flattenString(None, Tag('feed')(xmlns="http://www.w3.org/2005/Atom")(Tag('link')(rel="something",
href="something"))).result
'<feed xmlns="http://www.w3.org/2005/Atom"><link href="something" rel="something" /></feed>'
```

otter/constants.py
manishtomar: This is a comment!

glyph: A always B be C cryptographically verifying even hypothetical network connectivity

PR #1968, opened 2017-01-05
use upstream effect testing functions

manishtomar:
As all of these functions have been moved to effect. See https://github.com/python-effect/effect/pull/74


manishtomar:
@glyph I can bare this PR from your precious time since this is merely test changes from otter to effect.

glyph:
Finally had a chance to look at it, and I would have said "+1, go ahead" anyway, so - OK :).

